PR #1, opened 2025-07-29
koalabear: spec finite field

tcoratger:


src/lean_spec/subspecs/koalabear/field.py
```diff
@@ -0,0 +1,95 @@
...
+# Base Field Fp
+#
+# This class implements the finite field F_p where p is the KoalaBear prime.
+# All arithmetic is performed modulo P.
+# ==============================================================================
+
+class Fp:
+    """An element in the KoalaBear prime field F_p."""
+
+    def __init__(self, value: int):
```
fselmo: Since this `__init__()` performs some change on the input, I think this may be a perfect time to start playing around with some readable `pydantic` model implementations. If you think it takes away from the readability, we can definitely reconsider.

Something like:

```py
from pydantic import BaseModel, Field, field_validator

class Fp(BaseModel):
    """An element in the KoalaBear prime field F_p."""
    
    value: int = Field(
        ge=0,  # greater than or equal to
        lt=P,  # less than
        description="Field element value in range [0, P)"
    )
    
    # This field_validator will run every time an input is passed into `value`
    @field_validator('value', mode='before')
    @classmethod
    def reduce_modulo_p(cls, v: int) -> int:
        """Reduces an integer input modulo P."""
        if not isinstance(v, int):
            raise ValueError(f"Field element must be an integer, got {type(v).__name__}")
        return v % P
```

We could define methods as such:

```py
from typing import Self

def __add__(self, other: Self) -> Self:
    return Fp(value=self.value + other.value)


@classmethod
def two_adic_generator(cls, bits: int) -> Self:
    """
    Get a generator for the multiplicative subgroup of order 2^bits.
    
    Args:
        bits: (I'll leave this part to you 😅)
    
    Returns:
        A generator of the multiplicative subgroup of order 2^bits.
    
    Raises:
        ValueError: If bits is out of valid range.
    """
    if not (0 <= bits <= TWO_ADICITY):
        raise ValueError(f"bits must be between 0 and {TWO_ADICITY}")
    return cls(value=TWO_ADIC_GENERATORS[bits])
```

Using `Self` for the type here is better I think since if this is sub-classed it can correctly use the subclass instead of the hard-coded `"Fp"`.

tests/lean_spec/subspecs/koalabear/test_field.py
```diff
@@ -0,0 +1,62 @@
...
+def test_constants():
+    """Verify field constants."""
+    assert P == 2**31 - 2**24 + 1
+    assert (P - 1) % (2**TWO_ADICITY) == 0
+    assert (P - 1) % (2**(TWO_ADICITY + 1)) != 0
+    assert len(TWO_ADIC_GENERATORS) == TWO_ADICITY + 1
+
+def test_base_field_arithmetic():
+    """Test basic arithmetic, equality, and error handling in the base field Fp."""
+    a = Fp(5)
```
fselmo: If we go with pydantic, these would change to `a = Fp(value=5)`, etc.

fselmo (commented):
Thanks @tcoratger! This seems like the perfect time to try out some design ideas since this implementation is on the smaller end. I left a few questions on leveraging `pydantic` for the models and some minor comments.

I think we can use `Self` from `typing` at the very least. And we should figure out a good way to check lint while we don't have a CI set up yet but you can fix a couple and see the remaining lint errors with:

```
ruff format src tests && ruff check --fix src tests
```

Curious on your thoughts.

src/lean_spec/subspecs/koalabear/field.py
tcoratger: Thanks this is solved, I've used `pydantic`. I didn't know that much about this (I'm not a Python expert ahah) but I love this definitely! Super practical for our use case.

tests/lean_spec/subspecs/koalabear/test_field.py
tcoratger: Solved :)

fselmo:
> Thanks this is solved, I've used pydantic. I didn't know that much about this (I'm not a Python expert ahah) but I love this definitely! Super practical for our use case.

Awesome! Yeah, will be very nice to start with pydantic from the start 👌🏼 

> Small point: what's currently stopping us from having a CI with the linter running inside to automatically run it on all the PRs and commits we create?

Absolutely nothing, I can put something together today 😄

fselmo:
> Small point: what's currently stopping us from having a CI with the linter running inside to automatically run it on all the PRs and commits we create?

@tcoratger merged in [my PR to add this](https://github.com/leanEthereum/leanSpec/pull/2). We're triggering lint checks and tests for ubuntu and macOS for both supported python versions (3.12 and 3.13). We can tweak as needed!

tcoratger:
> > Small point: what's currently stopping us from having a CI with the linter running inside to automatically run it on all the PRs and commits we create?
> 
> @tcoratger merged in [my PR to add this](https://github.com/leanEthereum/leanSpec/pull/2). We're triggering lint checks and tests for ubuntu and macOS for both supported python versions (3.12 and 3.13). We can tweak as needed!
> 
> edit: Check runs for `main` [here](https://github.com/leanEthereum/leanSpec/actions/runs/16628427539/job/47051188997).

Thanks a lot :)

PR #3, opened 2025-07-31
subspec: implement Poseidon2 spec with tests

tcoratger:


src/lean_spec/subspecs/poseidon2/permutation.py
```diff
@@ -0,0 +1,325 @@
...
+"""
+The S-box exponent `d`.
+
+For fields where `gcd(d, p-1) = 1`, `x -> x^d` is a permutation.
+
+For KoalaBear, `d=3` is chosen for its low degree.
+"""
+
+
+class Poseidon2Params(NamedTuple):
```
fselmo: Hey @tcoratger, don't want to distract from a more proper review on the implementation itself but I'm going to try to shill `pydantic` here again as a base for (most) models 😅. I think validation alone could be useful here. Descriptions can be inline on fields and if we ever have JSON test vectors for these (or any other model), `pydantic` has native JSON serialization / deserialization (and can account for camelCase conversions, `alias` fields, etc).

There are a few nice configurations for these classes but just the added validation benefit is useful imo - whereas `NamedTuple` only provides static type hinting for the inputs.

```py
class Poseidon2Params(BaseModel):
    # can override configurations on the class, such as:
    model_config = ConfigDict(
        frozen=True,  # makes immutable like `NamedTuple`
        extra="forbid",  # prevents extra arguments from being passed in
        arbitrary_types_allowed=True,  # if any issues with Fp since I did not test this
        validate_default=True,  # validates default values as well
    )
    
    # choose between `ge` and `gt` appropriately here...
    width: int = Field(gt=0, description="The size of the state (t)")
    rounds_f: int = Field(gt=0, description="Full rounds count")
    rounds_p: int = Field(ge=0, description="Partial rounds count")
    internal_diag_vectors: List[Fp] = Field(min_length=1, description="Diagonal vectors")
```

Let me know your thoughts. And feel free to disagree if it doesn't need validation or reads better this way, etc. Just trying to give some Python design ideas while we're still early in the implementation process 🙂. This is my only thought on the PR, lgtm otherwise 👌🏼.

tcoratger: Thank you so much for your review. I love this and I just fixed it. I definitely have some improvement to make at this level of Python. I used to use it only for very small scripts without any caution about types and that kind of stuff, but I love Pydantic.

I just implemented a stronger validation.

fselmo: Happy to hear you like it. Yeah it's quite useful in so many ways to help Python with type validation. Setting a standard for it in this repo early will only help us in the future I think. I was the same way some years back where I'd only used Python for scripts and smaller projects but it has come a long way! `pydantic` is relatively new but already establishing itself as a very useful standard for models.

Thanks for entertaining it here 😄

src/lean_spec/subspecs/poseidon2/permutation.py
```diff
@@ -0,0 +1,346 @@
...
+    total_constants = (params.rounds_f * params.width) + params.rounds_p
+
+    # For the specification, we generate the constants as a deterministic d
+    # sequence of integers.
+    #
+    # This is sufficient to define the algorithm's mechanics.
+    #
+    # Real-world implementations would use constants generated from a secure,
+    # pseudo-random source.
+    return [Fp(value=i) for i in range(total_constants)]
```
khovratovich, line 92: Using arbitrary constant values whereas the rest of the spec follows the original paper looks odd. Consider pre-generating the constants and storing them in a constant array, or rename the function as non-compliant.

tcoratger: Thanks for your comment. I just fixed it and I agree.

Actually, having this super simple way of setting constants without having to generate an arbitrary array allowed me to compare the results with an existing implementation without adding too many lines of code with a hardcoded array because I didn't have any specific values to put in this array other than random ones.

What I propose:
- I've modified the name and description of the existing function to clearly state that it's for generating test values, not something production-ready.
- I'll open a follow-up issue. For BabyBear, we have hardcoded round constants: https://github.com/Plonky3/Plonky3/blob/bd6fb4116132d64c100cd5338567354b8d7dab1f/baby-bear/src/poseidon2.rs#L185-L243 which come from the HorizonLabs implementation (I imagine this set of values provides some advantages in terms of speed) but we don't have the equivalent for KoalaBear. Maybe it could be nice to have such a list for KoalaBear and once we have it, hardcode it inside the spec to remove the function and rely on this efficient round constants.

PR #5, opened 2025-08-06
feat: 1st draft pqdevnet0

unnawut:
## 🗒️ Description

Introducing the 1st draft of the specs for pqdevnet-0.

## 🔗 Related Issues or PRs

N/A

## ✅ Checklist
    ```console
    uvx --with=tox-uv tox
    ```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,6 +19,7 @@ classifiers = [
 ]
 requires-python = ">=3.12"
 dependencies = [
+    "remerkleable>=0.1.28,<0.2",
     "pydantic>=2.9.2,<3",
     "typing-extensions>=4.4",
 ]
--- a/src/lean_spec/subspecs/pqdevnet-0/block.py
+++ b/src/lean_spec/subspecs/pqdevnet-0/block.py
@@ -1,0 +1,28 @@
+"""
+A `Block` is a single link in the Lean Consensus chain. Each `Block` contains
+a `Header` and zero or more transactions. Each `Header` contains associated
+metadata like the block number, parent block hash, and how much gas was
+consumed by its transactions.
+
+Together, these blocks form a cryptographically secure journal recording the
+history of all state transitions that have happened since the genesis of the
+chain.
+"""
+
+from dataclasses import dataclass
+from remerkleable.basic import uint64
+from remerkleable.byte_arrays import Bytes32
+from remerkleable.complex import List
+from pydantic import BaseModel, ConfigDict
+
+from preset import VALIDATOR_REGISTRY_LIMIT
+from vote import Vote
+
+@dataclass
+class Block(BaseModel):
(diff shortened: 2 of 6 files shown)


pyproject.toml
```diff
@@ -20,6 +20,7 @@ classifiers = [
 requires-python = ">=3.12"
 dependencies = [
     "pydantic>=2.9.2,<3",
+    "remerkleable>=0.1.28,<0.2",
```
fselmo: It would be quite nice if [ethereum-types](https://github.com/ethereum/ethereum-types) could be used here instead. These are the types that were created and used by the [execution-specs](https://github.com/ethereum/execution-specs) team.

---
edit: see my comment below about `py-ssz` perhaps helping here as well.

src/lean_spec/subspecs/pqdevnet-0/block.py
```diff
@@ -0,0 +1,27 @@
...
+from pydantic import BaseModel, ConfigDict
+
+from preset import VALIDATOR_REGISTRY_LIMIT
+from vote import Vote
+
+@dataclass
+class Block(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    slot: uint64
```
fselmo: If we go with `ethereum-types`, we could use [U64](https://github.com/ethereum/ethereum-types/blob/master/src/ethereum_types/numeric.py#L743-L752) here, for example.

```suggestion
    slot: U64
```

src/lean_spec/subspecs/pqdevnet-0/block.py
```diff
@@ -0,0 +1,27 @@
...
+    parent: Bytes32
```
fselmo: [Bytes32](https://github.com/ethereum/ethereum-types/blob/master/src/ethereum_types/bytes.py#L90-L98) could be used here.

fselmo (commented):
Hey @unnawut. This looks good to me. It would be nice to try to share some types that the execution-specs team put together specifically for Ethereum types. I added that as a comment here. That would help with homogeneity across spec repos for users and for spec readers.

I'm also not sure if the pqdevnet should be under subspecs but feel free to arrange these dirs however you see best. If these are more general types, perhaps they belong under the `lean_spec` directory. But there's no reason they can't be moved to a broader scope when it makes sense to.

unnawut:
From interop call Guillaume pointed out that zeam uses the [union type](https://ethereum.github.io/consensus-specs/ssz/simple-serialize/#union) for optional field which makes sense.

I'll try it out and update this PR accordingly.

pyproject.toml
unnawut: Looking into it!

unnawut:
Closing in favor of #8. Will reopen this PR when there's more clarity around the specs format.

PR #7, opened 2025-08-13
pqdevnet0: networking specs

unnawut:
## 🗒️ Description

Introducing the 1st draft of the networking specs for pqdevnet-0. Basically a trimmed down version of [phase0 networking](https://ethereum.github.io/consensus-specs/specs/phase0/p2p-interface/).

## 🔗 Related Issues or PRs
N/A

## ✅ Checklist

- [x] Ran `tox` checks to avoid unnecessary CI fails:
    ```console
    uvx --with=tox-uv tox
    ```
- [x] Considered adding appropriate tests for the changes.
- [x] Considered updating the online docs in the [./docs/](/leanEthereum/leanSpec/tree/main/docs/) directory.

PR #8, opened 2025-08-13
pqdevnet0: client specs

unnawut:
## 🗒️ Description
Introducing the 1st draft of pqdevnet-0 client specs.

This PR contains only the markdown specs. The python specs is paused for clearer direction before resuming.

Closes https://github.com/leanEthereum/pm/issues/27

## 🔗 Related Issues or PRs
Combined #5 and #7.

## ✅ Checklist

    ```console
    uvx --with=tox-uv tox
    ```


docs/client/networking.md
```diff
@@ -0,0 +1,164 @@
+# Networking
+
+## Setup
+
+- Transport: QUIC on IPv4
+- Encryption and identification: [Libp2p-noise](https://github.com/libp2p/specs/tree/master/noise) with `secp256k1` identities
```
unnawut: Carried over @kamilsa's comment from previous PR:

> QUIC already integrates TLS, so maybe we don't have to bother with Noise?

docs/client/networking.md
```diff
@@ -0,0 +1,164 @@
...
+  responses): 6
+- `mcache_gossip` (number of windows to gossip about): 3
+- `seen_ttl` (expiry time for cache of seen message ids, seconds):
+  SECONDS_PER_SLOT * SLOTS_PER_EPOCH * 2
+
+#### Topics and messages
+
+Topics are plain UTF-8 strings and are encoded on the wire as determined by
+protobuf (gossipsub messages are enveloped in protobuf messages). Topic strings
+have form: `/leanconsensus/devnet0/Name/Encoding`. This defines both the type of
```
unnawut: Originally `/eth2/ForkDigestValue/Name/Encoding` in the beacon specs

docs/client/networking.md
```diff
@@ -0,0 +1,164 @@
...
+invalid payload.
+
+#### Encodings
+
+Topics are post-fixed with an encoding. Encodings define how the payload of a
+gossipsub message is encoded.
+
+- `ssz_snappy` - All objects are SSZ-encoded and then compressed with
+  [Snappy](https://github.com/google/snappy) block compression. Example: The
+  lean block topic string is `/leanconsensus/devnet0/lean_block/ssz_snappy`,
```
unnawut: Originally `/eth2/446a7232/beacon_aggregate_and_proof/ssz_snappy`

docs/client/networking.md
```diff
@@ -0,0 +1,166 @@
...
+case-sensitive UTF-8 string of the form:
+
+```
+/ProtocolPrefix/MessageName/SchemaVersion/Encoding
+```
+
+With:
+
+- `ProtocolPrefix` - messages are grouped into families identified by a shared
+  libp2p protocol name prefix. In this case, we use `/leanconsensus/req`.
```
unnawut: Originally `/eth2/beacon_chain/req`

docs/client/networking.md
```diff
@@ -0,0 +1,166 @@
...
+  [Snappy](https://github.com/google/snappy) frames compression. For objects
+  containing a single field, only the field is SSZ-encoded not a container with
+  a single field. For example, the `LeanBlocksByRoot` request is an
+  SSZ-encoded list of `Root`'s.
+
+#### Messages
+
+##### LeanBlocksByRoot v1
+
+**Protocol ID:** `/leanconsensus/req/lean_blocks_by_root/1/`
```
unnawut: Originally `/eth2/beacon_chain/req/beacon_blocks_by_root/1/`

docs/client/networking.md
```diff
@@ -0,0 +1,140 @@
...
+`data` can decompress to the same value, and (2) some message `data` can fail to
+snappy decompress altogether.
+
+The payload is carried in the `data` field of a gossipsub message, and varies
+depending on the topic:
+
+| Name                             | Message Type              |
+| -------------------------------- | ------------------------- |
+| `lean_block`                     | `Block`                   |
+| `lean_attestation`               | `Attestation`             |
```
syjn99:
```suggestion
| `lean_block`                     | `SignedBlock`                   |
| `lean_vote`               | `SignedVote`             |
```

Shouldn't signed data be carried on P2P network? Even though we don't support PQ signature in this phase, it seems more forward compatible.

docs/client/validator.md
syjn99, line 1: As multiple validators can run on a single node, I think ENR isn't appropriate for identification. Actually the only information we need to run a node in the context of `pqdevnet0` is the size of the (fixed) validator set for determining the duty. After PQC is integrated, we may need initial pubkeys for sig verification.

For networking and peer connection, I think it will be better to provide the list of static peers which is consisted of ENRs.

docs/client/networking.md
unnawut: @kaydenML @varun-doshi @kolbyML Can I defer to you guys on Kamil's question? I don't know deep enough on this. Also @g11tech.

docs/client/networking.md
unnawut: Yep agreed

docs/client/validator.md
unnawut: I've separated node vs validator identification in d4881e1. Please have a look!

docs/client/networking.md
```diff
@@ -0,0 +1,151 @@
...
+- `D_high` (topic stable mesh high watermark): 12
+- `D_lazy` (gossip target): 6
+- `heartbeat_interval` (frequency of heartbeat, seconds): 0.7
+- `fanout_ttl` (ttl for fanout maps for topics we are not subscribed to but have
+  published to, seconds): 60
+- `mcache_len` (number of windows to retain full messages in cache for `IWANT`
+  responses): 6
+- `mcache_gossip` (number of windows to gossip about): 3
+- `seen_ttl` (expiry time for cache of seen message ids, seconds):
+  SECONDS_PER_SLOT * SLOTS_PER_EPOCH * 2
```
syjn99, line 69: We probably don't have those configuration values.

unnawut: Assuming an epoch now becomes just 3 slots. So maybe this?

```suggestion
- `seen_ttl` (expiry time for cache of seen message ids, seconds):
  SECONDS_PER_SLOT * 3 * 2
```

docs/client/networking.md
```diff
@@ -0,0 +1,151 @@
+# Networking
+
+## Setup
+
+- Transport: QUIC on IPv4
+- Encryption and identification: [Libp2p-noise](https://github.com/libp2p/specs/tree/master/noise) with `secp256k1` identities
+- Protocol negotiation: [multistream-select 1.0](https://github.com/multiformats/multistream-select/)
+- Multiplexing: Native support by QUIC
+- Gossip: [gossipsub v1](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.0.md)
```
unnawut, line 8: Note that I went with gosipsub v1.0 rather than v1.1 because peer scoring seems overkill for devnet0 but my understanding here is limited.

docs/client/networking.md
```diff
@@ -0,0 +1,151 @@
+# Networking
+
+## Setup
+
+- Transport: QUIC on IPv4
+- Encryption and identification: [Libp2p-noise](https://github.com/libp2p/specs/tree/master/noise) with `secp256k1` identities
```
GrapeBaBa: When using QUIC, TLS 1.3 is the default security protocol, so that noise is not needed, right? And may I suggest to use `ED25519` for identities, because native Zig not support it right now. Use TLS1.3 and `ED25519` could a bit simplify zeam works.

docs/client/networking.md
KolbyML: Wouldn't an epoch be 1 slot?

unnawut: > Wouldn't an epoch be 1 slot?

I picked 3 slots from the max lookback that's relevant to 3SF's justification & finalization. Then `* 2` for ["This value should be chosen to approximate the propagation delay in the overlay, within a healthy margin."](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.0.md#message-cache) but not too sure

cc: @kamilsa thoughts?

KolbyML: That is fine, I just thought epochs is not a concept in 3sf so I was just a little confused

docs/client/networking.md
```diff
@@ -0,0 +1,151 @@
...
+  [Snappy](https://github.com/google/snappy) frames compression. For objects
+  containing a single field, only the field is SSZ-encoded not a container with
+  a single field. For example, the `LeanBlocksByRoot` request is an
+  SSZ-encoded list of `Root`'s.
+
+#### Messages
+
+##### LeanBlocksByRoot v1
+
+**Protocol ID:** `/leanconsensus/req/lean_blocks_by_root/1/`
```
KolbyML: Can we add a `status` message too, so we know what our peer's perceived head is?

KolbyML: As it is hard to use `lean_blocks_by_root` to forward sync unless we plan to solely use gossip messages to determine what to backfill from I guess

docs/client/networking.md
```diff
@@ -0,0 +1,151 @@
...
+
+- Transport: QUIC on IPv4
+- Encryption and identification: [Libp2p-noise](https://github.com/libp2p/specs/tree/master/noise) with `secp256k1` identities
+- Protocol negotiation: [multistream-select 1.0](https://github.com/multiformats/multistream-select/)
+- Multiplexing: Native support by QUIC
+- Gossip: [gossipsub v1](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.0.md)
+
+## Node identification
+
+Nodes are defined as a list of their ENRs in a yaml file at [`src/lean_spec/client/nodes.yaml`](../../src/lean_spec/client/nodes.yaml).
```
kamilsa, line 12: I am not super familiar with ENR format, however it would be great if we could include there information about peers role. 

In the first devnet everyone can have attester role, in future we may add aggregator role. With that knowledge we can significantly optimize gossiping when we start experimenting with something more exotic than traditional gossipsub

docs/client/networking.md
kamilsa: this cache shouldn't consume too much RAM and most of the messages will arrive in less than 2 seconds. So either option is fine

docs/client/chain.md
```diff
@@ -0,0 +1,22 @@
...
+| `FAST_CONFIRM_DUE_BPS`                | `uint64(7500)`            |      bps     | 75% into slot |
+| `VIEW_FREEZE_CUTOFF_BPS`              | `uint64(7500)`            |      bps     | 75% into slot |
+
+## Presets
+
+### State list lengths
+
+| Name                           | Value                                 |       Unit       |   Duration    |
+| ------------------------------ | ------------------------------------- | :--------------: | :-----------: |
+| `HISTORICAL_ROOTS_LIMIT`       | `uint64(2**18)` (= 262,144)           | historical roots |   12.1 days   |
```
g11tech, line 18: i think we need to modify 3sf mini to junk tracking & hence try justifying blocks beyond a certain window. may be that is 12 days, or we mention here that this might be just 3sf mini artifact
(although for our current purposes keeping it 12 days is a good simplification)

docs/client/containers.md
```diff
@@ -0,0 +1,78 @@
+# Containers
+
+## `Config`
+
+```python
+class Config(Container):
+    num_validators: uint64
```
g11tech, line 8:
```suggestion
    num_validators: uint64
    genesis_time: uint64
```
gensis time is a crucial property but its relevance is not same as it was when beacon chain had to reach some conditions for genesis to happen.
so may be we should discuss this because it could totally be a config value.

genesis validators root (which has similar origins for beacon chain) might not be required at all though because validators will be part of the state (devnet 1 onwards)

g11tech, line 8:
```suggestion
    // temporary property to support simplified round robin block production in absence of randao & deposit mechanisms
    num_validators: uint64
```

docs/client/containers.md
```diff
@@ -0,0 +1,78 @@
...
+class Checkpoint(Container):
+    root: Bytes32
+    slot: uint64
+```
+
+## `State`
+
+```python
+class State(Container):
+    config: Config
```
g11tech, line 24:
```suggestion
    config: Config
    slot: uint64
```
slot needs to be part of the state as even empty slots will cause state updates, slot anyway is integral part of state processing

docs/client/networking.md
unnawut: Added a new Status message as suggested: https://github.com/leanEthereum/leanSpec/pull/8/commits/e38593e58ef9adbcaf103b931558d874fe3e98e1

cc: @g11tech

docs/client/networking.md
unnawut: Is there something you'd like to test in pqdevnet0? Because if the info is not used immediately in pqdevnet0, I'd prefer to defer adopting it until it's needed actually. And introducing it later shouldn't be too troublesome since we all are expecting the specs to be changing a lot anyway.

docs/client/containers.md
unnawut: Simple enough and I don't think this creates any side effect in pqdevnet0 so I'll introduce this into the specs.

Flagging @syjn99 @KolbyML @kaydenML @shariqnayer for visibility

docs/client/containers.md
unnawut: I think it'll be relevant for production, but for pqdevnet0 especially without proper state validation I'm not sure we need this...

I think its only use now is preventing an orphan validator from a previous setup trying to join the network which is probably unlikely?

docs/client/containers.md
```diff
@@ -0,0 +1,80 @@
...
+    slot: uint64
```
g11tech, line 25:
```suggestion
    slot: uint64
    latest_block_header: BlockHeader
```
this is needed to track and validate the block is being applied to the correct parent state, and is basic for chaining

docs/client/containers.md
```diff
@@ -0,0 +1,80 @@
...
+        HISTORICAL_ROOTS_LIMIT * VALIDATOR_REGISTRY_LIMIT
+    ]
+```
+
+## `Block`
+
+```python
+class Block(Container):
+    slot: uint64
+    parent: Bytes32
```
g11tech:
```suggestion
    parent_root: Bytes32
```

docs/client/containers.md
```diff
@@ -0,0 +1,80 @@
...
+    votes: List[Vote, VALIDATOR_REGISTRY_LIMIT]
```
g11tech:
```suggestion
    votes: List[Vote, VALIDATOR_REGISTRY_LIMIT]
```
this should go inside body, we should be close to the beacon abstractions, also easier to lift and rely upon some of the blockToHeader and also blockToLatestBlockHeader abstractions of the beacon

docs/client/containers.md
```diff
@@ -0,0 +1,80 @@
...
+    state_root: Bytes32
```
g11tech, line 49: its also better to put state_root before votes in terms of having the variable body/votes part towards the end

docs/client/containers.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+## `SignedBlock`
+
+```python
+class SignedBlock(Container):
+    data: Block,
+    signature: Bytes32,
```
g11tech, line 68: should we put the correct signature size here even if its 
```suggestion
    signature: Bytes32,
```
we should choose the correct size even if its zero bytes to converge to the params we are going for

image
https://github.com/b-wagn/hashsig-parameters

docs/client/containers.md
```diff
@@ -0,0 +1,80 @@
...
+    parent: Bytes32
+    votes: List[Vote, VALIDATOR_REGISTRY_LIMIT]
+    state_root: Bytes32
+```
+
+## `SignedBlock`
+
+```python
+class SignedBlock(Container):
+    data: Block,
```
g11tech:
```suggestion
    message: Block,
```
staying closer to beacon terminology

docs/client/containers.md
```diff
@@ -0,0 +1,80 @@
...
+    justifications_validators: Bitlist[
+        HISTORICAL_ROOTS_LIMIT * VALIDATOR_REGISTRY_LIMIT
+    ]
+```
+
+## `Block`
+
+```python
+class Block(Container):
+    slot: uint64
```
g11tech, line 46:
```suggestion
    slot: uint64
    proposer_index: uin64
```
for simple slot proposer validation

docs/client/networking.md
g11tech: cc: @GrapeBaBa @spiral-ladder
without knowing the nitty gritties, I would tend to agree with @kamilsa if TLS is sufficient

docs/client/networking.md
g11tech: yes peer scoring is a mute point as this is a trusted setup

docs/client/networking.md
g11tech: yes we can definitely add a field in the ENR about advertising roles of a node, but we can do this in next devnet. also ENR eventually won't be sufficient as we will need to add metadata req/resp which would also need this role info

docs/client/networking.md
```diff
@@ -0,0 +1,202 @@
...
+
+**Protocol ID:** `/meshsub/1.0.0`
+
+**Gossipsub Parameters**
+
+The following gossipsub
+[parameters](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.0.md#parameters)
+will be used:
+
+- `D` (topic stable mesh target count): 8
```
g11tech, line 58: @GrapeBaBa @spiral-ladder
pls review there params and add comments of any

docs/client/networking.md
g11tech: @GrapeBaBa @spiral-ladder

docs/client/networking.md
g11tech: we should probably retain ForkDigestValue here instead of devnet0

docs/client/networking.md
```diff
@@ -0,0 +1,202 @@
...
+*Note*: The above logic handles two exceptional cases: (1) multiple snappy
+`data` can decompress to the same value, and (2) some message `data` can fail to
+snappy decompress altogether.
+
+The payload is carried in the `data` field of a gossipsub message, and varies
+depending on the topic:
+
+| Name                             | Message Type              |
+| -------------------------------- | ------------------------- |
+| `lean_block`                     | `SignedBlock`                   |
```
g11tech:
```suggestion
| `block`                     | `SignedBlock`                   |
```
wouldn't prefer lean prefix here

docs/client/networking.md
```diff
@@ -0,0 +1,202 @@
...
+  responses): 6
+- `mcache_gossip` (number of windows to gossip about): 3
+- `seen_ttl` (expiry time for cache of seen message ids, seconds):
+  SECONDS_PER_SLOT * SLOTS_PER_EPOCH * 2
+
+#### Topics and messages
+
+Topics are plain UTF-8 strings and are encoded on the wire as determined by
+protobuf (gossipsub messages are enveloped in protobuf messages). Topic strings
+have form: `/leanconsensus/devnet0/Name/Encoding`. This defines both the type of
```
g11tech:
```suggestion
have form: `/ethconsensus/lean/Name/Encoding`. This defines both the type of
```
or how about this? ties into narrative that lean is a fork not a separate chain

docs/client/networking.md
```diff
@@ -0,0 +1,202 @@
...
+The optional `from` (1), `seqno` (3), `signature` (5) and `key` (6) protobuf
+fields are omitted from the message, since messages are identified by content,
+anonymous, and signed where necessary in the application layer.
+
+The `message-id` of a gossipsub message MUST be the following 20 byte value
+computed from the message data:
+
+- If `message.data` has a valid snappy decompression, set `message-id` to the
+  first 20 bytes of the `SHA256` hash of the concatenation of
+  `MESSAGE_DOMAIN_VALID_SNAPPY` with the snappy decompressed message data, i.e.
```
g11tech, line 92: need to have MESSAGE_DOMAIN_VALID_SNAPPY and MESSAGE_DOMAIN_INVALID_SNAPPY values copied over or we can assume any unmentioned references go to beacon specs

docs/client/networking.md
syjn99: @GrapeBaBa 

> that noise is not needed, right? 

Yes, we can delete the noise part. [Using QUIC **implies** using TLS1.3](https://github.com/libp2p/specs/blob/master/tls/tls.md#handshake-protocol), which means we can make this document more neat.

> use `ED25519` for identities

Either works for Ream. We're using `rust-libp2p` and changing DS scheme is not a problem.

---

So I would suggest replacing those two points into one bullet point:

```
- Transport: QUIC on IPv4, secured by TLS 1.3 with `ED25519` identities
```

cc. @unnawut

unnawut: We just discussed about this in ream call and looks doable to us. But before I update accordingly can I also flag @kamilsa for opinion since qdvrm is also planning to participate as a client?

docs/client/chain.md
syjn99, line 1: I'd like to suggest one modification from original [3sf-mini](https://github.com/ethereum/research/tree/master/3sf-mini) specification, regarding **calculation of the current slot**.  This is one minor change but this is something we should agree on.

TL;DR: **Don't add `2` when calculating the slot** and **Do nothing until `current_slot < 2`**. ([Diff](https://github.com/syjn99/research/pull/1/commits/4dfca57e7486691604f25d094b8878b6e6e70104#diff-c851d58a1c6f249a74d3739430600640f430be46ed8649c0172f97757cdc7989))

---
```python
    def get_current_slot(self):
        return self.network.time // SLOT_DURATION + 2
```

I'm feeling quite bizarre to follow the above function ([Link](https://github.com/ethereum/research/blob/d225a6775a9b184b5c1fd6c830cc58a375d9535f/3sf-mini/p2p.py#L100-L101)), as it seems counter-intuitive and doesn't match with the behavior of current beacon chain specification. 

Regarding consensus, the beacon chain does nothing until it is ready for voting source and target like:

```python
def process_justification_and_finalization(state: BeaconState) -> None:
    # Initial FFG checkpoint values have a `0x00` stub for `root`.
    # Skip FFG updates in the first two epochs to avoid corner cases that might result in modifying this stub.
    if get_current_epoch(state) <= GENESIS_EPOCH + 1:
        return
```
I suggest we follow this behavior for our `pq-devnet`s. I've tested a [minor code modification](https://github.com/syjn99/research/pull/1/commits/4dfca57e7486691604f25d094b8878b6e6e70104#diff-c851d58a1c6f249a74d3739430600640f430be46ed8649c0172f97757cdc7989) in 3sf-mini, and the change doesn't break consensus. Instead, it increases familiarity with the concept of a "slot."

I'm interested to hear your thoughts on this proposed change. @g11tech and others, what do you think?

docs/client/networking.md
unnawut: For "lean is a fork not a separate chain", wouldn't it be `/eth2/lean/Name/Encoding`?

docs/client/networking.md
unnawut: I moved them over here so we don't have to refer to two specs back and forth: https://github.com/leanEthereum/leanSpec/pull/8/commits/29913a29fcbeeeddccb77302cd0d6e0bf783c730

docs/client/networking.md
unnawut: Sorry I was just reminded @kamilsa suggested exactly this in https://github.com/leanEthereum/leanSpec/pull/8#discussion_r2274018644 so I'll just update this

docs/client/networking.md
unnawut: Looks like the discussion is happening in two places, the other one at https://github.com/leanEthereum/leanSpec/pull/8#discussion_r2278848830 but we're in agreement to drop noise so I'll do that accordingly.

docs/client/networking.md
unnawut: Updated in 65daac2

docs/client/networking.md
unnawut: Updated in [65daac2](https://github.com/leanEthereum/leanSpec/pull/8/commits/65daac241956ac13f77ac2eff072f718ca7937b2)

docs/client/networking.md
```diff
@@ -0,0 +1,210 @@
+# Networking
+
+## Setup
+
+- Transport: QUIC on IPv4, secured by TLS 1.3 with `ED25519` identities
+- Protocol negotiation: [multistream-select 1.0](https://github.com/multiformats/multistream-select/)
+- Multiplexing: Native support by QUIC
+- Gossip: [gossipsub v1](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.0.md)
+
+## Node identification
```
Vid201, line 10: It would make sense to specify exactly how the ENR will look since it will deviate from the standard form defined in the [EIP-778](https://eips.ethereum.org/EIPS/eip-778).

docs/client/chain.md
```diff
@@ -0,0 +1,22 @@
+# Lean Consensus Experimental Chain
+
+## Configuration
+
+### Time parameters
+
+| Name                                  | Value                     |     Unit     |   Duration    |
+| ------------------------------------- | ------------------------- | :----------: | :-----------: |
+| `SLOT_DURATION_MS`                    | `uint64(4000)`            | milliseconds | 4 seconds     |
+| `PROPOSER_REORG_CUTOFF_BPS`           | `uint64(2500)`            |      bps     | 25% into slot |
```
unnawut: Since we're doing just simple `/4` interval, I'll replace these 4 params with `INTERVALS_PER_SLOT: 4`

docs/client/networking.md
GrapeBaBa: Default params should be OK.

docs/client/containers.md
unnawut: but meh, why not! :D

docs/client/networking.md
Kayden-ML: topic_prefix/fork/kind/encoding

I think it should be `lean_consensus/devnet0/Name/Encoding`  



> For "lean is a fork not a separate chain", wouldn't it be `/eth2/lean/Name/Encoding`?

Lean Consensus isn't eth2, eth2 means Beacon Chain. We should have a fork id or devnet0 or B32::from_slice("devnet0".as_bytes()). Eth1 Would be the execution chain etc.

The Execution Chain and Beacon Chain are separate chains. The Lean Consensus Chain is separate from the Beacon Chain they are distinct.

We will have forks of the lean chain, also we will have multiple different devnets. We want to distinguish to p2p peers that devnet0 and devnet1 and mainnet aren't the same fork.

We are swapping the Consensus chain from Beacon to Lean, well maintaining the Execution Chain.

docs/client/networking.md
```diff
@@ -0,0 +1,210 @@
...
+`data` can decompress to the same value, and (2) some message `data` can fail to
+snappy decompress altogether.
+
+The payload is carried in the `data` field of a gossipsub message, and varies
+depending on the topic:
+
+| Name                             | Message Type              |
+| -------------------------------- | ------------------------- |
+| `block`                     | `SignedBlock`                   |
+| `lean_vote`               | `SignedVote`             |
```
Kayden-ML:
```suggestion
| `lean_block`                     | `SignedBlock`                   |
| `lean_vote`               | `SignedVote`             |
```

On the Beacon Chain the name is `beacon_block` so I think we should use `lean_block`

unnawut: @g11tech Wdyt?

docs/client/networking.md
unnawut: Added https://github.com/leanEthereum/leanSpec/pull/8/commits/cdcf9ae71d7fea87ecabfe4e03dcf802966cc43a

But I'm not sure how feasible it is to replace the `secp256k1` field with `ed25519` in the ENR implementation, @g11tech @varun-doshi can you have a check? cc: @syjn99 @Kayden-ML

docs/client/networking.md
unnawut: `lean_consensus/devnet0/Name/Encoding` makes sense. But will wait to hear from @g11tech

docs/client/networking.md
KolbyML: All Enr implementations already support this though right. Sigp/enr does why wouldn't the others etc

KolbyML: https://github.com/sigp/enr/blob/master/src/lib.rs#L34

Read the comment

docs/client/networking.md
unnawut: Related convo at https://github.com/leanEthereum/leanSpec/pull/8#discussion_r2284614507 so what's concluded there will be updated here.

docs/client/containers.md
unnawut: Updated in https://github.com/leanEthereum/leanSpec/pull/8/commits/6ef1f1027e3cc6c8f2306473b1c5143c594c3d2a

docs/client/chain.md
```diff
@@ -0,0 +1,19 @@
+# Lean Consensus Experimental Chain
+
+## Configuration
+
+### Time parameters
+
+| Name                                  | Value                     |     Unit     |   Duration    |
+| ------------------------------------- | ------------------------- | :----------: | :-----------: |
+| `SLOT_DURATION_MS`                    | `uint64(4000)`            | milliseconds | 4 seconds     |
+| `INTERVALS_PER_SLOT`                  | `uint64(4)`               | intervals    | 1 second      |
```
g11tech, line 10: are intervals 4 to take in account ePBS?

docs/client/networking.md
g11tech: without getting into the discussions of if lean will be fork or a chain now, lets do `/consensus/lean/Name/Encoding` i just feel devnet0 there is a bit weird, but as such this isn't a blocker anyway

docs/client/networking.md
g11tech: yes, but even beacon there is sort of redundant, since lean is already part of the topic, but in any case this ain't a blocker as well. since we didn't prefix Lean in any of the container types, we should just follow the same pattern here

docs/client/validator.md
```diff
@@ -0,0 +1,31 @@
+# Validator
+
+## Validator identification
+
+To ensure a good distribution of block proposer duties in a round-robin manner
+and avoid clashing IDs, validator IDs are pre-assigned to each client
+implementation in a yaml file at 
+[`src/lean_spec/client/validators.yaml`](../../src/lean_spec/client/validators.yaml).
```
g11tech, line 8: so this file needs to be the output of the genesis generator?

docs/client/networking.md
g11tech: @GrapeBaBa does this works for us? i.e.  replacing the secp256k1  with ed25519 key in our ENR impl?

GrapeBaBa: Yes, ed25519 is more simple for us since the Zig std not support bitcoin secp256k1 right now.

GrapeBaBa: @unnawut @KolbyML @g11tech @Vid201 I just read [EIP-778](https://eips.ethereum.org/EIPS/eip-778), it only defines secp256k1 algo and currently only rust ENR implementation extends with ed25519, others such as JS, Go don't have that. And after read [zig-discv5](https://github.com/ChainSafe/zig-discv5) code, its design only support secp256k1 too. So it may be better to stick to secp256k1, I can use this [wrapper](https://github.com/Syndica/secp256k1-zig) to replace Zig std secp256k1 in the [zig-discv5](https://github.com/ChainSafe/zig-discv5).

g11tech: i guess we can go with ed25519 and we can extend 778 EIP into a new one since we are not dealing with beacon clients anyway

g11tech: on second thought, I wouldn't want to change this if this breaks tooling unless there is massive benefit. lets finalize this in this wednesday's call

docs/client/networking.md
KolbyML: then we should remove the `lean_` in `lean_vote`

docs/client/networking.md
KolbyML: we will be having multiple devnets and testnets, so I think we should have a fork field. We shouldn't have lean be the fork for all of our devnet's and testnet's

KolbyML: devnet0 hence should be the fork digest

KolbyML: if lean was a fork of the beacon chain we would use ``eth2/fork_digest/Name/Encoding`` etc

docs/client/networking.md
```diff
@@ -0,0 +1,228 @@
...
+
+##### LeanBlocksByRoot v1
+
+**Protocol ID:** `/leanconsensus/req/lean_blocks_by_root/1/`
+
+Request Content:
+
+```
+(
+  List[Root, MAX_REQUEST_BLOCKS]
```
KolbyML, line 202: Should we define MAX_REQUEST_BLOCKS here?

unnawut: Copied over `2*10` (= 1024) from beacon specs: d19d93

This is `1024 * 4 / 60 = 68` minutes worth of block under 4s block time

docs/client/chain.md
unnawut: https://github.com/ethereum/research/blob/master/3sf-mini/p2p.py#L78-L98

docs/client/containers.md
unnawut: To update to exact size (~3KiB) based on researchers' input

docs/client/containers.md
```diff
@@ -0,0 +1,90 @@
...
+    target: Checkpoint
+    source: Checkpoint
+```
+
+## `SignedVote`
+
+```python
+class SignedVote(Container):
+    data: Vote,
+    signature: Bytes32,
```
unnawut, line 87: To update to exact size (~3KiB) based on researchers' input

docs/client/networking.md
unnawut: Changed back to secp256k1

docs/client/networking.md
unnawut: `lean_*` prefixes removed

PR #9, opened 2025-08-13
xmss: scaffolding for the signature spec

tcoratger:


src/lean_spec/subspecs/xmss/interface.py
```diff
@@ -0,0 +1,100 @@
...
+def key_gen() -> Tuple[PublicKey, SecretKey]:
+    """
+    Generates a new cryptographic key pair.
+
+    This function is a placeholder. In a real implementation, it would involve
+    generating a master secret, deriving all one-time keys, and constructing
+    the full Merkle tree.
+
+    For the formal specification of this process, please refer to:
+    - "Hash-Based Multi-Signatures for Post-Quantum Ethereum" [DKKW25a]
```
b-wagn: provide eprint links?

src/lean_spec/subspecs/xmss/interface.py
```diff
@@ -0,0 +1,100 @@
...
+"""
+
+from __future__ import annotations
+
+from typing import Tuple
+
+from .structures import PublicKey, SecretKey, Signature
+
+
+def key_gen() -> Tuple[PublicKey, SecretKey]:
```
b-wagn: interface should take activation epoch and duration, right?

src/lean_spec/subspecs/xmss/interface.py
```diff
@@ -0,0 +1,100 @@
...
+    - "Technical Note: LeanSig for Post-Quantum Ethereum" [DKKW25b]
+    - The canonical Rust implementation: https://github.com/b-wagn/hash-sig
+    """
+    raise NotImplementedError(
+        "key_gen is not part of this specification. "
+        "See the Rust reference implementation."
+    )
+
+
+def sign(sk: SecretKey, epoch: int, message: bytes) -> Signature:
```
b-wagn, line 42: should also take a rng, right (and key gen should as well)? Or are we generally ignoring this for the Python spec?

src/lean_spec/subspecs/xmss/interface.py
```diff
@@ -0,0 +1,100 @@
...
+    1.  **Re-encode Message**: The verifier uses the randomness `rho` from the
+        signature to re-compute the codeword $x = (x_1, \dots, x_v)$ from the
+        message `m`.
+        This includes calculating the checksum or checking the target sum.
+
+    2.  **Reconstruct One-Time Public Key**: For each intermediate hash `y_i`
+        in the signature, the verifier completes the corresponding hash chain.
+        Since `y_i` was computed with $x_i$ steps, the verifier applies the
+        hash function an additional $w - 1 - x_i$ times to arrive at the
+        one-time public key component `otpk_{ep,i}`.
```
b-wagn: following the paper, I would call this pk_{ep,i}

src/lean_spec/subspecs/xmss/interface.py
```diff
@@ -0,0 +1,100 @@
...
+
+    3.  **Compute Merkle Leaf**: The verifier hashes the reconstructed one-time
```
b-wagn, line 85: Note: in a future impl I would maybe use the same sub-algorithms as in either the rust code or in the paper, and then maybe even point to the pages of the paper.

src/lean_spec/subspecs/xmss/constants.py
```diff
@@ -0,0 +1,87 @@
+"""Defines the cryptographic constants for the XMSS specification."""
+
+from ..koalabear import Fp
```
b-wagn, line 17: maybe we can leave a note that currently the rust code uses BabyBear but this will be updated and KoalaBear is the goal, and KoalaBear is consistent with the papers, see, e.g., Section 5 of the newer one.

src/lean_spec/subspecs/xmss/structures.py
```diff
@@ -0,0 +1,52 @@
...
+class HashTreeOpening(BaseModel):
+    """
+    A Merkle authentication path.
+
+    It contains a list of sibling nodes required to reconstruct the path
+    from a leaf node up to the Merkle root.
+    """
+
+    model_config = ConfigDict(frozen=True, arbitrary_types_allowed=True)
+    siblings: List[List[Fp]] = Field(
```
b-wagn: maybe it makes sense to define a type for hash output?

src/lean_spec/subspecs/xmss/structures.py
```diff
@@ -0,0 +1,52 @@
...
+        ..., description="List of sibling hashes, from bottom to top."
+    )
+
+
+class PublicKey(BaseModel):
+    """The public key for the Generalized XMSS scheme."""
+
+    model_config = ConfigDict(frozen=True, arbitrary_types_allowed=True)
+    root: List[Fp] = Field(..., max_length=HASH_LEN_FE, min_length=HASH_LEN_FE)
+    parameter: List[Fp] = Field(
```
b-wagn: maybe it makes sense to define a type for Parameter?

src/lean_spec/subspecs/xmss/structures.py
```diff
@@ -0,0 +1,52 @@
...
+        ..., max_length=PARAMETER_LEN, min_length=PARAMETER_LEN
+    )
+
+
+class Signature(BaseModel):
+    """A signature in the Generalized XMSS scheme."""
+
+    model_config = ConfigDict(frozen=True, arbitrary_types_allowed=True)
+    path: HashTreeOpening
+    rho: List[Fp] = Field(..., max_length=RAND_LEN_FE, min_length=RAND_LEN_FE)
```
b-wagn: type for encoding randomness?

b-wagn (commented):
This is an initial review. I will probably do another one later.

src/lean_spec/subspecs/xmss/interface.py
tcoratger: done

src/lean_spec/subspecs/xmss/interface.py
tcoratger: Yep I wanted initially to do something ultra minimalistic but this is better you are right

src/lean_spec/subspecs/xmss/interface.py
tcoratger: Yes indeed I think it's better to ignore this for the Python spec because generating randomness in Python is much simpler than in Rust and therefore there isn't really an associated type. I think for the sake of the spec it's simpler to ignore it and refer to the Rust implementation for this specific.

src/lean_spec/subspecs/xmss/structures.py
tcoratger: fixed

src/lean_spec/subspecs/xmss/structures.py
tcoratger: fixed

src/lean_spec/subspecs/xmss/structures.py
tcoratger: fixed

src/lean_spec/subspecs/xmss/constants.py
tcoratger: Yep good idea, just done

src/lean_spec/subspecs/xmss/interface.py
tcoratger: Ok that's fine, maybe once done, we can just modify the doc here to point to this.

src/lean_spec/subspecs/xmss/interface.py
tcoratger: added

src/lean_spec/subspecs/xmss/interface.py
b-wagn: Makes sense 👍

b-wagn: But maybe we can add a comment that this algorithm is randomized, and for verify that it is deterministic? wdyt?

src/lean_spec/subspecs/xmss/structures.py
```diff
@@ -0,0 +1,69 @@
+"""Defines the data structures for the Generalized XMSS signature scheme."""
+
+from typing import List
+
+from pydantic import BaseModel, ConfigDict, Field
+
+from ..koalabear import Fp
+from .constants import HASH_LEN_FE, PARAMETER_LEN, RAND_LEN_FE
+
+HashDigest = List[Fp]
```
b-wagn: is it possible to add lengths to it?

src/lean_spec/subspecs/xmss/structures.py
```diff
@@ -0,0 +1,69 @@
...
+
+class Signature(BaseModel):
+    """A signature in the Generalized XMSS scheme."""
+
+    model_config = ConfigDict(frozen=True, arbitrary_types_allowed=True)
+    path: HashTreeOpening
+    rho: Randomness = Field(
+        ..., max_length=RAND_LEN_FE, min_length=RAND_LEN_FE
+    )
+    hashes: List[List[Fp]]
```
b-wagn: should probably be List[HashDigest], right?

tcoratger: oups fixed

src/lean_spec/subspecs/xmss/structures.py
tcoratger: yep, just added a stronger typing

src/lean_spec/subspecs/xmss/interface.py
tcoratger: yep, added

PR #11, opened 2025-08-20
subspec: add chain model

tcoratger:
Related https://github.com/leanEthereum/leanSpec/pull/8


unnawut (approved):
Thanks so much for trailblazing this!

PR #15, opened 2025-08-25
feat: add devnet0 STF for chain-ing and 3sf mini

g11tech:
## 🗒️ Description
lean chain STF

WIP


docs/client/chain.md
```diff
@@ -1,4 +1,40 @@
...
+
+<!-- mdformat-toc end -->
+
+## Introduction
+
+This document specs the behavior and functionality of the lean chain. This is a minimal chain spec to prepare the lean clients to be ready for post quantum signature devnets with the following features:
+
+| Devnet | Purpose                    | ETA |
+| ------ | -------                    | :-: |
+| 0      | pq signature preparation   | Oct 2025 |
```
unnawut:
```suggestion
| 0      | minimal consensus          | Sep 2025 |
| 1      | pq signature               | Oct 2025 |
```

docs/client/chain.md
```diff
@@ -1,4 +1,40 @@
...
+
+
+### Devnet 0 Functionality
+
+1. Chain-ing
+  This is beacon style chaining via latest_block_header in state verified against parent hash of the new block. LMD-Ghost (without application of beacon chain style filter block tree)
```
unnawut:
```suggestion
  This is beacon style chaining via latest_block_header in state verified against parent hash of the new block. RLMD-Ghost (without application of beacon chain style filter block tree)
```

I'm not too sure I need to digest a bit more, but I think there's a [max 3 blocks look back with RLMD-GHOST](https://github.com/ethereum/research/blob/master/3sf-mini/p2p.py#L141-L149) that differs from LMD-GHOST?

docs/client/chain.md
```diff
@@ -17,3 +53,225 @@
...
+It is invoked with state as input `get_justifications(state)`fetches cached map for justifications under consideration. If not cached the map is to be constructed from the flattened data from state:
+ - `state.justifications_roots` which is a list of block roots that are under voting consideration
+ - `state.justifications_validators` which are flattened list of validator ids which have voted for the respective roots
+
+ Note that in prover, one would need to construct this map on the fly inside zkVMs.
+
+### State mutators
+
+#### `set_justifications`
+
```
unnawut, line 103: missing `set_justifications` details

docs/client/chain.md
```diff
@@ -17,3 +53,225 @@
...
+## Genesis
+
+The genesis for lean devnets especially `Devnet0` is rather simple. This can even be generated by clients locally:
+Let `genesis_state = generate_genesis_state(genesis_time, num_validators)`
+where:
+ - `genesis_time` is the configured time for the genesis slot
+ - `num_validators` is the configured number of validators to run the chain with
+
+ Even though `Devnet0` has no individual validators tracking, there would also be no need for `genesis_validators_root` going further as well because the state and hence state root would already encode the genesis validator's array.
```
unnawut, line 149: Not sure what's the implication of this

docs/client/chain.md
```diff
@@ -17,3 +53,225 @@
...
+### Genesis state
+
+```python
+def generate_genesis_state(genesis_time: uint64, num_validators: uint64) -> State {
+  state = State(
+    config=Config(
+      genesis_time=genesis_time,
+      num_validators=num_validators,
+    ),
+    latest_block_header=BlockHeader(body_root=hash_tree_root(BeaconBlockBody())),
```
unnawut:
```suggestion
    latest_block_header=BlockHeader(body_root=hash_tree_root(BlockBody())),
```

docs/client/chain.md
```diff
@@ -17,3 +53,225 @@
...
+  );
+
+  return state;
+}
+```
+
+### Genesis block
+
+Let `genesis_block = Block(state_root=hash_tree_root(genesis_state))`.
```
unnawut, line 169: Should this follow 3SF-mini.py exactly? https://github.com/ethereum/research/blob/master/3sf-mini/test_p2p.py#L120-L130

```python
    genesis_block = Block(slot=1, parent=ZERO_HASH)
    genesis_state = State(
        latest_finalized_hash=ZERO_HASH,
        latest_finalized_slot=0,
        latest_justified_hash=ZERO_HASH,
        latest_justified_slot=0,
        historical_block_hashes=[ZERO_HASH],
        justified_slots=[True],
        config=Config(num_validators=NUM_STAKERS)
    )
    genesis_block.state_root = compute_hash(genesis_state)
```

docs/client/chain.md
```diff
@@ -17,3 +53,225 @@
...
+        process_slot(state)
+        state.slot = Slot(state.slot + 1)
+```
+
+```python
+def process_slot(state: BeaconState) -> None:
+    # Cache latest block header state root
+    if state.latest_block_header.state_root == Bytes32():
+        previous_state_root = hash_tree_root(state)
+        state.latest_block_header.state_root = previous_state_root
```
unnawut, line 211: Does `state.latest_block_header.slot` need to be updated to `state.slot` too?

docs/client/chain.md
```diff
@@ -1,4 +1,40 @@
...
+<!-- mdformat-toc start --slug=github --no-anchors --maxlevel=6 --minlevel=2 -->
+
+- [Introduction](#introduction)
+- [Configuration](#Configuration)
+  - [Time parameters](#Time parameters)
+- [Presets](#Presets)
+- [Genesis](#genesis)
+  - [Genesis state](#genesis-state)
+  - [Genesis block](#genesis-block)
+- [STF](#STF)
```
syjn99: Those contents part can be regenerated with `mdformat`.

docs/client/chain.md
```diff
@@ -17,3 +53,225 @@
...
+  state = State(
+    config=Config(
+      genesis_time=genesis_time,
+      num_validators=num_validators,
+    ),
+    latest_block_header=BlockHeader(body_root=hash_tree_root(BeaconBlockBody())),
+  );
+
+  return state;
+}
```
syjn99:
```suggestion
def generate_genesis_state(genesis_time: uint64, num_validators: uint64) -> State {
  state = State(
    config=Config(
      genesis_time=genesis_time,
      num_validators=num_validators,
    ),
    latest_block_header=BlockHeader(body_root=hash_tree_root(BlockBody())),
  )

  return state
}
```

We don't need trailing semi-colons, and fixed `BeaconBlockBody` -> `BlockBody`

docs/client/chain.md
```diff
@@ -17,3 +53,225 @@
...
+        state_root=Bytes32(),  # Overwritten in the next process_slot call
+        body_root=hash_tree_root(block.body),
+    )
+```
+
+#### Operations
+
+```python
+def process_operations(state: State, body: BlockBody) -> None:
+    # process attestations/votes
```
syjn99: I think we often uses attestations <-> votes for the same meaning, and it's quite misleading for me. I'd like to suggest to consolidate all wordings into `Vote` as it is much intuitive and matches with original 3SF paper and 3SF-mini.

Choosing `Attestation` can be beneficial for consistency with current consensus-specs, but I prefer `Vote` in the context of the upcoming devnets.

docs/client/chain.md
g11tech: we will update this once forkchoice specs are presented & finalilized

docs/client/chain.md
g11tech: yes we should add the actual python code for both `get_justifications` and `set_justifications`

docs/client/chain.md
unnawut: Ok so just that we don't need `genesis_validators_root`

docs/client/chain.md
```diff
@@ -17,3 +54,225 @@
...
+
+```python
+def generate_genesis_state(genesis_time: uint64, num_validators: uint64) -> State {
+  state = State(
+    config=Config(
+      genesis_time=genesis_time,
+      num_validators=num_validators,
+    ),
+    latest_block_header=BlockHeader(body_root=hash_tree_root(BlockBody())),
+  );
```
g11tech:
```suggestion
  )
```

docs/client/chain.md
g11tech: we have modified 3sm mini slightly to have genesis block at slot 0 itself like the beacon specification, the resulting differences to STF have also been noted in the STF code in comments

docs/client/chain.md
unnawut: following beacon specs

docs/client/chain.md
```diff
@@ -17,3 +54,225 @@
...
+    # to correctly set genesis block root as already justified and finalized. 
+    # This is not possible at the time of genesis state generation and are set at 
+    # zero bytes because genesis block is calculated using genesis state causing a 
+    # circular dependancy
+    if(state.latest_block_header.slot == 0)
+        # block.parent_root is the genesis root
+        state.latest_justified.root = block.parent_root
+        state.latest_finalized.root = block.parent_root
+
+    num_empty_slots = block.slot - state.latest_block_header.slot -1;
```
g11tech: need to check if this is just -1 or -2, because historical_block_hashes and justified_slots need to be filled to the parent only

docs/client/chain.md
```diff
@@ -17,3 +54,225 @@
...
+    while(num_empty_slots > 0){
+      state.historical_block_hashes.push(ZERO_HASH)
+      state.justified_slots.push(False)
```
g11tech:
```suggestion
      state.justified_slots.push(False)
      num_empty_slots--
```

docs/client/chain.md
g11tech: makes sense, lets just go with Attestation and also reflect the name change in the containers, even though here it means unaggregated votes.

docs/client/chain.md
g11tech: -1 is just correct that gives up empty slots after the parent of this block but ancestor to this block slot, and for each of those ancestor slots we fill up historiical blockhash with zero hash and justified slots with false entries

docs/client/chain.md
unnawut: @g11tech @syjn99 I've just added my pseudo code for `get_justifications()` and `set_justifications()` directly into this PR please have a look.

It's pseudo code because it's untested. I'm not sure there's an easy way to test without moving to full python specs.

unnawut: The current code get/set justifications by flattening/reconstructing the map don't work because it doesn't guarantee ordering between reconstructions, I think I need to update this to delete/shift-left method that we're using in ream instead...

g11tech: just sort the keys and store them in-order

docs/client/chain.md
```diff
@@ -17,3 +54,273 @@
...
+where:
+ - `genesis_time` is the configured time for the genesis slot
+ - `num_validators` is the configured number of validators to run the chain with
+
+ Even though `Devnet0` has no individual validators tracking, there would also be no need for `genesis_validators_root` going further as well because the state and hence state root would already encode the genesis validator's array.
+
+### Genesis state
+
+```python
+def generate_genesis_state(genesis_time: uint64, num_validators: uint64) -> State {
```
syjn99:
```suggestion
def generate_genesis_state(genesis_time: uint64, num_validators: uint64) -> State:
```

Python issue :)

docs/client/chain.md
```diff
@@ -1,4 +1,41 @@
...
+| Devnet | Purpose                    | ETA |
+| ------ | -------                    | :-: |
+| 0      | minimal consensus          | Sep 2025 |
+| 1      | pq signature               | Oct 2025 |
+
+
+### Devnet 0 Functionality
+
+1. Chain-ing
+  This is beacon style chaining via latest_block_header in state verified against parent hash of the new block. LMD-Ghost (without application of beacon chain style filter block tree)
```
syjn99:
```suggestion
  This is beacon style chaining via `latest_block_header` in state verified against parent hash of the new block. LMD-Ghost (without application of beacon chain style filter block tree)
```

image


And also this part looks quite weird in markdown viewer.

docs/client/chain.md
```diff
@@ -17,3 +54,273 @@
...
+      num_empty_slots -= 1;
+
+    # Cache current block as the new latest block
+    state.latest_block_header = BeaconBlockHeader(
+        slot=block.slot,
+        proposer_index=block.proposer_index,
+        parent_root=block.parent_root,
+        state_root=Bytes32(),  # Overwritten in the next process_slot call
+        body_root=hash_tree_root(block.body),
+    )
```
syjn99:
```suggestion
def process_block_header(state: State, block: Block) -> None:
    # Verify that the slots match
    assert block.slot == state.slot
    # Verify that the block is newer than latest block header
    assert block.slot > state.latest_block_header.slot
    # Verify that proposer index is the correct index
    assert block.proposer_index == block.slot % state.config.num_validators
    # Verify that the parent matches
    assert block.parent_root == hash_tree_root(state.latest_block_header)

    # If this was first block post genesis, 3sf mini special treatment is required
    # to correctly set genesis block root as already justified and finalized. 
    # This is not possible at the time of genesis state generation and are set at 
    # zero bytes because genesis block is calculated using genesis state causing a 
    # circular dependancy
    if state.latest_block_header.slot == 0:
        # block.parent_root is the genesis root
        state.latest_justified.root = block.parent_root
        state.latest_finalized.root = block.parent_root

    # now that we can vote on parent, push it at its correct slot index in the structures
    state.historical_block_hashes.push(block.parent_root)
    # genesis block is always justified
    state.justified_slots.push(state.latest_block_header.slot == 0)

    # if there were empty slots, push zero hash for those ancestors
    num_empty_slots = block.slot - state.latest_block_header.slot - 1
    while num_empty_slots > 0:
        state.historical_block_hashes.push(ZERO_HASH)
        state.justified_slots.push(False)
        num_empty_slots -= 1

    # Cache current block as the new latest block
    state.latest_block_header = BlockHeader(
        slot=block.slot,
        proposer_index=block.proposer_index,
        parent_root=block.parent_root,
        state_root=Bytes32(),  # Overwritten in the next process_slot call
        body_root=hash_tree_root(block.body),
    )
```

Python issue :)

and also changed `BeaconBlockHeader` -> `BlockHeader`

docs/client/chain.md
```diff
@@ -17,3 +54,273 @@
...
+
+```python
+def process_operations(state: State, body: BlockBody) -> None:
+    # process attestations
+    process_attestations(state, body.attestations)
+    # other operations will get added as the functionality evolves
+```
+
+```python
+def process_attestations(state: State, attestations: SignedVote[]) -> None:
```
syjn99:
```suggestion
def process_attestations(state: State, attestations: List[SignedVote]) -> None:
```

docs/client/containers.md
```diff
@@ -83,10 +95,23 @@ class Vote(Container):
 
 ```python
 class SignedVote(Container):
```
syjn99, line 112: Hmm how about following structures of `Attestation` of the one in the Beacon spec? (as per this spec is mostly inspired by the Beacon spec) `Vote` is still here, which is quite confusing.

```python
class AttestationData(Container):
    slot: uint64
    head: Checkpoint
    source: Checkpoint
    target: Checkpoint

class Attestation(Container):
    aggregation_bits: Bitlist[VALIDATOR_REGISTRY_LIMIT]
    data: AttestationData
    signature: Bytes32
```

`Attestation` can also be meaningful without aggregation, as it includes `signature`. Until `Devnet2`, `aggregation_bits` will be the `Bitlist` that is only one bit flipped.

docs/client/chain.md
unnawut: Fixed in https://github.com/leanEthereum/leanSpec/pull/15/commits/d03519e2ca32faef7cb97aadac4332de73d973ef and https://github.com/leanEthereum/leanSpec/pull/15/commits/592e65dfee891f615c147609321a37fdb40fac55

Issue #14, opened 2025-08-25
Determine signature size for devnet0


unnawut:
#8 is merged but the signature size still needs to be determined:

- https://github.com/leanEthereum/leanSpec/pull/8/files#r2297420265
- https://github.com/leanEthereum/leanSpec/pull/8/files#r2297420883

I'm now working on https://github.com/b-wagn/hypercube-hashsig-parameters to get the size

PR #16, opened 2025-08-27
feat: add approximate Signature container

unnawut:
## 🗒️ Description

Introduces an _approximate_ Signature container to the client specs.

This is "_approximate_" because while the [size is accurate](https://github.com/leanEthereum/leanSpec/issues/14#issuecomment-3223355674) (`32 * 32 + 28 + 32 * 64 = 3100`), the structure is a simplified version of the [XMSS subspec](https://github.com/leanEthereum/leanSpec/blob/main/src/lean_spec/subspecs/xmss/structures.py#L56-L64) which required `HashTreeOpening`, `HashDigest`, `Randomness`, which is a lot of structure unneeded in devnet0. So I think they're more suitable to be handled in devnet1 where PQ signature will be actually implemented.

Multiple signature types are also not incorporated here because I think it's only relevant for devnet1.

## 🔗 Related Issues or PRs

- Closes https://github.com/leanEthereum/leanSpec/issues/14

### DIFF
--- a/docs/client/containers.md
+++ b/docs/client/containers.md
@@ -17,18 +17,6 @@ class Checkpoint(Container):
     slot: uint64
 ```
 
-## `Signature`
-
-```python
-class Signature(Container):
-    # path: hash_len_fe (8) * bytes_per_fe (4) * log_lifetime (32)
-    path: Vector[Bytes32, 32]
-    # rho: rand_len_fe (7) * bytes_per_fe (4)
-    rho: Bytes28
-    # hashes: hash_len_fe (8) * bytes_per_fe (4) * num_chains (64)
-    hashes: Vector[Bytes32, 64]
-```
-
 ## `State`
 
 ```python
@@ -64,7 +52,6 @@ class Block(Container):
 
 ## `BlockBody`
 
-
 ```python
 class BlockBody(Container):
     votes: List[Vote, VALIDATOR_REGISTRY_LIMIT]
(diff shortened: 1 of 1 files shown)


unnawut:
Following interop call yesterday, I'll update this to a simple `List[byte, 4000]` and we can apply a structure in devnet1

unnawut:
> Following interop call yesterday, I'll update this to a simple `List[byte, 4000]` and we can apply a structure in devnet1

@g11tech @syjn99 please have another look

docs/client/containers.md
```diff
@@ -65,7 +64,7 @@ Remark: `votes` will be replaced by aggregated attestations.
 ```python
 class SignedBlock(Container):
     message: Block,
-    signature: Bytes32,
+    signature: List[byte, 4000],
```
syjn99, line 67:
```suggestion
    signature: ByteList[MAX_SIGNATURE_SIZE],
```

- `ByteList[N]` is just a type alias for `List[byte, N]`. ([Reference](https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md#aliases))
- Seems like number `4000` can be tentative, so maybe we might introduce new preset value?

syjn99 (approved):
Left a single minor comment, and overall OK on my side

docs/client/containers.md
unnawut: > ByteList[N] is just a type alias for List[byte, N]. ([Reference](https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md#aliases))

Yeah so I think `List[byte, N]` is easier to understand and well-tested since it's also used elsewhere in both beacon and lean, while `ByteList` is not.

> Seems like number `4000` can be tentative, so maybe we might introduce new preset value?

Since it'll definitely be replaced in devnet1, I'm thinking it's simpler this way than introducing another reference.

PR #18, opened 2025-08-27
containers: add subspec for containers

tcoratger:
Related to this doc file https://github.com/leanEthereum/leanSpec/blob/main/docs/client/containers.md


src/lean_spec/subspecs/containers/block.py
```diff
@@ -0,0 +1,82 @@
...
+
+from ..chain import config
+from ..types import Bytes32, uint64
+from .vote import Vote
+
+
+class BlockBody(BaseModel):
+    """The body of a block, containing payload data."""
+
+    model_config = ConfigDict(frozen=True, extra="forbid")
```
fselmo: If this will be a common pattern, and I assume it will, we could also create our own base model and inherit these from that. Something like:


`src/leanSpec/types/base.py`
```py
from pydantic import BaseModel, ConfigDict


class StrictBaseModel(BaseModel):
    """A strict, immutable pydantic base model."""

    model_config = ConfigDict(frozen=True, extra="forbid")
```

Then we can let each type implementation just define the fields a bit cleaner. 

I suggest here to put this in a broader `types/` scope as well, rather than keep in subspecs, since I imagine this could be useful across the whole of the project. Take or leave the suggestion 🙂

src/lean_spec/subspecs/types/hash.py
```diff
@@ -0,0 +1,16 @@
+"""Hash Type Specification."""
+
+from pydantic import Field
+from typing_extensions import Annotated
+
+Bytes32 = Annotated[
```
fselmo: I alluded to this in the previous comment but perhaps these more general types, `Bytes32`, `uint64`, etc, should live in a broader scope `src/leanSpec/types` - maybe only the types particular to the subspecs should live inside `/subspecs/types` or perhaps all `types/` can live in broader scope. Just a suggestion, not a blocker.

Also curious on the choice of `uint64` (lowercase) and `Bytes32` (uppercase). Is this because `uint64` was perhaps influenced by the `remerkleable` class and maybe `Bytes32` from `ethereum-types`? 

Doesn't matter too much to me, mostly curious for consistency in reading the code. I'm hopeful that we can consolidate the types across the specs one day as well so as long as it works for you it is good with me 🙂.

src/lean_spec/subspecs/containers/block.py
fselmo: Perhaps `base.py` is better than `.../pydantic.py` since these are all pydantic models in one way or another. I edited accordingly.

tcoratger: Good idea, just applied it :)

src/lean_spec/subspecs/types/hash.py
tcoratger: I completely agree; it's true that it makes more sense for the types to reside at the upper level. I just made the change.

Indeed, regarding the naming, you were right about the influence of the two different repos. I just rename `uint64` to `Uint64`, so we have uppercase everywhere, and this feels more logical to me.

src/lean_spec/subspecs/containers/state.py
```diff
@@ -0,0 +1,64 @@
...
+    """The latest justified checkpoint."""
+
+    latest_finalized: Checkpoint
+    """The latest finalized checkpoint."""
+
+    # Historical data
+    historical_block_hashes: Annotated[
+        list[Bytes32],
+        Field(max_length=DEVNET_CONFIG.historical_roots_limit),
+    ]
```
unnawut, line 39: Super fast work! Similar question to the previous PR on readability, is there a way to represent these 4 lines closer to this?

```python
historical_block_hashes: List[Bytes32, DEVNET_CONFIG.historical_roots_limit]
```

g11tech (approved):
lgtm

src/lean_spec/subspecs/containers/state.py
tcoratger: Yep good idea, just fixed this :) For now, I've made the list only accept a max length, but we can make it more flexible to also have an optional min length as well in the future or create another type for this.

unnawut: Amazing! Is there a reason we need to call it `LeanList` rather than `List`?

tcoratger: Because otherwise this has a conflict (same name) with the other `List` from the typing package. Actually, I just reverted the commits because there were typing issues, I need to think more calmly about a proper thing to do here. I'll do that in a followup PR, this is definitely on my TODO list.

PR #21, opened 2025-08-28
fix(python,uv): Fix issues with local tox use; clarify min python version in README

fselmo:
Minor updates to help debug running things locally. Reported issue:

- Running `uvx --with=tox-uv tox -e all-checks` when the user did not have python >= 3.12 (current requirement for the project) does not yield a very good error message and we don't currently walk through this step in the README.

Updates README and the `uv_python` that `tox-uv` should look for. I believe all a user needs to do is `uv python install 3.12` (or any later version) and this should be resolved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,22 @@ The Lean Ethereum protocol specifications and cryptographic subspecifications.
 
 ### Prerequisites
 
-- Python 3.12 or later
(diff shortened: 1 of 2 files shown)


tcoratger (approved):

tcoratger:
This works perfectly locally, thanks

PR #34, opened 2025-09-06
chore: uint base types and skeleton for ssz

tcoratger:
In this PR:
- I redefined some basic unsigned integers from scratch so that we have all the operations well defined and a complete suite of unit tests to ensure everything is properly done.
- I defined some skeleton for ssz: constants and generalized index. I need to add tests for this but I just wanted to have a generation validation is before proceeding.

@fselmo I think that here your review could be super helpful especially for the uint part so that we have something super clean.


src/lean_spec/types/uint.py
```diff
@@ -0,0 +1,311 @@
...
+
+    def __rrshift__(self, other: Any) -> Self:
+        """Handle the reverse right bit-shift operator (`>>`)."""
+        if not isinstance(other, type(self)):
+            self._raise_type_error(other, ">>")
+        return type(self)(int(other) >> int(self))
+
+    def __eq__(self, other: object) -> bool:
+        """Handle the equality operator (`==`)"""
+        if not isinstance(other, type(self)):
```
fselmo, line 247: I think if we call `Uint64(123) == 123`, it should return `False` rather than raising. Thoughts? If so, then we should make `__ne__()` behave the same way (return `False` instead of raise).

tests/lean_spec/types/test_uint.py
```diff
@@ -0,0 +1,216 @@
...
+def test_instantiation_too_large(uint_class: Type[BaseUint]) -> None:
+    """Tests that instantiating with a value >= MAX raises OverflowError."""
+    max_value = 2**uint_class.BITS
+    with pytest.raises(OverflowError):
+        uint_class(max_value)
+
+
+@pytest.mark.parametrize("uint_class", ALL_UINT_TYPES)
+def test_instantiation_from_float(uint_class: Type[BaseUint]) -> None:
+    """Tests that floats are correctly truncated during instantiation."""
```
fselmo: The biggest question is, do we _want_ truncation? I think this could potentially be a foot-gun. Perhaps we should instead leave the burden on the code to be more explicit in converting floats, rather than doing this internally with pydantic. I'll defer to your discretion here but I'm a bit wary of it.

src/lean_spec/types/uint.py
```diff
@@ -0,0 +1,311 @@
...
+from pydantic_core import core_schema
+from typing_extensions import Self
+
+
+class BaseUint(int):
+    """A base class for custom unsigned integer types that inherits from `int`."""
+
+    BITS: ClassVar[int]
+    """The number of bits in the integer (overridden by subclasses)."""
+
```
fselmo, line 17: We could add something like:

```py
@classmethod
@property
def MAX(cls):
    return 2**self.BITS
```

So that we don't have to manually do this in many places: `Uint64.MAX`, etc. Thoughts?

src/lean_spec/types/uint.py
```diff
@@ -0,0 +1,311 @@
...
+    def __new__(cls, value: SupportsInt) -> Self:
+        """
+        Create and validate a new Uint instance.
+
+        Raises:
+            OverflowError: If `value` is outside the allowed range [0, 2**BITS - 1].
+        """
+        int_value = int(value)
+        if not (0 <= int_value < (2**cls.BITS)):
```
fselmo, line 31: See my comment about accepting `float` truncation. If we decide to disallow it, we need to change this. Also, `int(val)` accepts `string`. I think if we have a string we probably want to leave the burden on the developer to be explicit about this conversion with something like:

```py
int_val = int(val)  # val is a string so convert to int explicitly
uint_64 = Uint64(int_val)
```

I think this should be true for `bool` as well, which `isinstance(False, int)` passes as `True`. It's a bit deceiving but we should probably outright reject these cases for the sake of readability and safety. If we go with this stricter approach, we can do something like:

```py
    def __new__(cls, value: int) -> Self:
        """
        Create and validate a new Uint instance.

        Raises:
            TypeError: If `value` is not an int (rejects bool, string, float).
            OverflowError: If `value` is outside the allowed range [0, 2**BITS - 1].
        """
        if not isinstance(value, int) or isinstance(value, bool):
            raise TypeError(f"Expected int, got {type(value).__name__}")
        if not (0 <= value < 2**cls.BITS):
            raise OverflowError(f"{value} is out of range for {cls.__name__}")
        return super().__new__(cls, value)
```

And we should write unit tests for every case here to make sure only valid `int` is accepted and not anything that can be convert to `int` via `int(value)`. I am curious on your thoughts here though. I figured all of this ultimately just helps readability and explicitness in the specs.

src/lean_spec/types/uint.py
```diff
@@ -0,0 +1,315 @@
...
+        2. Otherwise, validate the input as an integer within the defined bit range
+            and then instantiate the class.
+        3. For serialization (e.g., to JSON), convert the instance to a plain int.
+        """
+        # Validator that takes an integer and returns an instance of the class.
+        from_int_validator = core_schema.no_info_plain_validator_function(cls)
+
+        # Schema that first validates the input as an int, then calls our validator.
+        python_schema = core_schema.chain_schema(
+            [core_schema.int_schema(ge=0, lt=2**cls.BITS), from_int_validator]
```
fselmo:
```suggestion
            [core_schema.int_schema(ge=0, lt=2**cls.BITS, strict=True), from_int_validator]
```

I think we should forbid `string`, `bool`, and `float`. See my comment on the `__new__()` validations. Should help readability and explicitness in specs. Setting `strict=True` here achieves the same thing from the pydantic side. We should have tests for these cases and make sure they are rejected and forced to explicitly convert beforehand.

src/lean_spec/types/uint.py
```diff
@@ -0,0 +1,315 @@
...
+            self._raise_type_error(other, "%")
+        return type(self)(super().__mod__(other))
+
+    def __rmod__(self, other: Any) -> Self:
+        """Handle the reverse modulo operator (`%`)."""
+        if not isinstance(other, type(self)):
+            self._raise_type_error(other, "%")
+        return type(self)(int(other) % int(self))
+
+    def __pow__(self, exponent: Any, modulo: Any | None = None) -> Any:
```
fselmo, line 166: I think it may be too restrictive to raise for certain operations when not the same type. Happy to have my mind changed here but since we validate and return the `type(self)(...)` at the end, perhaps we can allow for `int` to be used as the modulo and the exponent. So we could do `Uint256(2)**2` and get `Uint256(4)`. It feels wrong to me to have to use `Uint256(2) ** Uint256(2)`.

The same is probably true for `__rshift__` and `__lshift__` since we're talking about positions there. Probably best to leave modulo and floordiff strict though?

Thoughts? I'll defer to your preference here. We should validate `int` only for these if so (strict). Perhaps we should abstract out the "is `int` actually an `int`" check into an internal method to be reused for validation across all these methods if we go this route.

src/lean_spec/types/uint.py
```diff
@@ -0,0 +1,315 @@
...
+
+        # Schema that first validates the input as an int, then calls our validator.
+        python_schema = core_schema.chain_schema(
+            [core_schema.int_schema(ge=0, lt=2**cls.BITS), from_int_validator]
+        )
+
+        return core_schema.union_schema(
+            [
+                # Case 1: The value is already the correct type.
+                core_schema.is_instance_schema(cls),
```
fselmo, line 59: I really like the added efficiency with the `is_instance_schema`. And it's still pretty clear and readable what's happening 👌🏼

src/lean_spec/types/uint.py
tcoratger: Mhhh I think that for now this is maybe better to raise an error so that the user should never do that. 

Actually, if we return `False`, I think there might be some tricky and error-prone use cases. For example, `Uint64(123) != 124` would be false, meaning that the user might believe that `Uint64(123)` and `124` are equal if they are not neq... So I wonder if it's not simpler to raise so that the user should simply not do this.

tests/lean_spec/types/test_uint.py
tcoratger: Yep I agree with this, I took this feature from the original `ethereum-types` but this could be prone to errors, so I just modified this to raise an error with a proper message if the user tries to instantiate using a float.

src/lean_spec/types/uint.py
fselmo: Hmm yeah I see your point. It's not very "pythonic" in that sense but that doesn't matter much to me as long as it makes sense to the specs and avoids errors / friction. I'm happy to defer to your discretion here 👍🏼

src/lean_spec/types/uint.py
tcoratger: This sounds like a cool idea, but my IDE gives me this message

```shell
Class properties are deprecated in Python 3.11 and will not be supported in Python 3.13 (Pyright reportDeprecated)
```

Do you think this is normal?

and I think that the linter is not fully happy with the naming here

```shell
N802 Function name `MAX` should be lowercase
  --> src/lean_spec/types/uint.py:20:9
   |
18 |     @classmethod
19 |     @property
20 |     def MAX(cls) -> int:
   |         ^^^
21 |         """The maximum value for this unsigned integer type."""
22 |         return (2**cls.BITS) - 1
   |

Found 1 error.
all-checks: exit 1 (0.02 seconds) /Users/tcoratger/Documents/ethereum/leanSpec> ruff check --no-fix --show-fixes src tests pid=66137
.pkg: _exit> python /Users/tcoratger/Documents/ethereum/leanSpec/.venv/lib/python3.13/site-packages/pyproject_api/_backend.py True hatchling.build
  all-checks: FAIL code 1 (1.58=setup[1.57]+cmd[0.02] seconds)
  evaluation failed :( (1.64 seconds)
```

so I just added this in the following form

```python
@classmethod
    def max(cls) -> int:
        """The maximum value for this unsigned integer type."""
        return int((2**cls.BITS) - 1)
```

is it ok to you?

fselmo: Yep that makes sense to me!

fselmo: I do wonder though if we should return the class type and not `int` here. ~~And perhaps a property might be nice `@property` instead of `@classmethod`?~~

And for any use in the case of, say, `lt` (less than) definition we can do `cls.max.as_int`. ~~The `@property` is more a matter of preference~~ whereas the `max` may actually be better to return as the value with the actual type for most cases. `Uint64.max_value()` could then be used directly instead of `Uint64(Uint64.max_value())`.

src/lean_spec/types/uint.py
tcoratger: yep fully aligned here, just fixed this with the suggestion and the tests accordingly.

src/lean_spec/types/uint.py
tcoratger: Yes you are right, it was overkill and useless here, just fixed according to your idea

src/lean_spec/types/uint.py
fselmo: Actually I prefer leaving as method here to not conflate class / instance usage (edited previous comment). Something like:

```py
    @classmethod
    def max_value(cls) -> Self:
        return cls(2**cls.BITS - 1)
```

I think it's a nice to have. Let me know what you think.

tcoratger: Yes that makes sense just fixed

fselmo: Sorry I was very fickle here 😅. I think the added benefit of your original approach but with the `cls()` (instead of `int`) is that you can use this on the class too: `Uint64.max_value()` instead of only on instances. I was overlooking that benefit.

tcoratger: > Actually I prefer leaving as method here to not conflate class / instance usage (edited previous comment). Something like:
> 
> ```python
>     @classmethod
>     def max_value(cls) -> Self:
>         return cls(2**cls.BITS - 1)
> ```
> 
> I think it's a nice to have. Let me know what you think.

No problem, just fixed according to this :) you have much more experience than me on pythonic things so that I fully trust you on this.

fselmo: Ok, yep, looks good to me! Perhaps let's just add a test for this for all types as a sanity check. Something simple like:

```py
@pytest.mark.parametrize("uint_class", ALL_UINT_TYPES)
def test_max_value(uint_class) -> None:
    assert uint_class.max_value() == uint_class(2**uint_class.BITS - 1)
```

tcoratger: sure, just added this test

PR #35, opened 2025-09-06
tests: Add property-based tests for hypercube using hypothesis

divyaranjan1905:
## 🗒️ Description
This PR adds [hypothesis](https://github.com/HypothesisWorks/hypothesis) to the test group and adds a few property tests to ensure invariants related to the hypercube construction.

All of the invariants have followed from the original paper ([Khovratovich, et. al 2025](https://eprint.iacr.org/2025/889)).

Hypothesis strategies:

- `small_hypercube()`: generates `(w, v)` with 2 ≤ w ≤ 6, 1 ≤ v ≤ 6, and `w**v ≤ 6000`.

- `small_hypercube_with_layer_and_index()`: generates `(w, v, d, x)` where `d` is a layer and `x` is a valid index inside that layer.

- `two_distinct_index()`:  generates `(w, v, d, x1, x2)` with two distinct valid indices in the same layer.

Property tests based on the strategies:

- `test_property_layer_sizes_conservation_monotonicity`: Verifies `prepare_layer_info(w)[v]` return sizes and `prefix_sums` such that:

   - `sum(sizes) == w**v (prefix_sums[-1] == w**v)` layers partition the hypercube.

   - `sizes[d] >= 0` for all `d`.

   - `prefix_sums` is monotone (non-decreasing).

- `test_property_map_vertex_roundtrip`: For random valid (w, v, d, x):

  - `map_to_vertex(w,v,d,x)` returns a v-length vector with coordinates in `[0, w-1]`.

  - The returned vertex lies in layer `d` `((w-1)*v - sum(coords) == d)`.

  - Roundtrip: `map_to_integer(w,v,d,vertex)` (existing function in tests) returns the original `x`.

- `test_property_bijectivity_enum`: For small hypercubes `(w**v ≤ 2000)` exhaustively enumerates all `(d,x)` and asserts `map_to_vertex` produces exactly `w**v` distinct vertices (one-to-one and onto).

- `test_property_map_oob`: Asserts `map_to_vertex(w,v,d,size)` raises ValueError  when index equal to layer size is out-of-bounds.

- `test_property_find_layer_prefix`: For a few representative global indices `(0, 1, total-1, total//2)`, checks `hypercube_find_layer(w,v,x)` returns `(d,r)` consistent with `prefix_sums` (i.e. `prefix_sums[d-1] + r == x`).

- `test_property_layer_injectivity`: Randomized injectivity check within a layer: two distinct `x` values map to distinct vertices.

## ✅ Checklist

    ```console
    uvx --with=tox-uv tox
    ```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 """Pytest configuration and shared fixtures."""
+
 from hypothesis import settings
 
 # Create a profile named "no_deadline" with deadline disabled.
--- a/tests/lean_spec/subspecs/xmss/test_hypercube.py
+++ b/tests/lean_spec/subspecs/xmss/test_hypercube.py
@@ -1,10 +1,13 @@
 """Tests for the hypercube mathematical operations."""
 
+import itertools
 import math
 from functools import lru_cache
-from typing import List
+from typing import Any, Callable, List, Tuple
 
 import pytest
+from hypothesis import assume, given, settings
+from hypothesis import strategies as st
 
 from lean_spec.subspecs.xmss.hypercube import (
     MAX_DIMENSION,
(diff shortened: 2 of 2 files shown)


tcoratger (approved):
Thanks a lot for this, I like the tests, please @fselmo can you just check briefly from a Python point of view, I've never used hypothesis but this seems to be the appropriate tool to do this. If you confirm, I'll merge :) 

This is good to add these tests for the top of the hypercube encoding here. I would suggest not spending too much time on the specific "top of the hypercube" encoding file because this is something that could change in the future due to the recent observations we had with the minimal zkvm and this particular encoding.

But of course, more than happy to apply this to others areas of the code as well.

PR #36, opened 2025-09-08
Feat/lean rainbowstaking

vlladin:
#### This PR introduces the concept of the Staker and the roles it can assume. It is intended as a discussion starter and a foundation to build on top.

The main idea behind these roles is to give a Staker the ability to activate one or more roles at any given time. These roles can be: Attester, Includer, Proposer. They are currently defined in [src/lean_spec/subspecs/staker/role.py](https://github.com/leanEthereum/leanSpec/compare/main...gorondan:leanSpec:feat/lean-rainbowstaking?expand=1#diff-c7e78de21022d10b8eae91edf167e324ed00d76e27d14ce3b0a3f1c6fc29dd46). I also mocked some parameters I expect will be useful moving forward.

I’ve added a subspec called staker (though it might need renaming to rainbow staking) where the roles and staker settings are defined, along with a new Staker container that uses them. There is no real implementation of the roles yet (so no tests). I also added some types needed by the roles (e.g., Gwei, Epoch). I assume these types will need to be expanded anyway, so I went ahead and started on that.

One discussion point I want to raise is the separation of state between roles. For example, the balance of a Staker could be global (shared between roles) or separated for each role. I’m not sure what the best approach is, so for now I assumed some form of separation (also needed for staking/delegation).

Currently, these roles connect to the Staker via [src/lean_spec/subspecs/staker/settings.py](https://github.com/leanEthereum/leanSpec/compare/main...gorondan:leanSpec:feat/lean-rainbowstaking?expand=1#diff-baa8e2328dfced0cf2098c66f18acfde683187e8bc86c31a89a6529ea81a11c8). Stakers are now part of the State via the Staker container.

Getting these roles set up is necessary to advance the Rainbow Staking specs. As discussions and development begin to crystallize, I will update the docs and tests as needed.

## ✅ Checklist

    ```console
    uvx --with=tox-uv tox
    ```

### DIFF
--- a/src/lean_spec/subspecs/containers/staker.py
+++ b/src/lean_spec/subspecs/containers/staker.py
@@ -1,18 +1,24 @@
 """
 Staker Container
 
-Lean Consensus participants are part of a unified pool of stakers, as described in this [3TS design proposal](https://ethresear.ch/t/three-tier-staking-3ts-unbundling-attesters-includers-and-execution-proposers/21648/1).
+Lean Consensus participants are part of a unified pool of stakers,
+as described in this [3TS design proposal](
+https://ethresear.ch/t/three-tier-staking-3ts-unbundling-attesters-includers
+-and-execution-proposers/21648/1).
 
 Each slot, the **Staker** chooses from three different available roles:
 - Attester
 - Includer
 - (Execution) Proposer
 
-Each staker explicitly opts into the role(s) that they wish to take as protocol participants.
-One, two, or all three roles can be chosen, based on the stakers preferences and level of sophistication.
+Each staker explicitly opts into the role(s) that they wish to take as
+protocol participants.
+One, two, or all three roles can be chosen, based on the stakers preferences
+and level of sophistication.
 Mixing and matching multiple roles is possible, under certain constraints.
 
-Each role can be delegated to a target staker and each role can be set as delegatable for other stakers to execute as operators.
+Each role can be delegated to a target staker and each role can be set as
+delegatable for other stakers to execute as operators.
 The staker can update the staking configuration at any time.
 """
 
@@ -30,7 +36,8 @@ class Staker(StrictBaseModel):
         list[StakerSettings],
         Field(min_length=3, max_length=3),
     ]
-    """The list contains the settings for each of the roles the staker can activate."""
+    """The list contains the settings for each of the roles the staker can
+    activate."""
 
     attester_role: AttesterRole
     """
@@ -45,7 +52,8 @@ class Staker(StrictBaseModel):
     Contains the state related to the Includer role.
 
     This role upholds censorship resistance by creating inclusion lists (ILs)
-    that constrain block producers. This field tracks all inclusion-specific data.
+    that constrain block producers. This field tracks all inclusion-specific
+    data.
     """
 
     proposer_role: ProposerRole
--- a/src/lean_spec/subspecs/staker/config.py
+++ b/src/lean_spec/subspecs/staker/config.py
@@ -1,6 +1,7 @@
 """This file defines the parameters used for the Staker and staking."""
 
 from typing_extensions import Final
+
 from lean_spec.types import StrictBaseModel, Uint64
 
 DELEGATIONS_REGISTRY_LIMIT: Uint64 = Uint64(2**12)


src/lean_spec/subspecs/containers/staker.py
```diff
@@ -0,0 +1,26 @@
+"""Staker Container"""
```
tcoratger: Here in the header can we put a 5 lines (IA generated no problem) summary of this 3TS proposal https://ethresear.ch/t/three-tier-staking-3ts-unbundling-attesters-includers-and-execution-proposers/21648/1 as I imagine this is where this comes from? 

We could also put the link or multiple links if needed for the reader to understand here what we are talking about.

I suspect that this container will be able to evolve over time but this is good to have an updated summary of what we are doing each time and update the description as this evolves.

src/lean_spec/subspecs/containers/staker.py
```diff
@@ -0,0 +1,26 @@
...
+from lean_spec.types import StrictBaseModel
+
+
+class Staker(StrictBaseModel):
+    """The consensus staker object."""
+
+    role_config: Annotated[
+        list[StakerSettings],
+        Field(min_length=3, max_length=3),
+    ]
```
tcoratger, line 38: We should be able to use `List` or `Vector` now that we have them.

See https://github.com/leanEthereum/leanSpec/blob/main/src/lean_spec/types/collections.py and usage over the codebase.

src/lean_spec/subspecs/containers/staker.py
```diff
@@ -0,0 +1,26 @@
...
+    """The list contains the settings for each of the roles the staker can activate."""
+
+    attester_role: AttesterRole
+    """Represents the state of the Attester Role."""
+
+    includer_role: IncluderRole
+    """Represents the state of the Includer Role."""
+
+    proposer_role: ProposerRole
+    """Represents the state of the Proposer Role."""
```
tcoratger:
```suggestion
    attester_role: AttesterRole
    """
    Contains the state related to the Attester role.
    
    This role is responsible for providing economic security by voting on the
    validity of blocks. This field tracks all attestation-specific data.
    """

    includer_role: IncluderRole
    """
    Contains the state related to the Includer role.
    
    This role upholds censorship resistance by creating inclusion lists (ILs)
    that constrain block producers. This field tracks all inclusion-specific data.
    """

    proposer_role: ProposerRole
    """
    Contains the state related to the Execution Proposer role.
    
    This role focuses on performance by building and proposing valuable
    execution blocks, including transaction ordering and MEV extraction.
    """
```

src/lean_spec/subspecs/containers/state.py
```diff
@@ -60,6 +61,12 @@ class State(StrictBaseModel):
     ]
     """A bitlist of validators who participated in justifications."""
 
-    def is_proposer(self, validator_index: ValidatorIndex) -> bool:
+    stakers: Annotated[
+        list[Staker],
+        Field(max_length=DEVNET_CONFIG.staker_registry_limit),
+    ]
```
tcoratger: Same here you should be able to use `List` instead now

src/lean_spec/subspecs/staker/config.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class _StakerConfig(BaseModel):
+    """
+    A model holding the canonical, immutable configuration constants
+    for the Staker and staking.
+    """
+
+    # Configuration to make the model immutable.
+    model_config = ConfigDict(frozen=True, extra="forbid")
```
tcoratger:
```suggestion
class _StakerConfig(StrictBaseModel):
    """
    A model holding the canonical, immutable configuration constants
    for the Staker and staking.
    """
```

src/lean_spec/subspecs/staker/config.py
```diff
@@ -0,0 +1,27 @@
+"""This file defines the parameters used for the Staker and staking."""
+
+from pydantic import BaseModel, ConfigDict
+from typing_extensions import Final
+
+from lean_spec.types import Uint64
+
+DELEGATIONS_REGISTRY_LIMIT: Final = 2**12
+"""The maximum number of delegations that can be stored in the state, per staker role."""
```
tcoratger:
```suggestion
DELEGATIONS_REGISTRY_LIMIT: Uint64 = Uint64(2**12)
"""The maximum number of delegations that can be stored in the state, per staker role."""
```

src/lean_spec/subspecs/staker/role.py
```diff
@@ -0,0 +1,134 @@
...
+    slashed: bool
+    """Determines if the staker has been slashed for this role."""
+
+    staker_quota: Uint64
+    """The quota of the staker's role in the delegation."""
+
+    delegations_quotas: Annotated[
+        list[Uint64],
+        Field(max_length=DEVNET_STAKER_CONFIG.delegations_registry_limit),
+    ]
```
tcoratger: Same we should be able to use `List`

src/lean_spec/subspecs/staker/role.py
```diff
@@ -0,0 +1,134 @@
...
+    """The quotas of each delegated balance for this role. This list is
+    parallel with the stakers list from the state."""
+
+    delegated_balances: Annotated[
+        list[Uint64],
+        Field(max_length=DEVNET_STAKER_CONFIG.delegations_registry_limit),
+    ]
```
tcoratger: List also

src/lean_spec/subspecs/staker/role.py
```diff
@@ -0,0 +1,134 @@
...
+    """The delegated balances for each staker. This list is parallel with
+    the stakers list from the state."""
+
+    total_delegated_balance: Gwei
+    """This is the sum of every value in `delegated_balances` and is used
+    for performance optimisation purposes."""
+
+
+class IncluderRole(BaseModel):
```
tcoratger: Should be a strictbasemodel as well no?

src/lean_spec/subspecs/staker/role.py
```diff
@@ -0,0 +1,134 @@
...
+    balance: Gwei
+    """The actual balance of the staker's role."""
+
+    staker_quota: Uint64
+    """The quota of the staker's role in the delegation."""
+
+    delegations_quotas: Annotated[
+        list[Uint64],
+        Field(max_length=DEVNET_STAKER_CONFIG.delegations_registry_limit),
+    ]
```
tcoratger: List also

src/lean_spec/subspecs/staker/role.py
```diff
@@ -0,0 +1,134 @@
...
+    """The quotas of each delegated balance for this role. This list is
+    parallel with the stakers list from the state."""
+
+    delegated_balances: Annotated[
+        list[Uint64],
+        Field(max_length=DEVNET_STAKER_CONFIG.delegations_registry_limit),
+    ]
```
tcoratger: List

src/lean_spec/subspecs/staker/role.py
```diff
@@ -0,0 +1,134 @@
...
+    """The delegated balances for each staker. This list is parallel with
+    the stakers list from the state."""
+
+    total_delegated_balance: Gwei
+    """This is the sum of every value in `delegated_balances` and is used
+    for performance optimisation purposes."""
+
+
+class ProposerRole(BaseModel):
```
tcoratger: strict base model no?

tcoratger: Proposer and attester seem to have the same fields, shouldn't we create a meta class and them have inheritance for this to simplify and avoid duplicated code?

src/lean_spec/subspecs/staker/role.py
```diff
@@ -0,0 +1,134 @@
...
+        list[Uint64],
+        Field(max_length=DEVNET_STAKER_CONFIG.delegations_registry_limit),
+    ]
+    """The quotas of each delegated balance for this role. This list is
+    parallel with the stakers list from the state."""
+
+    delegated_balances: Annotated[
+        list[Uint64],
+        Field(max_length=DEVNET_STAKER_CONFIG.delegations_registry_limit),
+    ]
```
tcoratger: List could be used here as well

src/lean_spec/subspecs/staker/settings.py
```diff
@@ -0,0 +1,29 @@
...
+
+from pydantic import BaseModel
+
+from lean_spec.types import Bytes1, StakerIndex, Uint64
+
+
+class StakerSettings(BaseModel):
+    """Parameters for the StakerSettings."""
+
+    role_identifier: Bytes1
```
tcoratger: Should be Byte no? https://github.com/leanEthereum/leanSpec/blob/main/src/lean_spec/types/byte.py

src/lean_spec/types/bytes.py
```diff
@@ -0,0 +1,16 @@
...
+    bytes,
+    Field(
+        min_length=1,
+        max_length=1,
+        description="A Byte1 type.",
+    ),
+]
+"""
+A type alias for a 1 byte value.
+"""
```
tcoratger: I think that we don't need this since we have https://github.com/leanEthereum/leanSpec/blob/main/src/lean_spec/types/byte.py

src/lean_spec/subspecs/staker/role.py
gorondan: Not sure that once design is consolidated the two will still have the same fields, I would leave it separate for now

tcoratger: Ok fine for me

src/lean_spec/subspecs/chain/config.py
```diff
@@ -71,9 +71,8 @@
 of approximately 12.1 days.
 """
 
-VALIDATOR_REGISTRY_LIMIT: Final = Uint64(2**12)
-"""The maximum number of validators that can be in the registry."""
-
+STAKER_REGISTRY_LIMIT: Final = 2**12
```
tcoratger: Why this is not a `Uint64` anymore?

src/lean_spec/subspecs/staker/config.py
```diff
@@ -0,0 +1,23 @@
...
+    for the Staker and staking.
+    """
+
+    delegations_registry_limit: Uint64
+
+
+# The Devnet Staker Configuration.
+DEVNET_STAKER_CONFIG: Final = _StakerConfig(
+    delegations_registry_limit=DELEGATIONS_REGISTRY_LIMIT,
+)
```
tcoratger:
```suggestion
DEVNET_STAKER_CONFIG: Final = _StakerConfig(
    delegations_registry_limit=DELEGATIONS_REGISTRY_LIMIT,
)
"""The Devnet Staker Configuration."""
```

tcoratger:
@vlladin Can I close this one for now? 

Seems like stale and rebasing would be a nightmare so that it is better to reopen when needed, what do you think?

PR #58, opened 2025-09-17
docs: use mdformat to format markdown files under `docs/`

syjn99:
## 🗒️ Description

Use `mdformat` and its extensions for better documentation. This PR includes:

- Add `mdformat` as deps.
- New task for `tox` so that CI will automatically catch bad formatted files.
- Format existing files under `docs/`: mostly Python issues.

## 🔗 Related Issues or PRs

N/A

## ✅ Checklist

    ```console
    uvx --with=tox-uv tox
    ```

### DIFF
--- a/docs/client/chain.md
+++ b/docs/client/chain.md
@@ -117,7 +117,9 @@ def set_justifications(state: State, justifications: Dict[str, List[bool]]) -> N
         flattened_justifications.extend(justifications[root])
 
     # Create a new Bitlist with all the flattened votes
-    justifications_validators = Bitlist[HISTORICAL_ROOTS_LIMIT * VALIDATOR_REGISTRY_LIMIT](*flattened_justifications)
+    justifications_validators = Bitlist[HISTORICAL_ROOTS_LIMIT * VALIDATOR_REGISTRY_LIMIT](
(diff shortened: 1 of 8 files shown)


tcoratger:
@fselmo Can you check this one quickly, I think you are much better than me for this?

To be noted that since everything is in Python now, I expect that we will not need the Python code in markdown anymore (to be deleted soon, but let us keep it for now at least for devnet0).

tcoratger:
> @tcoratger
> 
> Noted, I can delete `mdformat-ruff` from the dependencies, so that it only checks the formatting issue in markdown itself. Or are you planning to delete all files in `docs/client`?

No no let us keep as is for now. In the future, what I plan to do is:
- To have in docs/ only summary or notes about important stuffs,
- But no Python code (everything should be inside the lean subspecs).

But for now we will keep the markdown with the code inside so the PR is fine.

fselmo (approved):
Yes this seems fine to me if it helps. But it seems like the plan is to remove the markdown eventually, right? For me, I think whatever will help development seems good for now. Don't let this be a blocker at all.

> But for now we will keep the markdown with the code inside so the PR is fine.

Agreed, sgtm 🙂

PR #61, opened 2025-09-19
Minor updates to tests; parallelize tests in CI

fselmo:
## 🗒️ Description

- Parallelize tests in CI to help with speeds now that we have some heavier tests. Since we have mostly fast tests and only a handful of slow tests, use `worksteal` distribution for workers so idle workers can pick up tests that are lagging behind a slow run. This seems to be the best `dist` approach for us at the moment.
- Mark the slowest tests (I used roundabout >10s as the marker here) in case they want to be skipped locally with pytest marker `-m "not slow"`
- Show the slowest 10 test durations in the CI test run with their times displayed (`--durations=10`)
- Updated README.md

## ✅ Checklist

    ```console
    uvx --with=tox-uv tox
    ```

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -58,7 +58,7 @@ commands = mdformat docs/
 description = Run tests (pytest)
 extras = test
 # parallelize with `auto` but keep max processes reasonable for CI
-commands = pytest {posargs} tests -n auto --maxprocesses=10 --durations=10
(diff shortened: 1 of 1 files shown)


fselmo:
@tcoratger going to merge this but wanted you to get a peek at the changes here. Reduced all test runs from 2-3 min down to 1 min. Can also run locally `-m "not slow"` to skip slow tests for quicker dev sanity checks. Conversely, can run slow tests with `-m slow`.

PR #64, opened 2025-09-23
refactor: refactor type factories

fselmo:
## 🗒️ Description

@tcoratger I spent a lot of time iterating on ideas here and ultimately the only thing that made sense for me was drawing a line between primitive types that are themselves representative of Python types (these are the `SSZType` definitions: `Bytes`, `Uint`), and those that hold data (these are the `SSZModel` type definitions: `SSZVector`, `SSZList`, etc...). `SSZModel` inherits from `SSZType` and must implement the same methods so this also bodes well when defining more complex types that still remain as ssz types.

I'm not tied to any of this but this makes a lot of sense for me. In many of these collections, we were manually validating which is exactly what pydantic should set up nicely for us. The factory approach to these types is what broke for `mypy` checks. Instead of this approach we can define very explicit class-based types for the specifications that will appropriately validate and also provide quite a nice API to work with.

I think we should strive to implement the principles from our conversations with the formal verification team on defining very explicit, specific types for data inside of spec'd classes and this bodes quite well with this approach. I think this can help be explicit with types even if they are similar in bounds. For example, I've made a decision in this PR that we can definitely discuss, but I moved `state.py` into a `state/` module (directory) and defined  some state-scoped types appropriate for `state.py` within a `state/types.py`.

`state/types.py`
```py
# Maximum number of historical roots to keep
HISTORICAL_ROOTS_LIMIT = 262144


class HistoricalBlockHashes(SSZList):
    """List of historical block root hashes up to historical_roots_limit."""

    ELEMENT_TYPE = Bytes32
    LIMIT = HISTORICAL_ROOTS_LIMIT


class JustificationRoots(SSZList):
    """List of justified block roots up to historical_roots_limit."""

    ELEMENT_TYPE = Bytes32
    LIMIT = HISTORICAL_ROOTS_LIMIT


class JustifiedSlots(BitlistBase):
    """Bitlist tracking justified slots up to historical roots limit."""

    LIMIT = HISTORICAL_ROOTS_LIMIT


class JustificationValidators(BitlistBase):
    """Bitlist for tracking validator justifications (262144^2 limit)."""

    LIMIT = HISTORICAL_ROOTS_LIMIT * HISTORICAL_ROOTS_LIMIT
```

`state/state.py`
```py
class State(Container):
    """The main consensus state object."""

    config: Config
    """The chain's configuration parameters."""

    slot: Slot
    """The current slot number."""

    latest_block_header: BlockHeader
    """The header of the most recent block."""

    # Checkpoints
    latest_justified: Checkpoint
    """The latest justified checkpoint."""

    latest_finalized: Checkpoint
    """The latest finalized checkpoint."""

    historical_block_hashes: HistoricalBlockHashes
    """A list of historical block root hashes."""

    justified_slots: JustifiedSlots
    """A bitfield indicating which historical slots were justified."""

    justifications_roots: JustificationRoots
    """Roots of justified blocks."""

    justifications_validators: JustificationValidators
    """A bitlist of validators who participated in justifications."""
    
    ...
```

I've also worked in the API so that `ssz_model[index]` works just the same, returning the `data` that it holds at that index. I spent quite a bit of time here to make sure that this approach wasn't just going to "work" but would allow us to get rid of all `type: ignore` cases (other than two overrides in the impl itself for `__iter__()` for example which I think is OK for now but worth a look). It also allowed me to remove some `cast` usage.

I made another decision in this PR I'd like to explicitly point our for discussion, which is to coerce types into our pydantic models. This removes some explicitness out of it so I think this is something that is quite important to think about (and perhaps important to not do since these are specs). When we validate inputs to pydantic models, we can coerce types into the expected type. We can do this by making use of the `ELEMENT_TYPE` you had defined previously for collections. For example, in `SSZList`:

simplified copy of `SSZList` (previously was `List` - we should also discuss the naming here, not tied to this 🙂)
```py
class SSZList(SSZModel):
    ELEMENT_TYPE: ClassVar[Type[SSZType]]
    LIMIT: ClassVar[int]

    data: Tuple[SSZType, ...] = Field(default_factory=tuple)

    # ***** this field_validator can coerce the `data` into the `Tuple[SSZType, ...'` we defined above *****
    @field_validator("data", mode="before")
    @classmethod
    def _validate_list_data(cls, v: Any) -> Tuple[SSZType, ...]:
        """Validate and convert input to a tuple of SSZType elements."""
        if not hasattr(cls, "ELEMENT_TYPE") or not hasattr(cls, "LIMIT"):
            raise TypeError(f"{cls.__name__} must define ELEMENT_TYPE and LIMIT")

        # Handle various input types
        if isinstance(v, (list, tuple)):
            elements = v
        elif hasattr(v, "__iter__") and not isinstance(v, (str, bytes)):
            elements = list(v)
        else:
            raise TypeError(f"List data must be iterable, got {type(v)}")

        # Check limit
        if len(elements) > cls.LIMIT:
            raise ValueError(
                f"{cls.__name__} cannot contain more than {cls.LIMIT} elements, got {len(elements)}"
            )

        # Convert and validate each element
        typed_values = []
        for i, element in enumerate(elements):
            if isinstance(element, cls.ELEMENT_TYPE):
                typed_values.append(element)
            else:
                try:
                    typed_values.append(cast(Any, cls.ELEMENT_TYPE)(element))
                except Exception as e:
                    raise ValueError(
                        f"Element {i} cannot be converted to {cls.ELEMENT_TYPE.__name__}: {e}"
                    ) from e

        return tuple(typed_values)
```

These are just some things to keep in mind when reviewing. I did my best to clean this up and perhaps there is more to be done, but I am going to put this up as a "mostly cleaned up" WIP so we can discuss together.

## ✅ Checklist

    ```console
    uvx --with=tox-uv tox
    ```


src/lean_spec/subspecs/containers/block/block.py
```diff
@@ -61,11 +60,7 @@ class SignedBlock(Container):
...
-    """The block data that was signed."""
+    """The block being signed."""
 
     signature: Bytes32
-    """
-    The proposer's signature of the block message.
-
-    Note: Bytes32 is a placeholder; the actual signature is much larger.
-    """
+    """The proposer's signature over the block."""
```
tcoratger: Just because this is for devnet 0, in the future, we will put the real signature type.

```suggestion
    """
    The proposer's signature of the block message.
    Note: Bytes32 is a placeholder; the actual signature is much larger.
    """
```

src/lean_spec/subspecs/containers/block/types.py
```diff
@@ -0,0 +1,13 @@
...
+
+from ..vote import SignedVote
+
+
+# Domain-specific list type for BlockBody
+class Attestations(SSZList):
+    """List of signed votes (attestations) included in a block."""
+
+    ELEMENT_TYPE = SignedVote
+    LIMIT = 4096  # VALIDATOR_REGISTRY_LIMIT
```
tcoratger: Is there a reason why we don't use `VALIDATOR_REGISTRY_LIMIT` directly here to reuse the constant instead of having hardcoded number?

src/lean_spec/types/bitfields.py
```diff
@@ -256,230 +111,115 @@ def encode_bytes(self) -> bytes:
...
-
-Cache for specialized Bitlist classes: key = (base class, limit).
-"""
+            if byte_index < len(data):
+                bit_value = bool((data[byte_index] >> bit_index_in_byte) & 1)
+            else:
+                bit_value = False
+            bits.append(bit_value)
 
+        return cls(data=tuple(bits))
```
tcoratger: Just a small refactor suggestion here:

```python
bits_generator = (
    bool((data[i // 8] >> (i % 8)) & 1)
    for i in range(cls.LENGTH)
)
return cls(data=tuple(bits_generator))
```

This was my fault but I think that the if `byte_index < len(data)`: check is redundant because the initial validation if `len(data) != expected_len`: already guarantees the byte index will be in bounds.

src/lean_spec/types/bitfields.py
```diff
@@ -256,230 +111,115 @@ def encode_bytes(self) -> bytes:
...
+                    typed_values.append(Boolean(element))
+                except Exception as e:
+                    raise ValueError(f"Bit {i} cannot be converted to Boolean: {e}") from e
+
+        return tuple(typed_values)
+
+    def __getitem__(self, key: int | slice) -> Boolean | list[Boolean]:
+        """Get a bit by index or slice."""
+        if isinstance(key, slice):
+            return [self.data[i] for i in range(*key.indices(len(self.data)))]
```
tcoratger: Do we need this? Or can't we go directly to the return because Python's built-in tuple already supports slicing? Not sure

src/lean_spec/types/byte_arrays.py
```diff
@@ -129,166 +137,96 @@ def decode_bytes(cls, data: bytes) -> Self:
...
+        """Return the hash of the bytes."""
+        return hash((type(self), bytes(self)))
 
-        Args:
-            length: Exact number of bytes the specialized type must contain (N ≥ 0).
+    def hex(self, sep: str | bytes = "", bytes_per_sep: SupportsIndex = 1) -> str:
+        """Return the hexadecimal string representation of the underlying bytes."""
+        if sep == "":
+            return bytes(self).hex()
+        return bytes(self).hex(sep, bytes_per_sep)
```
tcoratger:
```suggestion
    def hex(self, sep: str | bytes = "", bytes_per_sep: SupportsIndex = 1) -> str:
        """Return the hexadecimal string representation of the underlying bytes."""
        if sep == "":
            return bytes(self).hex()
        return bytes(self).hex(sep, bytes_per_sep)
```
```suggestion
    def hex(self, sep: str | bytes = "", bytes_per_sep: SupportsIndex = 1) -> str:
        """Return the hexadecimal string representation of the underlying bytes."""
        return super().hex(sep, bytes_per_sep)
```

src/lean_spec/types/byte_arrays.py
```diff
@@ -365,115 +308,62 @@ def decode_bytes(cls, data: bytes) -> Self:
...
         """Return the concatenation of the argument and the byte list."""
         if isinstance(other, (bytes, bytearray)):
-            return bytes(other) + self._b
-        return bytes(other) + self._b
-
-    def __getitem__(self, i: int) -> int:
-        """Return the i-th byte of the byte list."""
-        return self._b[i]
+            return bytes(other) + self.data
+        return bytes(other) + self.data
```
tcoratger: I think this is my fault but here we could directly have (without the if as the two return statements are the same):

```python
return bytes(other) + self.data
```

src/lean_spec/types/collections.py
```diff
@@ -198,202 +136,181 @@ def deserialize(cls, stream: IO[bytes], scope: int) -> Self:
...
+                raise ValueError(f"Scope {scope} is not divisible by element size {element_size}")
+
+            num_elements = scope // element_size
+            if num_elements > cls.LIMIT:
+                raise ValueError(f"Too many elements: {num_elements} > {cls.LIMIT}")
+
+            elements = []
+            for _ in range(num_elements):
+                element = cls.ELEMENT_TYPE.deserialize(stream, element_size)
+                elements.append(element)
```
tcoratger:
```suggestion
            elements = [
                  cls.ELEMENT_TYPE.deserialize(stream, element_size)
                  for _ in range(num_elements)
             ]
```

src/lean_spec/types/collections.py
```diff
@@ -198,202 +136,181 @@ def deserialize(cls, stream: IO[bytes], scope: int) -> Self:
...
-        """Extend the list with an iterable, checking the limit."""
-        typed_values = [cast(Any, self.ELEMENT_TYPE)(v) for v in values]
-        self._check_capacity(len(typed_values))
-        super().extend(typed_values)
-
-    def insert(self, index: SupportsIndex, value: Any) -> None:
-        """Insert an element at an index, checking the limit."""
-        self._check_capacity(1)
-        super().insert(index, cast(Any, self.ELEMENT_TYPE)(value))
+# Old List factory pattern removed, replaced with SSZList
```
tcoratger: I think that we can drop this comment no?

tests/lean_spec/subspecs/ssz/test_hash.py
```diff
@@ -476,6 +549,157 @@ def test_hash_tree_root_list_uint256() -> None:
     assert hash_tree_root(test_list).hex() == expected
 
 
+# Define explicit ByteVector and ByteList types for testing.
+# Using existing Bytes48 class instead of custom ByteVector48
+# class ByteVector48(BaseBytes):
+#     """A fixed-size byte vector of exactly 48 bytes."""
+#     LENGTH = 48
```
tcoratger: We can drop this no?

tests/lean_spec/types/test_bitfields.py
```diff
@@ -7,37 +7,40 @@
...
+    def test_class_creates_specialized_type(self) -> None:
+        """Tests that concrete Bitvector classes have the correct length."""
+
+        class Bitvector8(BitvectorBase):
+            LENGTH = 8
+
+        class Bitvector16(BitvectorBase):
+            LENGTH = 16
+
+        assert Bitvector8 is not Bitvector16  # type: ignore[comparison-overlap]
```
tcoratger: I think that now we can drop this line as well with the new definitions this is not needed anymore to do this assert

tests/lean_spec/types/test_bitfields.py
```diff
@@ -48,143 +51,123 @@ def test_instantiation_success(self) -> None:
...
+    def test_class_creates_specialized_type(self) -> None:
+        """Tests that concrete Bitlist classes have the correct limit."""
+
+        class Bitlist8(BitlistBase):
+            LIMIT = 8
+
+        class Bitlist16(BitlistBase):
+            LIMIT = 16
+
+        assert Bitlist8 is not Bitlist16  # type: ignore[comparison-overlap]
```
tcoratger: Same I think that we don't need to check this anymore no?

tcoratger (commented):
Thanks a lot for this super great work, this is amazing! I've just left a few little comments but the logic and the structure is much better than before so that after fixing these small nits I am in favor of merging and possibly continuing to iterate on it in the future but this is already super good.

Maybe then I'll be able to check in some functions if there are refactorization to be made to have something smoother but that is absolutely not related to this PR.

I love the new base definitions things so that it forces us to strongly type things which is a very good point.

src/lean_spec/subspecs/containers/block/types.py
fselmo: Ah, yes... I think we just need to use `int(VALIDATOR_REGISTRY_LIMIT)` 👍🏼

src/lean_spec/types/byte_arrays.py
fselmo: This isn't as straightforward. I did change this to be more readable. If the separator is at all defined, it raises, and this cannot be `None`. So I am making this just a bit more explicit by checking `if sep is None:`. Is this OK?

fselmo:
I also did a quick rename refactor `{ClassName}Base` -> `Base{ClassName}` for consistency with `BaseUint` and `BaseBytes`.

tcoratger (approved):
Thanks a lot :)

PR #66, opened 2025-09-25
fix: justifications_validators size and use DEVNET_CONFIG

unnawut:
## 🗒️ Description

- Use `DEVNET_CONFIG.historical_roots_limit` instead of declaring `HISTORICAL_ROOTS_LIMIT` separately
- Fix `JustificationValidators` size from `historical_roots_limit * historical_roots_limit` to `historical_roots_limit * validator_registry_limit`


tcoratger (approved):
Thanks

PR #67, opened 2025-09-25
aggregate all block signatures as naive list [devnet-1]

g11tech:
one of the important lean-ing step especially for STF proving considerations as well as to lighten the block network payload

Also the proposer vote isn't casted separate (so validator doesn't emit vote for a validator id if its that validator id's turn to vote)
but is directly bundled with the block because of the OTS scheme limitation

### DIFF
--- a/tests/lean_spec/subspecs/containers/test_state.py
+++ b/tests/lean_spec/subspecs/containers/test_state.py
@@ -503,9 +503,6 @@ def test_generate_genesis(sample_config: Config) -> None:
     assert state.slot == Slot(0)
     # Body root must commit to an empty body at genesis.
     expected_body = BlockBody(attestations=Attestations(data=[]))
-
-    print(f"body_root hex: {state.latest_block_header.body_root.hex()}")
-
     assert state.latest_block_header.body_root == hash_tree_root(expected_body)
     # History and justifications must be empty initially.
     assert not state.historical_block_hashes


src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -86,12 +90,24 @@ def generate_genesis(cls, genesis_time: Uint64, num_validators: Uint64) -> "Stat
         )
 
         # Build the genesis block header for the state.
+        empty_attestations = Attestations(data=[])
+        empty_checkpoint = Checkpoint(root=Bytes32.zero(), slot=Slot(0))
+        proposer_attestation = ProposerAttestationData(
+            target=empty_checkpoint,
```
syjn99:
```python
@classmethod
def default(cls) -> Self:
    """Return a default checkpoint."""
    return cls(root=Bytes32.zero(), slot=Slot(0))
```

Can we define this `default` method on `Checkpoint` so that it can be used in this PR?

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -398,11 +415,18 @@ def process_block(self, block: Block) -> "State":
         state = self.process_block_header(block)
 
         # Process justification votes (attestations).
-        return state.process_attestations(block.body.attestations)
+        raw_attestations: List[
+            Union[ValidatorAttestation, ProposerAttestationData]
```
g11tech: we don't need union here, lets convert ProposerAttestationData to the proposer's ProposerAttestationData (take target, source from there, head is the hashtreeroot of the current block and slot) and append 

(something that I missed doing that properly in markdown)

g11tech: actually we don't need to append it to attestation at all because this attestation is for the current block as head which techincally is in the future of the block, so we treat it as a freshly cast attestation in the forkchoice without having to process it in the state transition function

May be its better to transform the SignedBlock to SignedBlockAndVote :

```python
SignedBlockAndVote
  message:
    block: Block
    proposer_attestation: ValidatorAttestation
  signatures: List[..]
  ```
So First the block is constructed, and then the proposer can locally add block to their forkchoice and also construct attestation (without signing anything)

And proposer signature would be over this message while we can continue to process block and proposer attestation as separate.

This way block can continue to be a bit decoupled from attestation and wouldn't need to mess with most of the control flow and the STF

what say @syjn99 ?

File: docs/client/chain.md [hunk 3]
```diff
@@ -270,8 +271,8 @@ def process_block_header(state: State, block: Block) -> None:
 
 ```python
 def process_operations(state: State, body: BlockBody) -> None:
-    # process attestations
-    process_attestations(state, body.attestations)
+    attestations=[*body.attestations, body.proposer_attestation]
```
unnawut: These two items have differing types I think

body.attestations:

```py
attestations: List[ValidatorAttestation, VALIDATOR_REGISTRY_LIMIT]
```

```py
class ValidatorAttestation(Container):
    validator_id: uint64
    data: AttestationData
```

body.proposer_attestation:

```py
proposer_attestation: ProposerAttestationData
```

```py
class ProposerAttestationData(Container):
    target: Checkpoint
    source: Checkpoint
```

docs/client/forkchoice.md
```diff
@@ -328,15 +329,39 @@ def on_block(store: Store, block: Block) -> None:
...
-    for signed_vote in block.body.attestations:
-        # Add block votes to the onchain known last votes
-        on_attestation(store, signed_vote, True)
+    for i in [0..block.body.attestations.len]:
+        validator_attestation = block.body.attestations[i]
+        signed_attestation = SignedValidatorAttestation(
+            message=validator_attestation,
+            # eventually one would be able to associate and consume an aggregated signature
+            # for individual vote validity with that information encoded in the signature
+            signature=signatures[i]
```
unnawut: Just checking my understanding on this part. So we do `signatures[i]` for now because we're concatenating signatures.

But eventually individual signatures won't be available, and it will be replaced with something like `zk_verify_signature_inclusion(signed_block.signature, i)`?

docs/client/forkchoice.md
```diff
@@ -328,15 +329,39 @@ def on_block(store: Store, block: Block) -> None:
...
     update_head(store)
+
+    # the proposer vote for the current slot and block as head is to be
+    # treated as the vote is independently casted in the second interval
+    proposer_attestation = ValidatorAttestation(
+        slot=block.slot,
+        head=Checkpoint(root=block_hash,slot=block.slot),
+        source=block.body.proposer_attestation.source,
+        target=block.body.proposer_attestation.target,
+    )
```
unnawut: Any reason why we need to build `ValidatorAttestation` as part of `on_block`? Imho building it right into `block.body.proposer_attestation` makes more sense to me in terms of responsibility.

docs/client/forkchoice.md
```diff
@@ -328,15 +329,39 @@ def on_block(store: Store, block: Block) -> None:
...
+    signed_proposer_attestation=SignedValidatorAttestation(
+        message=proposer_attestation,
+        signature=signatures[signatures.len-1]
```
unnawut: I couldn't find where `signatures[signatures.len-1]` is set to be the proposer's signature

src/lean_spec/subspecs/containers/block/block.py
```diff
@@ -4,17 +4,33 @@
...
+
+    Individual signatures live in the aggregated block signature list, so
+    these entries contain only vote data without per-attestation signatures.
     """
-    A list of votes included in the block.
 
-    Note: This will eventually be replaced by aggregated attestations.
+    proposer_attestation: ProposerAttestationData = (
+        _DEFAULT_PROPOSER_ATTESTATION
+    )
```
unnawut: Shouldn't this part be only the container definition? Not sure if we should globally apply a default source & target values here

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -86,12 +90,24 @@ def generate_genesis(cls, genesis_time: Uint64, num_validators: Uint64) -> "Stat
...
         # Build the genesis block header for the state.
+        empty_attestations = Attestations(data=[])
+        empty_checkpoint = Checkpoint(root=Bytes32.zero(), slot=Slot(0))
+        proposer_attestation = ProposerAttestationData(
+            target=empty_checkpoint,
+            source=empty_checkpoint,
+        )
+
+        genesis_body = BlockBody(
+            attestations=empty_attestations,
```
unnawut: Super small nit but I find inlining easier to understand since `Attestations(data=[])` is self explanatory. But your call

```suggestion
            attestations=Attestations(data=[]),
```

docs/client/containers.md
```diff
@@ -84,54 +94,99 @@ class BlockHeader(Container):
...
+    # to be replaced by a single zk aggregated and verifiable signature in a future devnet
+    #
+    # Note that
+    #  i)signature list max is still validator registry limit because of proposer
+    #    attestation has no separate signature.
+    # ii)also note that current proposer signature is just the proposer signature over
+    #    ValidatorAttestation created from its ProposerAttestationData.
+    #   till we can actually consume a block signature for re-packing this block's
+    #   proposer vote by some other future proposer
+    signature: List[Vector[byte, 4000], VALIDATOR_REGISTRY_LIMIT]
```
unnawut: Not sure should we use this now?

```suggestion
    signature: List[Bytes4000, VALIDATOR_REGISTRY_LIMIT]
```

docs/client/forkchoice.md
```diff
@@ -328,15 +329,39 @@ def on_block(store: Store, block: Block) -> None:
     # sync parent chain if not available before adding block to forkchoice
     assert parent_state is not None, "Parent state not found, sync parent chain first"
 
+    valid_signatures = validate_signatures(block.signature)
```
jihoonsong: What was the reason that we don't do `assert validate_signatures(block.signature)` here and make `state_transition` receive `valid_signatures`?

docs/client/containers.md
```diff
@@ -84,54 +94,99 @@ class BlockHeader(Container):
...
 ```
 
-## `Vote`
+## `AttestationData`
 
 Vote is the attestation data that can be aggregated. Although note there is no aggregation yet in `devnet0`.
```
jihoonsong: Nit: there is no `Vote` anymore.

docs/client/forkchoice.md
```diff
@@ -113,8 +113,8 @@ class Store(object):
     latest_finalized: Checkpoint
     blocks: Dict[Root, Block] = field(default_factory=dict)
     states: Dict[Root, State] = field(default_factory=dict)
-    latest_known_votes: Dict[ValidatorIndex, Checkpoint] = field(default_factory=dict)
-    latest_new_votes: Dict[ValidatorIndex, Checkpoint] = field(default_factory=dict)
+    latest_known_votes: Dict[ValidatorIndex, SignedValidatorAttestation] = field(default_factory=dict)
+    latest_new_votes: Dict[ValidatorIndex, SignedValidatorAttestation] = field(default_factory=dict)
```
jihoonsong: Why do these have to have `SignedValidatorAttestation` instead of `Checkpoint`?

docs/client/forkchoice.md
```diff
@@ -279,45 +279,46 @@ def on_tick(store: Store, time: int, has_proposal: bool) -> None:
 #### `on_attestation`
 
 ```python
-def on_attestation(store: Store, signed_vote: SignedVote, is_from_block: bool = False) -> None:
+def on_attestation(store: Store, vote: SignedValidatorAttestation, is_from_block: bool = False) -> None:
```
jihoonsong: Using `signed_vote` seems more aligned with other parts of the spec.

docs/client/containers.md
jihoonsong: I probably missing something so I'm asking a question here. When we aggregate signatures, do we want to aggregate all different kinds of signatures? In other words, what's the best way to understand why we want to aggregate attestation signatures along with block signature?

src/lean_spec/subspecs/containers/block/block.py
```diff
@@ -4,17 +4,33 @@
...
+_DEFAULT_PROPOSER_ATTESTATION = ProposerAttestationData(
+    target=Checkpoint(root=Bytes32.zero(), slot=Slot(0)),
+    source=Checkpoint(root=Bytes32.zero(), slot=Slot(0)),
+)
 
 
 class BlockBody(Container):
     """The body of a block, containing payload data."""
 
     attestations: Attestations
```
jihoonsong, line 14: `Attestations` is a list of signed vote, which is not aligned with `BlockBody`'s definition.

docs/client/containers.md
```diff
@@ -84,54 +94,99 @@ class BlockHeader(Container):
...
-class Vote(Container):
+class AttestationData(Container):
     slot: uint64
     head: Checkpoint
     target: Checkpoint
     source: Checkpoint
 ```
 
-## `SignedVote`
+## `ProposerAttestationData`
```
jihoonsong: While I understand the constraints that made us bring up this approach, I believe the problem of signing multiple times in a slot remains. If we want to adopt this PG signature effort into a production-level protocol, it's highly likely that a proposer needs to sign more than one message in a slot. If we pack all those messages into one, a proposer has to broadcast all of their messages at once. In other words, there is only one deadline within a slot for a proposer. This is not flexible so I'm not sure if this would be enough to meet all the needs of the protocol.

IMO, we could do either find a proper solution for the problem or use a stopgap solution like this one that is just enough for the upcoming few devnets. If we were to go to the latter, current approach is one of solutions. But it seems it adds complexity, which is might not an ideal given that it's not a long-term solution. So I'd like to encourage to find a better one. Few things I have in my mind are:

1. Currently, `BlockBody` has a list of unsigned votes. We can append proposer's attestation at the end of the list and append proposer's signature at the `SignedBlock`'s signature. It's not very clean as there is an exception that proposer's signature is made over a block while others are over attestation. But this is already what we have if I'm not wrong, and this doesn't require to change original `Vote` and related classes nor to add `ProposerAttestationData`.

2. We can add a rule that proposers don't vote. This would make things much simpler.

jihoonsong (commented):
Can you rebase your PR?

docs/client/containers.md
```diff
@@ -52,6 +60,8 @@ class State(Container):
     historical_block_hashes: List[Bytes32, HISTORICAL_ROOTS_LIMIT]
     justified_slots: List[bool, HISTORICAL_ROOTS_LIMIT]
 
+    validators: List[Validator, VALIDATOR_REGISTRY_LIMIT]
```
jihoonsong: Why does `State` need `validators`?

docs/client/forkchoice.md
g11tech: yes

docs/client/chain.md
g11tech: this won't be changed at all, ProposerAttestationData will have no effect on this block's STF because its proposer vote for next block inclusion and for this slot will only be imported in forkchoice

(getting this changed)

docs/client/forkchoice.md
g11tech: at this point in this block slot, the proposer vote is not to be processed like block attestations, and is to be independently imported by the forkchoice like it imports all other validators votes into `new votes`

docs/client/forkchoice.md
g11tech: yes that part is missing in markdown, but since we are now totally moving to code it doesn't matter

docs/client/containers.md
g11tech: to enroll validators pub keys, even though validators don't rotate but we are converging to somewhat how it will look like so all the constraints are out in the open

src/lean_spec/subspecs/containers/block/block.py
g11tech: as per this PR proposal Attestations is list of ValidatorAttestation, without votes 
(eventually as well, as we don't have any signatures in block now, all aggregated away currently as a naive list, but eventually as a lean VM aggregated signature)

jihoonsong: My bad, I was in main branch while comparing something.

docs/client/containers.md
jihoonsong: Where does it used in this PR?

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -439,41 +446,47 @@ def process_attestations(
...
-            if source_slot_int < len(justified_slots):
-                source_is_justified = justified_slots[source_slot_int]
-            else:
-                continue  # Source is too far in the past
+            if source_slot >= len(justified_slots):
+                continue
 
+            source_is_justified = justified_slots[source_slot]
+
+            target_tracked = target_slot < len(justified_slots)
```
g11tech: @GrapeBaBa  lets not make any other changes in this PR so as to keep the focus and scope limited, we can do a followup PR if we want to improve readability

src/lean_spec/subspecs/containers/block/block.py
```diff
@@ -56,14 +57,29 @@ class Block(Container):
     """The block's payload."""
 
 
-class SignedBlock(Container):
-    """A container for a block and the proposer's signature."""
+class BlockAndProposerVote(Container):
```
g11tech: should we go with ProposerVote or ProposerAttestation nomenclature here @syjn99 @unnawut

src/lean_spec/subspecs/networking/messages.py
```diff
@@ -52,11 +52,11 @@ class Status(StrictBaseModel):
 """
 
 BlocksByRootResponse = Annotated[
-    list[SignedBlock],
+    list[SignedBlockAndVote],
```
g11tech: this change is not needed, Vote is packed with the Block only because of signature constrains how ever its processed and packed independently by a future block (hopefully the next)

ahh ok but we will the full message to correctly verify signature, so nevermind

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -106,90 +118,108 @@ class method acts as a factory for creating a new Store instance.
...
         Args:
-            signed_vote: Attestation to validate.
+            signed_attestation: Attestation to validate.
 
         Raises:
             AssertionError: If attestation fails validation.
         """
-        vote = signed_vote.data
+        attestation = signed_attestation.message
+        data = attestation.data
```
g11tech, line 130: @GrapeBaBa add a dummy/noop function to validate signature, currently validate it to be equal to zero bytes

we will integrate actual signature as a followup to this PR separately (so add a comment to this effect)

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -106,90 +118,108 @@ class method acts as a factory for creating a new Store instance.
...
-            # Update new votes if this is latest from validator
-            latest_new = self.latest_new_votes.get(validator_id)
-            if latest_new is None or latest_new.slot < vote.target.slot:
-                self.latest_new_votes[validator_id] = vote.target
+        # Update new votes if this is latest from validator
+        latest_new = self.latest_new_votes.get(validator_id)
+        if latest_new is None or latest_new.message.data.slot < attestation_slot:
+            self.latest_new_votes[validator_id] = signed_attestation
+
+    def _validate_block_signatures(
```
g11tech, line 205: lets also use in process_block even if its dummy so we have a full flow @GrapeBaBa

g11tech: never mind found it

src/lean_spec/subspecs/containers/block/block.py
g11tech: this container is no more

docs/client/forkchoice.md
g11tech: Its fine either way imo as this is spec, clients can choose to optimize/parallelize anywhich way they want but its good to be explicit in the STF that signatures have been validated

docs/client/containers.md
g11tech: will look into it if we wanna remove all vote references,

docs/client/forkchoice.md
g11tech: for block construction purposes, you need signed objects to get signatures you want to aggregate

docs/client/containers.md
g11tech: signatures are heavy now, also with OTS constraint proposer signature which is over the block and vote is sort of anyway not "pure" anymore so to take things to logical conclusion a block (and vote) signature will represent all valid objects in the block (and vote)

docs/client/containers.md
g11tech: correct, that part is to be added

@GrapeBaBa pls modify the generate_genesis to accept a sequence of pubkeys to assign to the validators object

since we haven't actually integrated sig libary yet (to be done in the followup PR), always pass it as a list of  `000` bytes for num_validators length for example from test fixtures

src/lean_spec/subspecs/containers/block/block.py
```diff
@@ -56,14 +57,29 @@ class Block(Container):
     """The block's payload."""
 
 
-class SignedBlock(Container):
-    """A container for a block and the proposer's signature."""
+class BlockAndVote(Container):
```
g11tech: should we go with ...Vote or ..Attestation nomenclature here @syjn99 @unnawut

Attestation could confuse with attestations in the block body

docs/client/containers.md
g11tech: for 
1. yes we are trying to move closer to aggregation so the list there is just a temporary thing till we have leanVM coming up. also till then its just proposer votes signature, so not even a full block & vote signature since we can't validate such a signature without leanVM

2. yes can be considered actually, was in my mind, but eventually signatures will be aggregated

g11tech: infact the signature part can just be totally replaced by full block validity proof including execution

so yes design space is open and we will see how to simplify it if it turns out convoluted

src/lean_spec/subspecs/containers/block/block.py
unnawut: I think there will be questionings here once we start to form production specs, but I'm more than happy to keep going with Vote for single, non-aggregated vote and Attestation for the aggregation. The distinction helps with readability here.

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -86,23 +90,26 @@ def generate_genesis(cls, genesis_time: Uint64, num_validators: Uint64) -> "Stat
...
             config=genesis_config,
             slot=Slot(0),
             latest_block_header=genesis_header,
-            latest_justified=Checkpoint(root=Bytes32.zero(), slot=Slot(0)),
-            latest_finalized=Checkpoint(root=Bytes32.zero(), slot=Slot(0)),
+            latest_justified=Checkpoint.default(),
+            latest_finalized=Checkpoint.default(),
             historical_block_hashes=HistoricalBlockHashes(data=[]),
             justified_slots=JustifiedSlots(data=[]),
+            validators=Validators(data=[]),
```
unnawut, line 109: Should we add validators import function into the specs? Could go as a separate PR though.

src/lean_spec/subspecs/containers/block/block.py
unnawut: I just re-read this PR's `vote.py` again. I strongly prefer `Vote` & `Attestation` over `ValidatorAttestation` & `Attestation`

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -402,7 +409,7 @@ def process_block(self, block: Block) -> "State":
 
     def process_attestations(
         self,
-        attestations: Attestations,
+        attestations: Iterable[ValidatorAttestation],
```
unnawut: Doesn't this change the type meaning quite a bit? Instead of strictly accepting `Attestations` looks like it's accepting any iterable `ValidatorAttestation`. Maybe we don't require this flexibility?

src/lean_spec/subspecs/containers/state/state.py
unnawut: Would really prefer if we could remove the refactors from the PR. Happy to review them as separate PRs though!

src/lean_spec/subspecs/forkchoice/helpers.py
```diff
@@ -15,7 +20,7 @@
 def get_fork_choice_head(
     blocks: Dict[Bytes32, Block],
     root: Bytes32,
-    latest_votes: Dict[ValidatorIndex, Checkpoint],
+    latest_votes: Dict[ValidatorIndex, SignedValidatorAttestation],
```
unnawut: Any reason why we need the signed data here? Doesn't look like we need to re-verify

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -106,90 +118,118 @@ class method acts as a factory for creating a new Store instance.
...
+            signed_attestation: Attestation to validate.
 
         Raises:
             AssertionError: If attestation fails validation.
         """
-        vote = signed_vote.data
+        attestation = signed_attestation.message
+        data = attestation.data
 
-        # Validate vote targets exist in store
```
unnawut, line 123: Would be nice to retain the code comments but maybe it's just me

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -106,90 +118,118 @@ class method acts as a factory for creating a new Store instance.
...
+                self.latest_known_votes[validator_id] = signed_attestation
 
             # Remove from new votes if this supersedes it
             latest_new = self.latest_new_votes.get(validator_id)
-            if latest_new is not None and latest_new.slot <= vote.target.slot:
+            if latest_new is not None and latest_new.message.data.slot <= attestation_slot:
                 del self.latest_new_votes[validator_id]
+            return
 
-        else:
```
unnawut, line 173: refactor in a separate PR would be great

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -106,90 +118,118 @@ class method acts as a factory for creating a new Store instance.
...
+        time_slots = Slot(self.time // SECONDS_PER_INTERVAL)
+        assert attestation_slot <= time_slots, "Attestation from future slot"
+
+        # Update new votes if this is latest from validator
+        latest_new = self.latest_new_votes.get(validator_id)
+        if latest_new is None or latest_new.message.data.slot < attestation_slot:
+            self.latest_new_votes[validator_id] = signed_attestation
+
+    @staticmethod
+    def _is_placeholder_signature(signature: Bytes4000) -> bool:
```
```suggestion
    def _is_valid_signature(signature: Bytes4000) -> bool:
```

Nit and subjective: imo it helps with spec readability to go with the intended name. The `TODO` comment is already self-explanatory that the implementation is not there yet.

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -106,90 +118,118 @@ class method acts as a factory for creating a new Store instance.
...
 
         Adds block to store, processes included attestations, and updates head.
 
         Args:
-            block: Block to process.
+            signed_block_vote: Block to process.
         """
+        envelope = signed_block_vote.message
+        block = envelope.block
+        proposer_attestation = envelope.proposer_attestation
```
```suggestion
        block = signed_block_vote.message.block
        proposer_attestation = signed_block_vote.message.proposer_attestation
```

Nit: I find this clearer, not sure if envelope distinction is needed

src/lean_spec/subspecs/forkchoice/helpers.py
```diff
@@ -66,7 +72,9 @@ def get_fork_choice_head(
         current = max(children, key=lambda x: (vote_weights.get(x, 0), blocks[x].slot, x))
 
 
-def get_latest_justified(states: Dict[Bytes32, "State"]) -> Optional[Checkpoint]:
+def get_latest_justified(
+    states: Dict[Bytes32, "State"],
+) -> Optional[Checkpoint]:
```
unnawut: Are you using a linter that's not part of the repo setup? If so would be great if you could add as a separate PR

src/lean_spec/types/ssz_base.py
```diff
@@ -102,43 +100,49 @@ class SSZModel(StrictBaseModel, SSZType):
...
+    with SSZ serialization. Use this for containers and complex types that
+    can benefit from Pydantic.
 
-    For simple types that need special inheritance (like int), use SSZType directly.
+    For simple types that need special inheritance (like int), use SSZType
+    directly.
 
-    SSZModel provides natural iteration and indexing for collections with a 'data' field:
+    SSZModel provides natural iteration and indexing for collections with a
+    'data' field:
```
unnawut: These should also go in a separate PR. Also seems like a different linter config

tests/lean_spec/subspecs/containers/test_state.py
```diff
@@ -64,33 +69,42 @@ def genesis_state(sample_config: Config) -> State:
...
     -------
-    SignedBlock
-        A signed block wrapper containing the constructed block.
+    Block
+        The constructed block message with attestations embedded.
     """
+    normalized_attestations = []
+    if votes is not None:
+        for attestation in votes:
+            normalized_attestations.append(attestation)
```
unnawut: This seems strange to me. Why do we need to reconstruct the same list? Or I just don't know python enough.

src/lean_spec/subspecs/containers/state/state.py
unnawut, line 109: Also since we are adding `state.validators` we should remove `state.config.num_validators` (or `state.config` altogether since there's no other use)

tests/lean_spec/subspecs/forkchoice/test_attestation_processing.py
```diff
@@ -1,20 +1,21 @@
 """Tests for attestation validation and processing."""
 
-import pytest
+import pytest  # type: ignore[import-not-found]
```
unnawut: Hmm only this file has this issue?

unnawut (commented):
The rebase and cleanup helped a lot. Much thanks!

Still needs a review after https://github.com/leanEthereum/leanSpec/pull/67/files#r2433502691 is decided and all namings updated

jihoonsong:
+1 for making sure this PR has well-defined scope and is self-contained. Some changes that the PR introduces are not aligned with the description, which makes readers confusing.

I don't have a strong preference on vote vs attestation, but I found using attestation for aggregated vote confusing. They are used interchangeably in state machine replication protocols quite often as far as I know.

GrapeBaBa:
@unnawut @jihoonsong Yes, I am reverting the changes which wrong changed by my flask configure, please wait to review as it is submitted by @g11tech so that I can't change it to draft or edit title to WIP

src/lean_spec/subspecs/containers/state/state.py
GrapeBaBa: can we also leave it another PR? it also impacts genesis generator tool.

GrapeBaBa: what do you mean 'validators import function'

src/lean_spec/subspecs/forkchoice/store.py
GrapeBaBa: added

src/lean_spec/subspecs/forkchoice/helpers.py
GrapeBaBa: @g11tech

src/lean_spec/subspecs/containers/block/block.py
jihoonsong: As what we have in a block is an attestation, I'd advocate for `BlockWithAttestation`.

src/lean_spec/subspecs/containers/block/block.py
```diff
@@ -56,14 +57,29 @@ class Block(Container):
...
+    """The proposer's vote corresponding to this block."""
+
+
+class SignedBlockAndVote(Container):
+    """Envelope carrying a block, proposer vote, and aggregated signatures."""
+
+    message: BlockAndVote
+    """The block plus proposer vote being signed."""
+
+    signature: BlockSignatures
```
jihoonsong, line 76: `BlockSignatures` sounds like a list of block signatures but it's not. Introducing `Signature` and `Signatures` seems better represent what it is and is align with other data structures such as `Attestation` and `Attestations`.

`Signatures` would be removed after we have LeanVM.

src/lean_spec/subspecs/containers/vote/vote.py
```diff
@@ -0,0 +1,89 @@
...
+class AttestationData(Container):
+    """Attestation content describing the validator's observed chain view."""
+
+    slot: Slot
+    head: Checkpoint
+    target: Checkpoint
+    source: Checkpoint
+
+
+class ValidatorAttestation(Container):
```
jihoonsong: There is no `ProposerAttestation` anymore so we don't need to `Validator` prefix. This can be just `Attestation`.

src/lean_spec/subspecs/containers/block/block.py
jihoonsong: We could use `AggregatedAttestations` for aggregated attestations. This is straightforward and is in line with `AggregatedSignatures`.

unnawut:
> Can we leave 3 to another PR, for 2 I agree, @g11tech what is your opinion?

Yeah leave to another PR is probably better otherwise the PR is bloated. We'll be adding more PRs before the spec is ready anyway.

I'm reviewing the rest of new changes asap. Thanks!

src/lean_spec/subspecs/containers/vote/vote.py
syjn99: I agree with this, as `ValidatorAttestation` is inducing lots of confusion here. `ValidatorAttestation` itself is ambiguous, thus reading this spec makes us much harder to understand. If we would like to avoid the term like `SingleAttestation` in current beacon spec, we might use `Attestation` and `AggregatedAttestation` as "attestation"s are not associated with aggregation. 

Note: I was trying to find out if "attestation" and "aggregation" are related with each other, but found the posts like [this](https://notes.ethereum.org/@fradamt/chained-3sf#Attestations) that use "attestation" as just a vote.

src/lean_spec/subspecs/containers/vote/vote.py
```diff
@@ -0,0 +1,89 @@
...
+    def data(self) -> ValidatorAttestation:
+        """Expose the message for backwards compatibility with SignedVote."""
+        return self.message
+
+
+class Attestation(Container):
+    """Aggregated attestation consisting of participation bits and message."""
+
+    aggregation_bits: AggregationBits
+    message: AttestationData
```
syjn99:
```suggestion
    data: AttestationData
```

We can stick to `message` when it comes to `Signed*` and `data` for `*Attestation`.

src/lean_spec/subspecs/containers/state/state.py
unnawut: Specifying how to parse validator data from a file

unnawut: But let's not worry about this in this PR

unnawut: sgtm!

src/lean_spec/subspecs/containers/vote/vote.py
g11tech: how do we just call this as `Vote` and let `Attestation` be aggregated attestation sort of like devnet0, I wanted to move away from Vote terminology but seems like that is more intutive

src/lean_spec/subspecs/containers/vote/vote.py
g11tech: yes makes sense actually, since Attestation itself is the message, @GrapeBaBa kindly update

src/lean_spec/subspecs/forkchoice/helpers.py
g11tech: the reason is that we store `SignedValidatorAttestation` maps because you need to pick and pack signed votes for the block (even though we now strip out signatures in one field)

So this is just carry forward of how we store signed votes, earlier spec was not all consistent in this regard

src/lean_spec/subspecs/forkchoice/helpers.py
```diff
@@ -41,10 +46,11 @@ def get_fork_choice_head(
     # Count votes for each block (votes for descendants count for ancestors)
     vote_weights: Dict[Bytes32, int] = {}
 
-    for vote in latest_votes.values():
-        if vote.root in blocks:
+    for attestation in latest_votes.values():
+        target = attestation.message.data.target
```
g11tech: @GrapeBaBa we don't use target here, its `head` part, vote target that is refered in the comment is "head" , kindly rectify

Or we can restore the function to what it was before, but then we will need to filter the dict we pass in the function to extract head as the validator<>checkpoint map

src/lean_spec/subspecs/containers/vote/vote.py
jihoonsong: I don't think `Vote` is the most accurate term here. It refers to attestation specifically. Using `Attestation` as an aggregated attestation is very confusing. Signature doesn't mean an aggregated signatures either. `Attestation` and `AggregatedAttestation` is much more straightforward and is in line with existing names.

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -198,22 +231,54 @@ def process_block(self, block: Block) -> None:
...
 
         # Add block and state to store
         self.blocks[block_hash] = block
         self.states[block_hash] = state
 
         # Process block's attestations as on-chain votes
-        for signed_vote in block.body.attestations:
-            self.process_attestation(signed_vote, is_from_block=True)
+        for index, attestation in enumerate(block.body.attestations):
+            signature = signatures[index] if index < len(signatures) else Bytes4000.zero()
```
g11tech: we should just throw so that process block fails, so I think this is just fine @GrapeBaBa 
```suggestion
            signature = signatures[index]
```

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -198,22 +231,54 @@ def process_block(self, block: Block) -> None:
...
 
         # Update forkchoice head
         self.update_head()
 
+        proposer_signature_index = len(block.body.attestations)
+        proposer_signature = (
+            signatures[proposer_signature_index]
+            if proposer_signature_index < len(signatures)
+            else Bytes4000.zero()
+        )
```
g11tech:
```suggestion
        proposer_signature_index = signatures[len(block.body.attestations)]
```
this is sufficient else we should just throw and process block fails @GrapeBaBa

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -420,25 +485,27 @@ def produce_block(self, slot: Slot, validator_index: ValidatorIndex) -> Block:
...
 
             # Find new valid attestations matching post-state justification
-            new_attestations: list[SignedVote] = []
-            for validator_id, checkpoint in self.latest_known_votes.items():
+            new_attestations: list[ValidatorAttestation] = []
+            for signed in self.latest_known_votes.values():
                 # Skip if target block is unknown in our store
-                if checkpoint.root not in self.blocks:
+                data = signed.message.data
+                if data.target.root not in self.blocks:
```
g11tech:
```suggestion
                if data.head.root not in self.blocks:
```

we need to check for `head` which is the target  of the attestation

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -420,25 +485,27 @@ def produce_block(self, slot: Slot, validator_index: ValidatorIndex) -> Block:
...
                     continue
 
                 # Create attestation with post-state's latest justified as source
-                vote = Vote(
-                    validator_id=validator_id,
-                    slot=checkpoint.slot,
-                    head=checkpoint,
-                    target=checkpoint,
+                attestation_data = AttestationData(
```
g11tech, line 505: we need to add a check here if data.source is same as post_state.latest_justified else we should skip adding this attestation data  to the block

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -365,9 +431,12 @@ def get_vote_target(self) -> Checkpoint:
         target_block = self.blocks[target_block_root]
         return Checkpoint(root=hash_tree_root(target_block), slot=target_block.slot)
 
-    def produce_block(self, slot: Slot, validator_index: ValidatorIndex) -> Block:
+    def produce_block(self, slot: Slot, validator_index: ValidatorIndex) -> BlockAndSignatures:
```
g11tech: @GrapeBaBa I added a new type to hold the signatures we need to accumulate for attestations, pls make sure everything is fine syntax/build wise

src/lean_spec/subspecs/containers/block/block.py
```diff
@@ -66,6 +66,18 @@ class BlockAndVote(Container):
     proposer_attestation: ValidatorAttestation
     """The proposer's vote corresponding to this block."""
 
+class BlockAndSignatures(Container):
+    """Bundle containing a block and its attestation signatures list."""
+
+    block: Block
+    """The proposed block message."""
+
+    signatures: BlockSignatures
```
unnawut: Maybe I got confused but does this mean we have signatures in `BlockAndSignatures.signatures` and also in `Block.body.attestations` which contains a list of `SignedVote`?

g11tech: no block.body.attestations is unsigned, signature has been stripped out in signatures

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -468,9 +543,16 @@ def produce_block(self, slot: Slot, validator_index: ValidatorIndex) -> Block:
         self.blocks[block_hash] = finalized_block
         self.states[block_hash] = final_post_state
 
-        return finalized_block
+        return BlockAndSignatures(
+            block=finalized_block,
```
g11tech:
```suggestion
            block=finalized_block,
            vote=proposer_vote,
```

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -468,9 +543,16 @@ def produce_block(self, slot: Slot, validator_index: ValidatorIndex) -> Block:
         self.blocks[block_hash] = finalized_block
         self.states[block_hash] = final_post_state
 
-        return finalized_block
+        return BlockAndSignatures(
```
g11tech:
```suggestion
        return BlockAndVoteAndSignatures(
```

src/lean_spec/subspecs/forkchoice/store.py
g11tech: @GrapeBaBa pls make this just as a return type in store.py and not a container

src/lean_spec/subspecs/containers/block/block.py
g11tech: ok lets do this:

ValidatorAttestation -> Attestation
SignedValidationAttestation -> SignedAttestation

Attestation -> AggreagtedAttestation
SignedAttestation -> SignedAggreagtedAttestation

cc @GrapeBaBa

src/lean_spec/subspecs/containers/block/block.py
g11tech: Also yes this container will be removed as it will only be used as a type in the forkchoice store for the purpose of representing intermediary step in block production

cc @GrapeBaBa

src/lean_spec/subspecs/containers/block/block.py
jihoonsong: To avoid any misunderstanding, this is `AggreagtedAttestations` and `SignedAggreagtedAttestations`, right?

src/lean_spec/subspecs/forkchoice/store.py
GrapeBaBa: I assume this should not change @g11tech

src/lean_spec/subspecs/forkchoice/store.py
GrapeBaBa: I assume this should not change @g11tech

tests/lean_spec/subspecs/forkchoice/test_validator.py
```diff
@@ -113,8 +142,8 @@ def test_produce_block_basic(self, sample_store: Store) -> None:
         slot = Slot(1)
         validator_idx = ValidatorIndex(1)  # Proposer for slot 1
 
-        block = sample_store.produce_block(slot, validator_idx)
-
+        block_and_signatures = sample_store.produce_block(slot, validator_idx)
```
jihoonsong:
```suggestion
        block, signatures = sample_store.produce_block(slot, validator_idx)
```
You can remove `BlockAndSignatures` as well.

src/lean_spec/subspecs/containers/__init__.py
```diff
@@ -1,19 +1,37 @@
...
     "Checkpoint",
     "Config",
-    "SignedBlock",
-    "SignedVote",
+    "SignedBlockWithAttestation",
+    "Validator",
+    "AttestationData",
+    "Attestation",
+    "SignedAttestation",
+    "SignedAggreagtedAttestations",
```
unnawut:
```suggestion
    "SignedAggregatedAttestations",
```

also applies to other places

src/lean_spec/subspecs/containers/attestation/attestation.py
```diff
@@ -0,0 +1,89 @@
...
+class AttestationData(Container):
+    """Attestation content describing the validator's observed chain view."""
+
+    slot: Slot
+    head: Checkpoint
+    target: Checkpoint
+    source: Checkpoint
+
+
+class Attestation(Container):
```
unnawut, line 26: Can we rename the folder and file name to attestation as well?

src/lean_spec/subspecs/containers/vote/vote.py
```diff
@@ -0,0 +1,89 @@
...
+
+    @property
+    def target(self) -> Checkpoint:
+        """Return the attested target checkpoint."""
+        return self.data.target
+
+    @property
+    def source(self) -> Checkpoint:
+        """Return the attested source checkpoint."""
+        return self.data.source
```
unnawut: Not sure if we need these accessor functions in the specs, seems more like implementation's decision to make than the specs

src/lean_spec/subspecs/containers/vote/vote.py
```diff
@@ -0,0 +1,89 @@
...
+
+    Unlike BLS, signatures over ValidatorAttestation messages that share the
+    same underlying AttestationData can be aggregated efficiently by the lean
+    signature VM.
+    """
+
+    @property
+    def data(self) -> Attestation:
+        """Expose the message for backwards compatibility with SignedVote."""
+        return self.message
```
unnawut: We don't need backward compatibility right now so this can be removed

src/lean_spec/subspecs/containers/vote/vote.py
```diff
@@ -0,0 +1,89 @@
...
+class AggreagtedAttestations(Container):
+    """Aggregated attestation consisting of participation bits and message."""
+
+    aggregation_bits: AggregationBits
+    data: AttestationData
+    """Combined vote data similar to the beacon chain format.
+
+    Multiple validator votes are aggregated here without the complexity of
+    committee assignments. This structure is defined for future use and is not
+    currently exercised by devnets.
```
unnawut:
```suggestion
    committee assignments.
```

I think it's needed in devnet1 where `aggregation_bits` will used to map `AggregatedSignatures` to their validator id.

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -106,91 +123,116 @@ class method acts as a factory for creating a new Store instance.
...
-        assert target_block.slot == vote.target.slot, "Target checkpoint slot mismatch"
+        assert source_block.slot == data.source.slot, "Source checkpoint slot mismatch"
+        assert target_block.slot == data.target.slot, "Target checkpoint slot mismatch"
 
         # Validate attestation is not too far in the future
-        current_slot = Slot(self.time // INTERVALS_PER_SLOT)
-        assert vote.slot <= Slot(current_slot + Slot(1)), "Attestation too far in future"
-
-    def process_attestation(self, signed_vote: "SignedVote", is_from_block: bool = False) -> None:
+        current_slot = Slot(self.time // SECONDS_PER_INTERVAL)
```
unnawut: `self.time // SECONDS_PER_INTERVAL` doesn't calculate to current slot because interval duration is different from slot time (and also results to `1` right now).

But I agree `self.time // INTERVALS_PER_SLOT` was correct only by coincidence that it's also `4`.

Did you mean `self.time // SLOT_DURATION_MS // 1000`?

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -106,91 +123,116 @@ class method acts as a factory for creating a new Store instance.
...
+            if latest_new is not None and latest_new.message.data.slot <= attestation_slot:
                 del self.latest_new_votes[validator_id]
-
         else:
             # Network gossip attestation processing
 
             # Ensure forkchoice is current before processing gossip
-            time_slots = self.time // INTERVALS_PER_SLOT
-            assert vote.slot <= time_slots, "Attestation from future slot"
+            time_slots = self.time // SECONDS_PER_INTERVAL
```
unnawut: Similar to previous comment:

```suggestion
            time_slots = self.time // SLOT_DURATION_MS // 1000
```

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -106,91 +123,116 @@ class method acts as a factory for creating a new Store instance.
...
+    def _validate_block_signatures(
+        self,
+        block: Block,
+        signatures: BlockSignatures,
+    ) -> bool:
+        """Temporary stub for aggregated signature validation."""
+        # TODO: Integrate actual aggregated signature verification.
+        return all(self._is_valid_signature(signature) for signature in signatures)
+
+    def process_block(self, signed_block_vote: SignedBlockWithAttestation) -> None:
```
unnawut: Since the type name changed

```suggestion
    def process_block(self, signed_block_with_attestation: SignedBlockWithAttestation) -> None:
```

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -421,32 +494,41 @@ def produce_block(self, slot: Slot, validator_index: ValidatorIndex) -> Block:
             post_state = advanced_state.process_block(candidate_block)
 
             # Find new valid attestations matching post-state justification
-            new_attestations: list[SignedVote] = []
-            for validator_id, checkpoint in self.latest_known_votes.items():
+            new_attestations: list[Attestation] = []
+            new_signatures: list[Bytes4000] = []
+            for signed in self.latest_known_votes.values():
```
```suggestion
            for signed_attestation in self.latest_known_votes.values():
```

naming ambiguity

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -421,32 +494,41 @@ def produce_block(self, slot: Slot, validator_index: ValidatorIndex) -> Block:
...
+                    validator_id=signed.message.validator_id,
+                    data=attestation_data,
+                )
 
-                # Include if not already in attestation set
-                if signed_vote not in attestations:
-                    new_attestations.append(signed_vote)
+                if candidate_attestation not in attestations:
+                    new_attestations.append(candidate_attestation)
+                    new_signatures.append(signed.signature)
```
unnawut: I don't think we can rebuild the attestation freely from `store.latest_known_votes`. This basically invalidates the attestion's signature.

But looking back at the [markdown specs](https://github.com/leanEthereum/leanSpec/blob/4b750f2748a3718fe3e1e9cdb3c65e3a7ddabff5/docs/client/validator.md#construction--broadcast) and [3SF-mini.py](https://github.com/ethereum/research/blob/master/3sf-mini/p2p.py#L104-L134) we weren't doing this. So maybe this was a hiccup from the markdown to python specs migration.

We can leave this in another PR though since this was before this PR.

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -455,7 +537,7 @@ def produce_block(self, slot: Slot, validator_index: ValidatorIndex) -> Block:
             proposer_index=validator_index,
             parent_root=head_root,
             state_root=Bytes32.zero(),  # Will be updated with computed hash
-            body=BlockBody(attestations=Attestations(data=attestations)),
+            body=BlockBody(attestations=Attestations(data=list(attestations))),  # need copy
```
unnawut: Not sure what the comment means?

src/lean_spec/subspecs/forkchoice/store.py
unnawut: Tracking separately as https://github.com/leanEthereum/leanSpec/issues/82

src/lean_spec/subspecs/containers/block/block.py
```diff
@@ -56,14 +57,29 @@ class Block(Container):
...
+    signature: BlockSignatures
+    """Aggregated signature payload for the block.
+
+    Signatures remain in attestation order followed by the proposer signature
+    over entire message. For devnet 1, however the proposer signature is just
+    over message.proposer_attestation since lean VM is not yet there to establish
+    validity/mergability of the signature for its packaging into the future blocks.
+
+    Eventually this field will be replaced by a single zk-aggregated signature.
     """
```
tcoratger, line 86:
```suggestion
    """Aggregated signature payload for the block.

    Signatures remain in attestation order followed by the proposer signature
    over entire message. For devnet 1, however the proposer signature is just
    over message.proposer_attestation since leanVM is not yet performant enough
    to aggregate signatures with sufficient throughput.

    Eventually this field will be replaced by a SNARK (which represents the aggregation of all signatures).
    """
```

src/lean_spec/types/byte_arrays.py
```diff
@@ -217,12 +217,24 @@ class Bytes48(BaseBytes):
...
 class Bytes96(BaseBytes):
     """Fixed-size byte array of exactly 96 bytes."""
 
     LENGTH = 96
 
 
+class Bytes4000(BaseBytes):
+    """Fixed-size byte array of exactly 4000 bytes."""
+
+    LENGTH = 4000
```
tcoratger, line 236: This type is very unusual and will probably be removed in the future.

As this file is about types that should be a bit more stable than the rest of the code, can you put a TODO or something to indicate that this will be removed in the future?

src/lean_spec/subspecs/containers/attestation/attestation.py
tcoratger: Agreed

src/lean_spec/subspecs/containers/vote/vote.py
tcoratger: Agreed, this makes the spec heavier and we can always come and get them with `self.data.source` for example

src/lean_spec/subspecs/containers/attestation/attestation.py
```diff
@@ -0,0 +1,89 @@
...
+    """Validator attestation bundled with its signature."""
+
+    message: Attestation
+    signature: Bytes4000
+    """Signature produced by the lean signature VM.
+
+    Unlike BLS, signatures over ValidatorAttestation messages that share the
+    same underlying AttestationData can be aggregated efficiently by the lean
+    signature VM.
+    """
```
tcoratger: We don't need to be more specific, otherwise this is confusing.

```suggestion
    """Signature aggregation produced by the leanVM (SNARKs in the future)."""
```

src/lean_spec/subspecs/containers/vote/vote.py
tcoratger: Agreed

src/lean_spec/subspecs/containers/attestation/attestation.py
```diff
@@ -0,0 +1,89 @@
...
+    @property
+    def source(self) -> Checkpoint:
+        """Return the attested source checkpoint."""
+        return self.data.source
+
+
+class SignedAttestation(Container):
+    """Validator attestation bundled with its signature."""
+
+    message: Attestation
```
tcoratger, line 39: If we have a doc string for signature then we need a doc string for the message as well I guess

src/lean_spec/subspecs/containers/attestation/attestation.py
```diff
@@ -0,0 +1,89 @@
...
+    @property
+    def data(self) -> Attestation:
+        """Expose the message for backwards compatibility with SignedVote."""
+        return self.message
+
+
+class AggreagtedAttestations(Container):
+    """Aggregated attestation consisting of participation bits and message."""
+
+    aggregation_bits: AggregationBits
```
tcoratger, line 49: We probably need doc string for this as well (just 1 line).

src/lean_spec/subspecs/containers/vote/vote.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+
+
+class SignedAggreagtedAttestations(Container):
+    """Aggregated attestation bundled with aggregated signatures."""
+
+    message: AggreagtedAttestations
```
tcoratger: We should probably have a doc string here as well?

src/lean_spec/subspecs/containers/attestation/attestation.py
```diff
@@ -0,0 +1,89 @@
...
+
+    message: AggreagtedAttestations
+    signature: AggregatedSignatures
+    """Aggregated vote plus its combined signature.
+
+    Stores a naive list of validator signatures that mirrors the attestation
+    order; this will be replaced by a single zk-verified signature in later
+    devnets. Lean signatures permit recursive aggregation even when individual
+    ValidatorAttestation and Attestation messages differ, so long as their
+    embedded AttestationData matches.
```
tcoratger:
```suggestion
    Stores a naive list of validator signatures that mirrors the attestation
    order.
    
    TODO: this will be replaced by a SNARK in future devnets.
```

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -13,23 +14,39 @@
...
 
 from .helpers import get_fork_choice_head, get_latest_justified
 
 
+@dataclass(slots=True)
+class BlockAndSignatures:
+    """Internal bundle pairing a block with its attestation signatures."""
+
+    block: Block
+    signatures: list[Bytes4000]
```
tcoratger: Why not `BlockAndAttestations` here to be consistent with the rest?

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -106,91 +123,116 @@ class method acts as a factory for creating a new Store instance.
...
-    def process_block(self, block: Block) -> None:
+            if latest_new is None or latest_new.message.data.slot < attestation_slot:
+                self.latest_new_votes[validator_id] = signed_attestation
+
+    @staticmethod
+    def _is_valid_signature(signature: Bytes4000) -> bool:
+        """Return True when the placeholder signature is the zero value."""
+        # TODO: Replace placeholder check once aggregated signatures are
+        # wired in as part of the multi-proof integration work.
+        return signature == Bytes4000.zero()
```
tcoratger, line 203: Here the method doesn't use self, meaning this is probably the wrong place for it. Shouldn't we have a specific type signature defined as a class and have this method implemented inside that class instead of here?

src/lean_spec/subspecs/forkchoice/store.py
tcoratger: Agreed, we should modify the doc string as well if we change this

tests/lean_spec/subspecs/forkchoice/test_validator.py
```diff
@@ -137,17 +166,29 @@ def test_produce_block_unauthorized_proposer(self, sample_store: Store) -> None:
...
+            head=Checkpoint(root=sample_store.head, slot=head_block.slot),
+            source=sample_store.latest_justified,
+            target=sample_store.get_vote_target(),
+        )
 
         slot = Slot(2)
         validator_idx = ValidatorIndex(2)  # Proposer for slot 2
 
-        block = sample_store.produce_block(slot, validator_idx)
+        block_and_signatures = sample_store.produce_block(slot, validator_idx)
```
tcoratger:
```suggestion
        block_and_attestations = sample_store.produce_block(slot, validator_idx)
```

src/lean_spec/subspecs/forkchoice/store.py
GrapeBaBa: merge issue, it should be `INTERVALS_PER_SLOT`

tests/lean_spec/subspecs/forkchoice/test_validator.py
GrapeBaBa: @jihoonsong suggest removing this type and just return multi values, is that preference

jihoonsong: Yeah the type is just for the return value and is not necessary. Clients can choose to use such intermediary types, if they prefer.

src/lean_spec/subspecs/forkchoice/store.py
GrapeBaBa: It should be replaced by the actual signature verification by following PR, already have a TODO: comment, maybe leave it to that PR

tcoratger (approved):

tests/lean_spec/subspecs/containers/test_state.py
```diff
@@ -469,6 +503,9 @@ def test_generate_genesis(sample_config: Config) -> None:
     assert state.slot == Slot(0)
     # Body root must commit to an empty body at genesis.
     expected_body = BlockBody(attestations=Attestations(data=[]))
+
+    print(f"body_root hex: {state.latest_block_header.body_root.hex()}")
+
```
jihoonsong:
```suggestion
```

jihoonsong (approved):
This PR is exactly 1 month old. This can go forever. Let's merge this and iterate fast.

PR #68, opened 2025-09-27
fix: :bug: votes bitlist should equal validator numbers not validator limit

GrapeBaBa:
## 🗒️ Description

This pull request fix a bug that votes bitlist size should equal validator numbers not validator limit

## ✅ Checklist

    ```console
    uvx --with=tox-uv tox
    ```


unnawut:
For later reference, we decided to go with num_validators because we don't expect a varying number of validators soon. This should be revisited one we have varying validators.

PR #76, opened 2025-10-21
feat(tests): Initial implementation for generating client test vectors

fselmo:
## 🗒️ Description

I have ported over some ideas from `execution-spec-tests` for vector filling. It may not be perfectly cleaned up yet but I wanted to put this out so we can start discussing the format for the vectors from implementers' sides and think about how you want to consume these.

If you're not familiar with `execution-spec-tests`, I will try to summarize some of the high-level ideas behind the approach here. Coincidentally, `execution-spec-tests` is in the middle of merging into `execution-specs` as a monorepo this week. I've been waiting a bit to see the final state of the monorepo organization so we can mimic this here as I think it's a good idea to try to keep things as familiar as possible for developers.

## Workspace with Test Package

The approach I took here is the same as we did for execution repos. I have set up `leanSpec` as a `uv` _workspace_ with a `packages/` directory where the `tests` framework lives. Inside this `packages/`, for now, only exists this testing framework and this is considered a separate package altogether, though they share the same workspace. In here you'll find it has its own setup with `pyproject.toml`, etc, though the main linting, spellcheck, and all these things are configured at the root level of the repository (leanSpec/pyproject.toml). All the `tox` commands have also been updated to run these checks across the entirety of the workspace, which makes things quite maintainable.

## Filling Test Vectors

I defined a basic set of fixtures for now. I think it would be quite nice to find the fewest entry points that we can test the consensus layer from, as the public API entry points, and write all tests from these as test fixtures / formats. What I mean is, we currently have `blockchain_test`, `state_test`, `benchmark_test` in execution specs. If we could somehow get some small subset of test fixtures for consensus, this could be ideal from an integration standpoint. If this is not possible, however, we can create as many fixture types as we need to but this is a chance to redesign the approach we are taking so we should think on this. If a minimal subset can be used, an added benefit is that we can have a more "integration" type of test where we're not only checking logic in one place but from the entry point and across all relevant logic.

The test fixtures I defined for now with some influence from `consensus-specs` vector designs:

- `state_transition_test`: Test via the STF as the entry point
- `fork_choice_test`: Tests that use fork choice as the test logic

We have a new command:

```
uv run fill --clean --fork=devnet
```

- `fill`: This is the command to fill test vectors from pytest tests that are written inside of `tests/spec_tests`
- `--clean`: This flag forces you to specify that you want to write over the current `fixtures/` so that you don't accidentally remove already-filled tests you want to keep
- `--fork`: Specifies the relevant fork to fill for, right now only `devnet` makes sense
- `--output`: Can be specified to fill fixtures to an output directory that isn't `fixtures/`, so you can compare two different sets if you want or fill them somewhere else entirely.

Please use this as a discussion as this is not by any means attempting to finalize the design for the test vectors.

---
Note: This commit (6c892e408fbf562ce94937549180c813384fd547) makes some assumptions that change behavior in the specs but I leaned into some changes that allowed the fork choice to move the chain so I could properly test some things as the `fork_choice_test` was not working without these.

## ✅ Checklist

    ```console
    uvx --with=tox-uv tox
    ```

### DIFF
--- a/packages/testing/src/consensus_testing/genesis.py
+++ b/packages/testing/src/consensus_testing/genesis.py
@@ -22,6 +22,7 @@ def generate_pre_state(**kwargs: Any) -> State:
     genesis_time = kwargs.get("genesis_time", Uint64(0))
 
     # If validators not provided, create a default set of 4 validators with dummy pubkeys
+    # TODO: Set an appropriate default here for test fixtures
     if "validators" not in kwargs:
         validators = Validators(data=[Validator(pubkey=Bytes52.zero()) for _ in range(4)])
     else:
--- a/packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
+++ b/packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
@@ -129,20 +129,12 @@ class ForkChoiceTest(BaseConsensusFixture):
                     store.advance_time(Uint64(step.time), has_proposal=False)
 
                 elif isinstance(step, BlockStep):
-                    # Build SignedBlockWithAttestation from BlockSpec if needed
-                    if isinstance(step.block, BlockSpec):
-                        signed_block = self._build_block_from_spec(step.block, store)
-                        # Store the actual Block object for serialization
-                        block = signed_block.message.block
-                        step.block = block
-                    else:
-                        # Already a Block object - this shouldn't happen in normal usage
-                        # but we handle it for completeness
-                        block = step.block
-                        # Would need to build SignedBlockWithAttestation wrapper
-                        raise NotImplementedError(
-                            "BlockStep with pre-built Block not yet supported"
-                        )
+                    # Build SignedBlockWithAttestation from BlockSpec
+                    signed_block = self._build_block_from_spec(step.block, store)
+
+                    # Store the filled Block for serialization
+                    block = signed_block.message.block
+                    step._filled_block = block
 
                     # Automatically advance time to block's slot before processing
                     # Compute the time corresponding to the block's slot
--- a/packages/testing/src/consensus_testing/test_types/step_types.py
+++ b/packages/testing/src/consensus_testing/test_types/step_types.py
@@ -1,8 +1,8 @@
 """Step types for fork choice tests."""
 
-from typing import Annotated, Literal, Union
+from typing import Annotated, Any, Literal, Union
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, ConfigDict, Field, PrivateAttr, field_serializer
 
 from lean_spec.subspecs.containers import SignedAttestation
 from lean_spec.subspecs.containers.block.block import Block
@@ -59,17 +59,55 @@ class BlockStep(BaseForkChoiceStep):
     Output: Block object built and processed through the spec.
     """
 
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
     step_type: Literal["block"] = "block"
     """Discriminator field for serialization."""
 
-    block: BlockSpec | Block
+    block: BlockSpec
     """
-    Block specification or built block.
+    Block specification for this step.
 
-    Input: Tests provide BlockSpec with optional field overrides.
-    Output: Framework builds Block and replaces during make_fixture().
-    Serialization: Only Block objects are serialized.
+    Tests provide a BlockSpec with required slot and optional field overrides.
+    The framework fills a complete Block during make_fixture() and stores it
+    in the private _filled_block attribute for serialization.
     """
+
+    # TODO: We should figure out a configuration to raise if a private attr is
+    #  attempted to be set during model initialization.
+    _filled_block: Block | None = PrivateAttr(default=None)
+    """The filled Block, processed through the spec."""
+
+    @field_serializer("block", when_used="json")
+    def serialize_block(self, value: BlockSpec) -> dict[str, Any]:
+        """
+        Serialize the filled Block instead of the BlockSpec.
+
+        This ensures the fixture output contains the complete Block that was
+        filled from the spec, not the input BlockSpec.
+
+        Parameters:
+        ----------
+        value : BlockSpec
+            The BlockSpec field value (ignored, we use _filled_block instead).
+
+        Returns:
+        -------
+        dict[str, Any]
+            The serialized Block.
+
+        Raises:
+        ------
+        ValueError
+            If _filled_block is None (make_fixture not called yet).
+        """
+        del value
+        if self._filled_block is None:
+            raise ValueError(
+                "Block not filled yet - make_fixture() must be called before serialization. "
+                "This BlockStep should only be serialized after the fixture has been processed."
+            )
+        return self._filled_block.model_dump(mode="json")
 
 
 class AttestationStep(BaseForkChoiceStep):
--- a/src/lean_spec/subspecs/forkchoice/store.py
+++ b/src/lean_spec/subspecs/forkchoice/store.py
@@ -227,7 +227,6 @@ class Store(Container):
         Process new block and update forkchoice state.
 
         Adds block to store, processes included attestations, and updates head.
-        This implements the `on_block` function from the fork choice spec.
 
         Args:
             signed_block_with_attestation: Block to process.
--- a/packages/testing/src/consensus_testing/test_fixtures/state_transition.py
+++ b/packages/testing/src/consensus_testing/test_fixtures/state_transition.py
@@ -1,8 +1,8 @@
 """State transition test fixture format."""
 
-from typing import ClassVar, List
+from typing import Any, ClassVar, List
 
-from pydantic import field_serializer
+from pydantic import ConfigDict, PrivateAttr, field_serializer
 
 from lean_spec.subspecs.containers.block.block import Block, BlockBody
 from lean_spec.subspecs.containers.block.types import Attestations
@@ -41,16 +41,28 @@ class StateTransitionTest(BaseConsensusFixture):
         "epochs, and finality"
     )
 
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
     pre: State
     """The initial consensus state before processing."""
 
-    blocks: List[BlockSpec | Block]
+    blocks: List[BlockSpec]
     """
-    The blocks that were processed through the spec.
+    Block specifications to process through the spec.
 
-    Input: Tests provide List[BlockSpec] with optional field overrides.
-    Output: Framework builds Block objects and replaces this list during make_fixture().
-    Serialization: Only Block objects are serialized (validated in make_fixture).
+    Tests provide a list of BlockSpec objects with required slots and optional
+    field overrides. The framework fills complete Block objects during
+    make_fixture() and stores them in the private _filled_blocks attribute.
+    """
+
+    # TODO: We should figure out a configuration to raise if a private attr is
+    #  attempted to be set during model initialization.
+    _filled_blocks: List[Block] = PrivateAttr(default_factory=list)
+    """
+    The filled Blocks, processed through the specs.
+
+    This is a private attribute not part of the model schema. Tests cannot set this.
+    The framework populates it during make_fixture().
     """
 
     post: StateExpectation | None = None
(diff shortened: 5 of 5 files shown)


packages/tests/src/lean_spec_tests/test_fixtures/fork_choice.py
```diff
@@ -0,0 +1,163 @@
...
+            state=self.anchor_state,
+            anchor_block=self.anchor_block,
+        )
+
+        # Process each step
+        for i, step in enumerate(self.steps):
+            try:
+                if isinstance(step, TickStep):
+                    # Advance time
+                    from lean_spec.types import Uint64
```
tcoratger: Can't we have the imports in the header of the file? I think we have the same thing at multiple places. This doesn't change anything but probably removes some lines of code :)

packages/tests/src/lean_spec_tests/test_fixtures/fork_choice.py
```diff
@@ -0,0 +1,163 @@
...
+
+                    store.advance_time(Uint64(step.time), has_proposal=False)
+
+                elif isinstance(step, BlockStep):
+                    # Automatically advance time to block's slot before processing
+                    # (Like consensus-specs' tick_and_add_block helper)
```
tcoratger: Maybe we don't need to reference the consensus spec no? Not sure

tests/consensus/devnet/fc_examples/test_head_selection.py
```diff
@@ -0,0 +1,85 @@
+"""Fork choice head selection tests for the devnet fork."""
```
tcoratger, line 1: I love this way of doing tests here, this is much simpler than before and we will be able in the future to drastically simplify the way tests are handled in the repo by just having this way of doing everywhere.

tcoratger (commented):
Maybe we should put in the `CLAUDE.md` the description of the PR you made or something similar so that our Claude agents know how to deal with the current way of doing tests?

unnawut:
@fselmo We just merged another big PR hence the merge conflict. Could you help rebase?

fselmo:
> @fselmo We just merged another big PR hence the merge conflict. Could you help rebase?

Done. Cleaned up the implementation a bit too.

unnawut:
I'm getting this error when running uv sync:

```
➜  leanSpec git:(feat/test-vectors) uv sync --all-extras
error: Workspace member `.../leanSpec/packages/tests` is missing a `pyproject.toml` (matches: `packages/*`)
```

tests/consensus/devnet/stf_examples/test_blocks.py
```diff
@@ -0,0 +1,71 @@
...
+        pre=genesis,
+        blocks=[
+            BlockSpec(slot=Slot(1)),
+            BlockSpec(slot=Slot(2)),
+            BlockSpec(slot=Slot(3)),
+        ],
+        post=StateExpectation(
+            slot=Slot(3),
+            validator_count=4,
+        ),
```
unnawut, line 55: Probably syntax error but is it possible to assert a nested element i.e. `state.config.num_validators == 4`?

```suggestion
        post=StateExpectation(
            slot=Slot(3),
            config=Config(num_validators=4),
        ),
```

fselmo: > Probably syntax error but is it possible to assert a nested element i.e. `state.config.num_validators == 4`?

I think #84 changes things, no? Depends which gets merged first.

unnawut: Right we can deal with this separately outside this PR

fselmo:
Rebased again. I also added a CI run that fills all tests with python 3.14. Right now it's just the example tests I added to illustrate the basic flow but as we tweak this and add actual tests, this will likely be a good sanity check that nothing is breaking.

unnawut (approved):
🚀🚀🚀

CONTRIBUTING.md
```diff
@@ -3,10 +3,10 @@
 ## Quick Start
 
 1. Fork and clone the repository
-2. Install dependencies: `uv sync`
+2. Install dependencies: `uv sync --all-packages`
 3. Make your changes
 4. Run checks: `uvx tox -e all-checks`
-5. Run tests: `uv run pytest`
+5. Run tests: `uvx pytest`
```
jihoonsong, line 9: README.md and CLAUDE.md use `uv run pytest ` for running tests and not sure why but running `uvx pytest` gives me
```
tests/conftest.py:3: in <module>
    from hypothesis import settings
E   ModuleNotFoundError: No module named 'hypothesis'
```

packages/testing/src/consensus_testing/genesis.py
```diff
@@ -0,0 +1,30 @@
...
+            - validators: Validators list (defaults to 4 validators with dummy pubkeys).
+
+    Returns:
+        State: A properly initialized consensus state.
+    """
+    genesis_time = kwargs.get("genesis_time", Uint64(0))
+
+    # If validators not provided, create a default set of 4 validators with dummy pubkeys
+    if "validators" not in kwargs:
+        validators = Validators(data=[Validator(pubkey=Bytes52.zero()) for _ in range(4)])
```
jihoonsong, line 27: Do these validators better to have different pubkeys?

packages/testing/src/consensus_testing/test_types/step_types.py
```diff
@@ -0,0 +1,97 @@
...
+    This updates the store's block tree and may trigger head updates.
+
+    Input: BlockSpec (can be partial or fully specified).
+    Output: Block object built and processed through the spec.
+    """
+
+    step_type: Literal["block"] = "block"
+    """Discriminator field for serialization."""
+
+    block: BlockSpec | Block
```
jihoonsong: AFAIK, using the same argument for both input and output is deemed as a bad practice as it's confusing. Is it going to hurt to separate block spec and block explicitly in the context of this framework?

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -216,6 +227,7 @@ def process_block(self, signed_block_with_attestation: SignedBlockWithAttestatio
         Process new block and update forkchoice state.
 
         Adds block to store, processes included attestations, and updates head.
+        This implements the `on_block` function from the fork choice spec.
```
jihoonsong: This comment is in the fork choice spec of Lean. Is the fork choice spec here refer to the Beacon chain spec..? Maybe we don't need this explanation, which might be false.

unnawut: Good catch. Agree this sentence is not needed.

fselmo: Thanks for catching. This is when Claude takes my directions literally and then spells them back out for me 😆. I borrowed ideas from `consensus-specs` fork choice test as a way to start with something familiar and see if we can improve upon it. Will remove 👍🏼

packages/testing/src/consensus_testing/test_types/step_types.py
fselmo: Yes this is definitely lazy on my part and should be improved. The goal is to take `BlockSpec` which is not a frozen model and does not need to be fully specified if we have some defaults we can fill in, then fill defaults if any, pass it through the spec, and serialize the end result - which at the moment is `Block`. I will clean this up 👍🏼 ... I think we can use a `field_serializer` here and use the filled block instead.

packages/testing/src/consensus_testing/genesis.py
fselmo: > Do these validators better to have different pubkeys?

Yes this is probably a good idea and this will likely change anyhow right? Also, what number of validators would be appropriate to set as a default here you think?

CONTRIBUTING.md
fselmo: Thanks this was a recent change in #85 and looks like it needs a little more polish. I'll address this 👍🏼

fselmo: ah, I missed the nuance that you weren't trying `uvx tox -e pytest` (which will run the tox env where we define the appropriate deps it needs). `uvx pytest` will only run `pytest` because `uvx` creates a fresh new environment every time. You'd need to specify to this fresh virtual env everything you need to run the command. Since our tests depend on `hypothesis`, `pytest-cov`, and `lean-spec`, you would need to run something like:

```
uvx --with=hypothesis --with=pytest-cov --with=. pytest  
```

Where `--with=.` points to the editable version of `lean-spec` (root dir). This does work... but this is also why we have `tox`. You can run `uvx tox -e pytest` which defines all the dependencies that the pytest tox env needs, uses the uv lock runner (so will read the versions we have defined in our `uv.lock`), and runs the commands we've specified.

This is why the docs say to use `uv run pytest` after syncing... because now you have synced all the deps you need and if you want to test something specific you can use `uv run pytest -k ... -m ...` or any pytest arg to test locally since you are now also synced with the `uv.lock` versions.

Does that make sense? 🤔

fselmo: You can also pass positional args to the `tox -e pytest` command using `--` to separate `tox` and `pytest` args:

```bash
uvx tox -e pytest -- -m "not slow"
```

But I would suggest `uv run pytest` after syncing the project because this is cleaner whereas `tox -e pytest` is mostly meant to run _all_ the tests in an efficient way so that we can use this in our CI as well. You can pass pos args to it but it's a bit of an unnecessary layer.

tcoratger (approved):

packages/testing/src/consensus_testing/genesis.py
fselmo: Added a TODO here... we can find an appropriate default configuration and update it here in a follow-up.

packages/testing/src/consensus_testing/test_types/step_types.py
fselmo: This should be cleaner now. We store the filled block and serialize the `block` fields as the filled block with the `field_serializer` approach I mentioned. This is much cleaner and we can refactor more if we'd like in the future but now we have a separation of concerns and proper serialization instructions. Thanks!

CONTRIBUTING.md
jihoonsong: Thanks! But what I was pointing out was CONTRIBUTING.md says to use `uvx pytest`, while other docs use `uv run pytest`.

packages/testing/src/consensus_testing/genesis.py
jihoonsong: There is no required minimum number of validators unlike beacon chain, right? I agree that we can leave it as a TODO as you said.

CONTRIBUTING.md
fselmo: Wow sorry that slipped right by me. Thought it was a local command issue you had. Will fix, thanks for flagging 🙂

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -0,0 +1,264 @@
...
+            If any step fails unexpectedly or checks don't match Store state.
+        """
+        # Ensure anchor_state and anchor_block are set
+        assert self.anchor_state is not None, "anchor_state must be set before make_fixture"
+        assert self.anchor_block is not None, "anchor_block must be set before make_fixture"
+
+        # Initialize Store from anchor
+        store = Store.get_forkchoice_store(
+            state=self.anchor_state,
+            anchor_block=self.anchor_block,
```
g11tech, line 121: actually we don't need block, because the state here one to one corresponds with the block unlike beacon where state could have been rolled to the checkpoint because of missing slots towards the end

tests/consensus/devnet/fc_examples/test_head_selection.py
g11tech: yes looks pretty cool

g11tech (commented):
post merge review, looks amazing :heart:

CONTRIBUTING.md
fselmo: Fixed in #96 !

jihoonsong: Amazing, thanks!

PR #80, opened 2025-10-23
feat: add python specs for Gossip Topics

varun-doshi:
## 🗒️ Description

There were no python specs for Gossip Topics. The only information about this is in [networking.md](https://github.com/leanEthereum/leanSpec/blob/a66265cb80acb3278ce513831d459c1ab12772cb/docs/client/networking.md?plain=1#L65C2-L73C20).
Since we are moving away from markdown specs and focusing on the Python version, it will be helpful to have some code specs for this.
- Adds a new file for GossipTopics and the respective allowed payload types.
- The reason this uses `Vote` as the topic although I previously opened a [PR](https://github.com/leanEthereum/leanSpec/pull/78) to change the topic to `Attestation` is because the respective updated payload is still in a working [PR](https://github.com/leanEthereum/leanSpec/pull/67). We can update this once that PR is merged in.

    ```console
    uvx --with=tox-uv tox
    ```

### DIFF
--- a/src/lean_spec/subspecs/networking/topics.py
+++ b/src/lean_spec/subspecs/networking/topics.py
@@ -1,19 +1,45 @@
 """Gossipsub topics"""
 
 from enum import Enum
-
-from pydantic import Field
-from typing_extensions import Annotated
+from typing import Any, Type
 
 from lean_spec.subspecs.containers.block.block import SignedBlock
 from lean_spec.subspecs.containers.vote import SignedVote
 
-BlockPayload = Annotated[SignedBlock, Field(description="Payload for block topic.")]
-VotePayload = Annotated[SignedVote, Field(description="Payload for vote topic.")]
 
+class GossipTopic(Enum):
+    """
+    Enumerates gossip topics, bundling a topic's name with its payload type.
 
-class GossipTopic(str, Enum):
-    """Enum representing allowed gossip topics and their payload types."""
+    Attributes:
+        value (str): The network name of the topic (e.g., "block").
+        payload_type (Type): The class representing the data structure
+            of the topic's message (e.g., `SignedBlock`).
+    """
 
-    BLOCK = BlockPayload
-    VOTE = VotePayload
+    def __init__(self, value: str, payload_type: Type[Any]):
+        """
+        Initializes the GossipTopic.
+
+        Args:
+            value: The topic in string.
+            payload_type: The associated gossip.
+        """
+        self._value_ = value
+        self.payload_type = payload_type
+
+    BLOCK = ("block", SignedBlock)
+    """
+    Topic for gossiping new blocks.
+
+    - `value`: "block"
+    - `payload_type`: `SignedBlock`
+    """
+
+    VOTE = ("vote", SignedVote)
+    """
+    Topic for gossiping new votes (attestations).
+
+    - `value`: "vote"
+    - `payload_type`: `SignedVote`
+    """


src/lean_spec/subspecs/networking/topics.py
```diff
@@ -0,0 +1,24 @@
...
+
+BlockPayload = Annotated[SignedBlock, Field(description="Payload for block topic.")]
+VotePayload = Annotated[SignedVote, Field(description="Payload for vote topic.")]
+
+
+class GossipTopic(str, Enum):
+    """Enum representing allowed gossip topics and their payload types."""
+
+    BLOCK = BlockPayload
+    VOTE = VotePayload
```
unnawut: BLOCK and VOTE seems undefined. Translating from https://github.com/leanEthereum/leanSpec/blob/4b750f2748a3718fe3e1e9cdb3c65e3a7ddabff5/docs/client/networking.md?plain=1#L125-L128 these should be the strings `"block"` and `"vote"`

src/lean_spec/subspecs/networking/topics.py
```diff
@@ -0,0 +1,24 @@
+"""
+Gossipsub topics
+
+- Allowed GossipTopic.
+- Allowed Payload annotated with the payload type.
```
unnawut:
```suggestion
```

I find these doc comments a bit cryptic. Can just remove them.

tests/lean_spec/subspecs/forkchoice/test_attestation_processing.py
```diff
@@ -215,6 +215,52 @@ def test_validate_attestation_too_far_future(self, sample_store: Store) -> None:
         with pytest.raises(AssertionError, match="Attestation too far in future"):
             sample_store.validate_attestation(signed_vote)
 
+    def test_validate_attestation_unknown_head_rejected(self, sample_store: Store) -> None:
```
unnawut, line 218: Changing the forkchoice test should be out of scope of this PR

varun-doshi: Yeah looks like I went one commit back by mistake...it should be solved now
No test chaneges only addition of new type

src/lean_spec/subspecs/networking/topics.py
tcoratger: I agree, we can simplify by removing them and just add appropriate precise documentation inside the file, function by function or object by object when needed.

src/lean_spec/subspecs/networking/topics.py
varun-doshi: If we set this to string, then we loose the mapping to the `Payload` types.
There are a few different nuances to this:
- `BlockPayload = Annotated[SignedBlock, Field(description="Payload for block topic.")]` the description specifies the exact topic string ("block")
- We can change the key of the `GossipTopic` to be the exact string so it will be block and vote (not BLOCK and VOTE)
- or we can set the value to the string something like BLOCK = "block". This would mean we need a function to get the `payload_type`

tcoratger: can't we have something like this

```python
class GossipTopic(Enum):
    """
    Enumerates gossip topics, bundling a topic's name with its payload type.

    Attributes:
        value (str): The network name of the topic (e.g., "block").
        payload_type (Type): The class representing the data structure of the topic's message (e.g., `SignedBlock`).
    """

    def __init__(self, value: str, payload_type: Type[Any]):
        self._value_ = value
        self.payload_type = payload_type

    BLOCK = ("block", SignedBlock)
    """
    Topic for gossiping new blocks.

    - `value`: "block"
    - `payload_type`: `SignedBlock`
    """

    VOTE = ("vote", SignedVote)
    """
    Topic for gossiping new votes (attestations).

    - `value`: "vote"
    - `payload_type`: `SignedVote`
    """
```

unnawut: @tcoratger’s suggestion lgtm and also in line with reqresp domain!

tox.ini
```diff
@@ -6,7 +6,7 @@ env_list =
 skip_missing_interpreters = true
 
 [testenv]
-basepython = python3
+basepython = python3.14.0
```
tcoratger: @fselmo do we want this?

fselmo: I don't think we want to force this. @varun-doshi any reason you added this in particular?

I have some updates to `tox` / `uv.lock` I want to get in with a separate PR. I will open one today.

varun-doshi: Ahh no this got committed by mistake. I was testing some changes in the `tox` file and got pushed accidentally. 
Removing this now

varun-doshi: Done!

fselmo: np, sounds good :)

tcoratger (approved):
LGTM

PR #92, opened 2025-10-28
feat: add Signature container

varun-doshi:
## 🗒️ Description

Adds a new container for `Signature`.
This should help abstract away `Bytes3100`
Moved `_is_valid_signature` from `state` to `signature`

## ✅ Checklist

    ```console
    uvx tox -e all-checks,pytest
    ```


src/lean_spec/subspecs/containers/signature.py
```diff
@@ -0,0 +1,62 @@
...
+
+    ELEMENT_TYPE = Uint8
+    LENGTH = 32
+
+
+class Parameter(SSZVector):
+    """A 32-byte parameter."""
+
+    ELEMENT_TYPE = Uint8
+    LENGTH = 32
```
unnawut: Is this Parameter used anywhere?

src/lean_spec/subspecs/containers/signature.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class Randomness(SSZVector):
+    """A 32-byte randomness seed."""
+
+    ELEMENT_TYPE = Uint8
+    LENGTH = 32
```
unnawut: This one would be

```suggestion
    LENGTH = XmssConfig.RAND_LEN_FE * BYTES_PER_FE
```

ref: https://github.com/leanEthereum/leanSpec/blob/main/src/lean_spec/subspecs/xmss/constants.py#L72

src/lean_spec/subspecs/containers/signature.py
```diff
@@ -0,0 +1,62 @@
...
+from lean_spec.types.uint import Uint8
+
+MAX_TREE_DEPTH = 32
+
+
+class HashDigest(SSZVector):
+    """A 32-byte hash digest."""
+
+    ELEMENT_TYPE = Uint8
+    LENGTH = 32
```
unnawut: Would be nice if this refers directly to the xmss specs so they're always in sync. Something like:

```suggestion
    LENGTH = XmssConfig.HASH_LEN_FE * BYTES_PER_FE
```

ref: https://github.com/leanEthereum/leanSpec/blob/main/src/lean_spec/subspecs/xmss/constants.py#L75

src/lean_spec/subspecs/containers/signature.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class HashDigestList(SSZList):
+    """A list of hash digests."""
+
+    ELEMENT_TYPE = HashDigest
+    LIMIT = MAX_TREE_DEPTH
```
unnawut: Assuming `HashDigestList` is for `signature.hashes` and not `signature.path`, this should be:

```suggestion
    LIMIT = XmssConfig.DIMENSION
```

src/lean_spec/subspecs/containers/signature.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class HashTreeOpening(Container):
+    """A Merkle authentication path."""
+
+    siblings: HashDigestList
```
unnawut: We cannot reuse `HashDigestList` here if it's already used for `signature.hashes` because they are of different sizes.

So we probably need something like `SiblingHashes` that's a list of `HashDigest` of size `XmssConfig.LOG_LIFETIME`

tests/lean_spec/subspecs/containers/test_state.py
```diff
@@ -182,6 +184,15 @@ def sample_block_header() -> BlockHeader:
...
 
+def sample_signature() -> Signature:
+    """Return a zero-value placeholder signature."""
+    return Signature(
+        path=HashTreeOpening(siblings=HashDigestList(data=[])),
+        rho=Randomness(data=[0] * Randomness.LENGTH),
+        hashes=HashDigestList(data=[]),
+    )
+
+
```
unnawut: I think we should avoid duplicating this function in each file. Can you look into creating a factory like this?

https://github.com/leanEthereum/leanSpec/blob/main/tests/lean_spec/subspecs/forkchoice/conftest.py#L85-L88

unnawut:
@tcoratger Would love your 2nd opinion on my comments above especially because it'll be referring xmss subspecs

src/lean_spec/subspecs/containers/signature.py
unnawut: @varun-doshi Sorry I know we chatted about these fixed values earlier, but this idea of referring to xmss subspecs instead came after I read the PR again

varun-doshi: Yup got it...these changes makes sense. Will update

src/lean_spec/subspecs/containers/signature.py
tcoratger: Agreed if not used we should delete

src/lean_spec/subspecs/containers/signature.py
```diff
@@ -0,0 +1,62 @@
...
+
+class Signature(Container):
+    """A signature produced by the `sign` function."""
+
+    path: HashTreeOpening
+    rho: Randomness
+    hashes: HashDigestList
+
+    @staticmethod
+    def _is_valid_signature(signature: Signature) -> bool:
```
tcoratger: Since this is already inside the `Signature` class, you can just call it : `_is_valid`

tests/lean_spec/subspecs/containers/test_state.py
tcoratger: Yep agreed

g11tech (changes requested):
to be honest,  why do we want to expose how signature looks like to the spec?

I am open to have this but want to know the benefits, con is obviously more tech debt seepage and breakdown of the abstraction layer

src/lean_spec/types/byte_arrays.py
```diff
@@ -231,11 +231,11 @@ class Bytes96(BaseBytes):
     LENGTH = 96
 
 
-class Bytes4000(BaseBytes):
-    """Fixed-size byte array of exactly 4000 bytes."""
+class Bytes3100(BaseBytes):
+    """Fixed-size byte array of exactly 3100 bytes."""
 
     # TODO: this will be removed when real signature type is implemented
```
jihoonsong: I think we don't need this TODO comment. Byte arrays here themselves are purpose-agnostic.

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -206,21 +206,14 @@ def process_attestation(
...
-    @staticmethod
-    def _is_valid_signature(signature: Bytes4000) -> bool:
-        """Return True when the placeholder signature is the zero value."""
-        # TODO: Replace placeholder check once aggregated signatures are
-        # wired in as part of the multi-proof integration work.
-        return signature == Bytes4000.zero()
-
     def _validate_block_signatures(
         self,
         block: Block,
```
jihoonsong, line 211: Why do we pass `block` here?

src/lean_spec/subspecs/containers/signature.py
```diff
@@ -0,0 +1,21 @@
...
+from __future__ import annotations
+
+from lean_spec.types import Bytes3100
+
+
+class Signature(Bytes3100):
+    """Represents aggregated signature produced by the leanVM (SNARKs in the future)."""
+
+    @staticmethod
+    def _is_valid_signature(signature: Bytes3100) -> bool:
```
jihoonsong:
```suggestion
    def _is_valid_signature(signature: Signature) -> bool:
```
I think we want the parameter type be Signature?

tests/lean_spec/subspecs/forkchoice/conftest.py
```diff
@@ -23,7 +24,7 @@
     JustifiedSlots,
 )
 from lean_spec.subspecs.ssz.hash import hash_tree_root
-from lean_spec.types import Bytes32, Bytes4000, Uint64, ValidatorIndex
+from lean_spec.types import Bytes32, Bytes3100, Uint64, ValidatorIndex
```
jihoonsong:
```suggestion
from lean_spec.types import Bytes32, Uint64, ValidatorIndex
```

tests/lean_spec/subspecs/containers/test_state.py
```diff
@@ -27,7 +28,7 @@
     Validators,
 )
 from lean_spec.subspecs.ssz import hash_tree_root
-from lean_spec.types import Boolean, Bytes32, Bytes52, Bytes4000, Uint64, ValidatorIndex
+from lean_spec.types import Boolean, Bytes32, Bytes52, Bytes3100, Uint64, ValidatorIndex
```
jihoonsong:
```suggestion
from lean_spec.types import Boolean, Bytes32, Bytes52, Uint64, ValidatorIndex
```
We are abstracting Bytes3100 away with Signature.

tests/lean_spec/subspecs/forkchoice/test_validator.py
```diff
@@ -25,7 +26,7 @@
 )
 from lean_spec.subspecs.forkchoice import Store
 from lean_spec.subspecs.ssz.hash import hash_tree_root
-from lean_spec.types import Bytes32, Bytes52, Bytes4000, Uint64, ValidatorIndex
+from lean_spec.types import Bytes32, Bytes52, Bytes3100, Uint64, ValidatorIndex
```
jihoonsong:
```suggestion
from lean_spec.types import Bytes32, Bytes52, Uint64, ValidatorIndex
```

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -17,7 +17,7 @@
 from lean_spec.subspecs.containers.state.state import State
 from lean_spec.subspecs.forkchoice import Store
 from lean_spec.subspecs.ssz import hash_tree_root
-from lean_spec.types import Bytes32, Bytes4000, Uint64, ValidatorIndex
+from lean_spec.types import Bytes32, Bytes3100, Uint64, ValidatorIndex
```
jihoonsong:
```suggestion
from lean_spec.types import Bytes32, Uint64, ValidatorIndex
```

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -254,7 +254,7 @@ def _build_block_from_spec(self, spec: BlockSpec, store: Store) -> SignedBlockWi
 
         # Create signed structure with placeholder signatures
         # One signature for proposer attestation + one for the block
-        signature_list = [Bytes4000.zero(), Bytes4000.zero()]
+        signature_list = [Bytes3100.zero(), Bytes3100.zero()]
```
jihoonsong:
```suggestion
        signature_list = [Signature.zero(), Signature.zero()]
```
This will require importing Signature from subspecs.containers.

tests/lean_spec/subspecs/forkchoice/test_attestation_processing.py
```diff
@@ -9,13 +9,14 @@
...
     Config,
+    Signature,
     SignedAttestation,
 )
 from lean_spec.subspecs.containers.block import Attestations
 from lean_spec.subspecs.containers.slot import Slot
 from lean_spec.subspecs.forkchoice import Store
 from lean_spec.subspecs.ssz.hash import hash_tree_root
-from lean_spec.types import Bytes32, Bytes4000, Uint64, ValidatorIndex
+from lean_spec.types import Bytes32, Bytes3100, Uint64, ValidatorIndex
```
jihoonsong:
```suggestion
from lean_spec.types import Bytes32, Uint64, ValidatorIndex
```

src/lean_spec/subspecs/forkchoice/store.py
unnawut: I suggest we review this as part of the upcoming signature integration PR(s). I'm expecting there'll be changes here from those PRs

varun-doshi: Agreed

tests/lean_spec/subspecs/forkchoice/test_attestation_processing.py
unnawut, line 1: Not sure why this file pops up in this PR

src/lean_spec/subspecs/containers/signature.py
```diff
@@ -0,0 +1,21 @@
...
+from __future__ import annotations
+
+from lean_spec.types import Bytes3100
+
+
+class Signature(Bytes3100):
+    """Represents aggregated signature produced by the leanVM (SNARKs in the future)."""
+
+    @staticmethod
+    def _is_valid(signature: Signature) -> bool:
```
unnawut: I don't think we need the underscore prefix

```suggestion
    def is_valid(signature: Signature) -> bool:
```

src/lean_spec/subspecs/containers/signature.py
```diff
@@ -0,0 +1,21 @@
...
+        """Return True when the placeholder signature is the zero value."""
+        # TODO: Replace placeholder check once aggregated signatures are
+        # wired in as part of the multi-proof integration work.
+        return signature == Signature.zero()
+
+    @classmethod
+    def zero(cls) -> Signature:
+        """Return the zero (placeholder) signature."""
+        return cls(Bytes3100.zero())
```
unnawut, line 21: This is only used in tests so better turn it into a test fixture.

varun-doshi: I was thinking this is probably going to be modified in the upcoming PRs and so its okay to let it be?

tests/lean_spec/subspecs/forkchoice/test_attestation_processing.py
varun-doshi: Because i removed an unused import from that file

varun-doshi: actually, you're right. Not sure why

src/lean_spec/subspecs/containers/signature.py
unnawut: ok sure since it was carried over from the previous PR anyway. Let's remove the unrelated test files and get this merged

tests/lean_spec/subspecs/forkchoice/test_attestation_processing.py
jihoonsong: This is coming from #98. I strongly oppose to using abbreviation especially in the higher level such as directory/file/class names.

jihoonsong (approved):
The PR looks good. I do want to resolve the `fc` stuff but that's not part of this PR.

tests/lean_spec/subspecs/forkchoice/test_attestation_processing.py
unnawut: btw. this file was removed since #100 so probably a rebase issue. Agree we should use full names for files and folders though and let's do that as separate PR

fselmo:
> This is coming from https://github.com/leanEthereum/leanSpec/pull/98. I strongly oppose to using abbreviation especially in the higher level such as directory/file/class names.

Agreed. I added this only to the sample test cases. For clarity, we don't need to separate tests as `forkchoice/` and `state_transition/`. They can exist in the same test files if it makes sense to. If you'd rather categorize tests under different scopes and have forkchoice _and_ state transition tests for that scope, you definitely can.

The way we do this in execution specs is usually `fork/eip1234_description_of_eip/test_*.py`. Hopefully that makes sense. Usually grouping by forks at higher level, then scope by EIPs, then have each test file further break down the scope of the tests within them... can even have further breakdown by directories if needed but usually not at this point. But then you can define any fixture / test type (`forkchoice_test` or `state_transition_test`) within those files.

I would say choose a structure that works best for you but by no means do we have to use the fixture type as the higher level scope here and in some cases this probably wouldn't make the most sense.

PR #94, opened 2025-10-29
fix: make `Store` mutable

syjn99:
## 🗒️ Description

Currently, `Store` inherits SSZ `Container` class which enforce the fork choice store immutable. This led some anti patterns like `object.__setattr__` to modify its field, and I think `Store` should be modifiable to track the necessary values. Also `Store` isn't a target for SSZ encoding/decoding.

This PR nukes immutability of `Store` by giving `BaseModel` from Pydantic as inheritance.

## ✅ Checklist

    ```console
    uvx tox -e all-checks,pytest
    ```

### DIFF
--- a/src/lean_spec/subspecs/forkchoice/store.py
+++ b/src/lean_spec/subspecs/forkchoice/store.py
@@ -14,6 +14,8 @@ __all__ = [
 import copy
 from typing import Dict
 
+from pydantic import BaseModel
+
 from lean_spec.subspecs.chain.config import (
     INTERVALS_PER_SLOT,
     SECONDS_PER_INTERVAL,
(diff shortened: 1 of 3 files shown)


tcoratger:
> Thanks for the explanation. Just to make sure I understood you correctly, we would want to return a new object whenever we want to modify it, right?

Yes, whenever possible, it's better to be friendly to formal verification. Then, of course, if we have one big, specific case where this isn't practical at all, then we should discuss and perhaps consider having something mutable, but whenever possible, yes, it's better to return a new object than to mutate an existing one.

For example, here's a good example of where I tried to implement this strategy:

https://github.com/leanEthereum/leanSpec/blob/e7976fa94649f2641fe46070fe940a65942451ec/src/lean_spec/subspecs/containers/state/state.py#L250

jihoonsong:
Sounds great, that approach looks aligned well with the direction. Having PRs that adopt that approach, or explore a better one if any, would be ahead of us.

syjn99:
Hey all, I think I can switch the goal of this PR to return a new copied object as @tcoratger provided. I'm gonna change the status of this one as draft at this moment. Thanks!

PR #97, opened 2025-10-29
fix(tooling): uvenv lock runner and deps

fselmo:
## 🗒️ Description

This builds on the changes from #85

- I split the dependency groups into appropriate categories, mostly just for organizing instead of having comments - `dev` still includes everything. What I hadn't noticed is the `uv-venv-lock-runner` for `tox` was not being defined in the right place `[testenv]` (not `[tox]`). This PR re-organizes everything and gets rid of the need to use `--all-packages` since `dev` dependency group is always synced by default and we now include the package as a dev dependency. This has the benefit of only needing `uv sync` (without `--all-packages`) for syncing the full dev environment which is quite nice imo.

- I re-organized the filler `pytest.ini` under the `packages/testing` framework as this makes more sense there.

- This updates the `tox` docs command to be: `docs-build` and `docs-serve` - to be more clear. `uvx tox` runs `all-checks`, `pytest`, and `docs-build` so is now the recommended check for PRs in the PR template as it's quick and covers almost everything.

- I removed what seem to be unused markdown deps from #58, @syjn99 if you can sanity check me that we indeed don't use these.

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -133,11 +133,6 @@ docs = [
     "mkdocs>=1.6.1,<2",
     "mkdocs-material>=9.5.45,<10",
     "mkdocstrings[python]>=0.27.0,<1",
-    "pyspelling>=2.8.2,<3",
-    "mdformat-gfm-alerts==2.0.0",
-    "mdformat-gfm==0.4.1",
-    "mdformat-ruff==0.1.3",
-    "mdformat-toc==0.3.0",
     "mdformat==0.7.22",
 ]
 dev = [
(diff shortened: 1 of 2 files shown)


tcoratger (approved):
Nice that is cleaner for the user, thanks!

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -9,7 +9,7 @@
 
 - [ ] Ran `tox` checks to avoid unnecessary CI fails:
     ```console
-    uvx tox -e all-checks,pytest
+    uvx tox
```
unnawut, line 12: Hey @tcoratger this command ends with serving the docs server, can you help exclude mkdocs serve somehow?

PR #98, opened 2025-10-29
tests: add forkchoice attestation processing test vectors

tcoratger:
## 🗒️ Description

This pull request adds basic test vectors for attestation processing via a new `AttestationCheck` model. We can, of course, elaborate further on what we test within it, but I think we have the basics here, and we can iterate over them.

This already provides a useful test base for the next steps, which client teams can consume. These tests should correspond to the old tests in the previous file of the same name (`test_attestation_processing.py`), which was located in the tests/lean_spec folder. I'll remove this old file in a follow-up pull request when this one is merged in order to process step by step.

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/packages/testing/src/consensus_testing/test_types/store_checks.py
+++ b/packages/testing/src/consensus_testing/test_types/store_checks.py
@@ -33,12 +33,11 @@ class AttestationCheck(CamelModel):
     target_slot: Slot | None = None
     """Expected target checkpoint slot."""
 
-    location: Literal["new", "known"] | None = None
+    location: Literal["new", "known"]
     """
     Expected attestation location:
-        - "new" for `latest_new_attestations`,
-        - "known" for `latest_known_attestations`,
-        - `None` to check both.
+        - "new" for `latest_new_attestations`
+        - "known" for `latest_known_attestations`
     """
 
     def validate_attestation(
@@ -220,7 +219,7 @@ class StoreChecks(CamelModel):
                 for check in expected_value:
                     validator_idx = check.validator
 
-                    # Determine where to look for the attestation
+                    # Check attestation location
                     if check.location == "new":
                         if validator_idx not in store.latest_new_attestations:
                             raise AssertionError(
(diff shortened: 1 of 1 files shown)


packages/testing/src/consensus_testing/test_types/store_checks.py
```diff
@@ -1,14 +1,88 @@
...
+
+    target_slot: Slot | None = None
+    """Expected target checkpoint slot."""
+
+    location: Literal["new", "known"] | None = None
+    """
+    Expected attestation location:
+        - "new" for `latest_new_attestations`,
+        - "known" for `latest_known_attestations`,
+        - `None` to check both.
```
unnawut: Is there a scenario where we need None? I can only think of cases where we want to check either or but not both

tests/consensus/devnet/fc/test_attestation_processing.py
```diff
@@ -0,0 +1,694 @@
...
+    """
+    Attestations move from latest_new to latest_known between blocks.
+
+    Scenario
+    --------
+    Process blocks at slots 1 and 2 (different proposers: validators 1 and 2).
+
+    Expected:
+        - After slot 1: new attestations = 1, known attestations = 0
+        - After slot 2: new attestations = 1, known attestations = 1
```
unnawut: I think these 2 lines are a bit misleading because new/known attestations should take account of non-proposer attestations too and the test is not asserting the count

tests/consensus/devnet/fc/test_attestation_processing.py
```diff
@@ -0,0 +1,694 @@
...
+                            location="new",
+                        ),
+                    ],
+                ),
+            ),
+        ],
+    )
+
+
+def test_slot_gaps_with_attestation_superseding(
```
unnawut, line 355: Nice scenario!

unnawut (commented):
Amazing stuff!

Just one non-trivial suggestion. I see we're only testing proposer's attestations right now. I think we should also test non-proposer attestations since they have different logic in `store.process_block()`.

They could either be added to the existing test cases or separate ones. But I think separate + in another PR is cleaner?

jihoonsong (approved):

packages/testing/src/consensus_testing/test_types/store_checks.py
tcoratger: Yes you are right, just fixed this.

jihoonsong:
I overlooked some details. Can we avoid abbreviation? `test_attestation_processing.py` is put under `consensus/devnet/fc` but can we name it `consensus/devnet/forkchoice`? We use `attestation` instead of `att`, `aggregation` instead of `agg` , `management` instead of `mngt`, and so on. @tcoratger

PR #101, opened 2025-10-30
tests: add forkchoice `test_attestation_target_selection`

tcoratger:
## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/tests/consensus/devnet/fc/test_attestation_target_selection.py
+++ b/tests/consensus/devnet/fc/test_attestation_target_selection.py
@@ -303,37 +303,54 @@ def test_attestation_target_justifiable_constraint(
     fork_choice_test: ForkChoiceTestFiller,
 ) -> None:
     """
-    Attestation target advances gradually while maintaining justifiability.
+    Attestation target advances while respecting justifiability rules.
 
     Scenario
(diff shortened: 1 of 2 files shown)


tests/consensus/devnet/fc/test_attestation_target_selection.py
```diff
@@ -0,0 +1,360 @@
...
+                checks=StoreChecks(
+                    head_slot=Slot(8),
+                    attestation_target_slot=Slot(5),  # Continues advancing
+                ),
+            ),
+        ],
+    )
+
+
+def test_attestation_target_justifiable_constraint(
```
unnawut, line 302: Is this basically this rule? https://github.com/leanEthereum/leanSpec/blob/main/src/lean_spec/subspecs/containers/slot.py#L39-L48

If so can we make direct reference to it? Otherwise the docstring and test impl kind of implies arbitrary slow down.

tcoratger: You are right, just added more doc to explain a bit more.

unnawut (approved):
Slick! ♥️

PR #103, opened 2025-10-30
tests: add `test_fork_choice_head` test vectors

tcoratger:
## 🗒️ Description

I guess that we can produce better scenarios and have deeper checks but I've setup here some basics to play with forks using a friendly format hopefully. Then we can iterate from there.

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/packages/testing/src/consensus_testing/test_types/store_checks.py
+++ b/packages/testing/src/consensus_testing/test_types/store_checks.py
@@ -7,6 +7,7 @@ from lean_spec.types import Bytes32, CamelModel, Uint64, ValidatorIndex
 
 if TYPE_CHECKING:
     from lean_spec.subspecs.containers import SignedAttestation
+    from lean_spec.subspecs.containers.block.block import Block
     from lean_spec.subspecs.forkchoice.store import Store
 
 
@@ -109,6 +110,18 @@ class StoreChecks(CamelModel):
     head_root: Bytes32 | None = None
     """Expected head block root."""
 
+    head_root_label: str | None = None
+    """
+    Expected head block root by label reference.
+
+    Alternative to head_root that uses the block label system.
+    The framework will resolve this label to the actual block root
+    and validate the head matches.
+
+    Example:
+        StoreChecks(head_root_label="fork_a")  # Validates head is fork_a block
+    """
+
     latest_justified_slot: Slot | None = None
     """Expected latest justified checkpoint slot."""
 
@@ -136,7 +149,9 @@ class StoreChecks(CamelModel):
     attestation_checks: list[AttestationCheck] | None = None
     """Optional list of attestation content checks for specific validators."""
 
-    def validate_against_store(self, store: "Store", step_index: int) -> None:
+    def validate_against_store(
+        self, store: "Store", step_index: int, block_registry: dict[str, "Block"] | None = None
+    ) -> None:
         """
         Validate these checks against actual Store state.
 
@@ -149,6 +164,8 @@ class StoreChecks(CamelModel):
             The fork choice store to validate against.
         step_index : int
             Index of the step being validated (for error messages).
+        block_registry : dict[str, Block] | None
+            Optional registry of labeled blocks for resolving head_root_label.
 
         Raises:
         ------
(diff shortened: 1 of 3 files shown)


packages/testing/src/consensus_testing/test_types/block_spec.py
```diff
@@ -52,3 +52,31 @@ class BlockSpec(CamelModel):
...
+    Enables explicit fork creation:
+    ```python
+    BlockSpec(slot=Slot(1), label="common")
+    BlockSpec(slot=Slot(2), parent_label="common", label="fork_a")
+    BlockSpec(slot=Slot(2), parent_label="common", label="fork_b")
+    ```
+
+    If None, parent is determined by the current canonical head.
+    If specified, parent_root is computed from the labeled block.
+    """
```
unnawut, line 82: Very handy!

tests/consensus/devnet/fc/test_fork_choice_head.py
```diff
@@ -0,0 +1,392 @@
...
+    ForkChoiceTestFiller,
+    StoreChecks,
+)
+
+from lean_spec.subspecs.containers.slot import Slot
+
+pytestmark = pytest.mark.valid_until("Devnet")
+
+
+def test_head_advances_through_deep_chain(
```
unnawut: Nit just for clarity because we're not asserting the root:

```suggestion
def test_head_slot_advances_through_deep_chain(
```

tests/consensus/devnet/fc/test_fork_choice_head.py
```diff
@@ -0,0 +1,392 @@
...
+                checks=StoreChecks(
+                    head_slot=Slot(i),
+                ),
+            )
+            for i in range(1, 21)
+        ],
+    )
+
+
+def test_head_with_gaps_in_slots(
```
unnawut, line 108:
```suggestion
def test_head_slot_with_gaps_in_slots(
```

tests/consensus/devnet/fc/test_fork_choice_head.py
```diff
@@ -0,0 +1,392 @@
...
+            ),
+            BlockStep(
+                block=BlockSpec(slot=Slot(9)),  # Skip slot 8
+                checks=StoreChecks(head_slot=Slot(9)),
+            ),
+        ],
+    )
+
+
+def test_head_with_large_gaps(
```
unnawut, line 163:
```suggestion
def test_head_slot_with_large_gaps(
```

tests/consensus/devnet/fc/test_fork_choice_head.py
```diff
@@ -0,0 +1,392 @@
...
+            ),
+            # Competing fork B at slot 2
+            BlockStep(
+                block=BlockSpec(
+                    slot=Slot(2),
+                    parent_label="common",  # Same parent
+                    label="fork_b",
+                ),
+                # Head determined by tie-breaker (lexicographic root order)
+                checks=StoreChecks(head_slot=Slot(2)),
```
unnawut: I think we should also check correct head_root here?

Otherwise we're only validating that slot progresses correctly but this statement is not validated: "Head determined by tie-breaker (lexicographic root order)"

tests/consensus/devnet/fc/test_fork_choice_head.py
```diff
@@ -0,0 +1,392 @@
...
+                ),
+                checks=StoreChecks(head_slot=Slot(2)),  # Still fork A
+            ),
+            # Extend fork B - gives it more weight
+            BlockStep(
+                block=BlockSpec(
+                    slot=Slot(3),
+                    parent_label="fork_b",  # Build on fork B
+                ),
+                checks=StoreChecks(head_slot=Slot(3)),  # Head switches to fork B!
```
unnawut: Similar to earlier comment. Not sure whether checking slot number is sufficient that the chain is correct. Client impl could be building wrong chain up to the correct slot. Wdyt?

tests/consensus/devnet/fc/test_fork_choice_head.py
tcoratger: Finally, I found it more appropriate to check the root as well, so I left the old name.

tests/consensus/devnet/fc/test_fork_choice_head.py
tcoratger: Yes, I thought we could save that for later, but you're right, while we're at it, let's do that too. Just fixed.

tests/consensus/devnet/fc/test_fork_choice_head.py
tcoratger: Fixed

packages/testing/src/consensus_testing/test_types/store_checks.py
```diff
@@ -183,6 +200,35 @@ def validate_against_store(self, store: "Store", step_index: int) -> None:
...
+                    raise ValueError(
+                        f"Step {step_index}: head_root_label='{expected_value}' specified "
+                        f"but block_registry not provided to validate_against_store()"
+                    )
+
+                if expected_value not in block_registry:
+                    available = list(block_registry.keys())
+                    raise ValueError(
+                        f"Step {step_index}: head_root_label='{expected_value}' not found "
+                        f"in block registry. Available labels: {available}"
```
unnawut, line 215: Amazing error hinting!

PR #107, opened 2025-11-01
Signature integration

anshalshukla:
## Description
This PR adds broadly adds 3 main things.
- Serialization, deserialization function in xmss Signature & PubKey
- Signature verification logic
- Adds XmssKeyManager for signing in tests

I'm not sure about a few things which can be updated upon discussion.
- The serialization, deserialization function assume signature length to be 3100 bytes and pads it in case of test config.
- Do we need signing in case of tests, or we can revert back to zero bytes for signature in tests. This is drastically increasing the testing time (~10-11 mins), I think it's not being validated in tests.

    ```console
    uvx tox
    ```

### DIFF
--- a/packages/testing/src/consensus_testing/keys.py
+++ b/packages/testing/src/consensus_testing/keys.py
@@ -2,7 +2,6 @@
 
 from lean_spec.subspecs.containers import (
     Attestation,
-    Block,
     Signature,
     Validator,
 )
@@ -70,23 +69,6 @@ class XmssKeyManager:
         _, sk = self.get_or_create_key(validator_id)
         message = bytes(hash_tree_root(attestation))
         epoch = int(attestation.data.slot)
-        xmss_sig = TEST_SIGNATURE_SCHEME.sign(sk, epoch, message)
-        return Signature.from_xmss(xmss_sig)
-
-    def sign_block(self, block: Block, proposer_index: ValidatorIndex) -> Signature:
-        """
-        Sign a block with the given proposer index.
-
-        Args:
-            block: The block to sign.
-            proposer_index: The validator index of the block proposer.
-
-        Returns:
-            A signature for the given block.
-        """
-        _, sk = self.get_or_create_key(proposer_index)
-        message = bytes(hash_tree_root(block))
-        epoch = int(block.slot)
         xmss_sig = TEST_SIGNATURE_SCHEME.sign(sk, epoch, message)
         return Signature.from_xmss(xmss_sig)
 
--- a/src/lean_spec/subspecs/xmss/constants.py
+++ b/src/lean_spec/subspecs/xmss/constants.py
@@ -139,37 +139,41 @@ TWEAK_PREFIX_MESSAGE: Final = Fp(value=0x02)
 PRF_KEY_LENGTH: int = 32
 """The length of the PRF secret key in bytes."""
 
-FE_SIZE_BYTES: int = 4  # Size of one field element in bytes
+# Public key and signature sizes for the production and test configurations
 
-PROD_HASH_LEN_FE = PROD_CONFIG.HASH_LEN_FE  # 8 field elements per hash digest
-PROD_PARAMETER_LEN = PROD_CONFIG.PARAMETER_LEN  # 5 field elements for parameter
-PROD_RAND_LEN_FE = PROD_CONFIG.RAND_LEN_FE  # 7 field elements for randomness
-PROD_LOG_LIFETIME = PROD_CONFIG.LOG_LIFETIME  # 32 layers in merkle tree
-PROD_DIMENSION = PROD_CONFIG.DIMENSION  # 64 hash chains
+FE_SIZE_BYTES: int = 4
+"""Size of a single field element in bytes."""
 
-TEST_LOG_LIFETIME = TEST_CONFIG.LOG_LIFETIME  # 8 layers in merkle tree
-TEST_DIMENSION = TEST_CONFIG.DIMENSION  # 16 hash chains
-TEST_HASH_LEN_FE = TEST_CONFIG.HASH_LEN_FE  # 8 field elements per hash digest
-TEST_PARAMETER_LEN = TEST_CONFIG.PARAMETER_LEN  # 5 field elements for parameter
-TEST_RAND_LEN_FE = TEST_CONFIG.RAND_LEN_FE  # 7 field elements for randomness
+# 52 bytes
+PROD_PUBKEY_SIZE_BYTES = (PROD_CONFIG.HASH_LEN_FE + PROD_CONFIG.PARAMETER_LEN) * FE_SIZE_BYTES
+"""Size in bytes of the production public key representation."""
 
-PROD_PUBKEY_SIZE_BYTES = (PROD_HASH_LEN_FE + PROD_PARAMETER_LEN) * FE_SIZE_BYTES  # 52 bytes
-TEST_PUBKEY_SIZE_BYTES = (TEST_HASH_LEN_FE + TEST_PARAMETER_LEN) * FE_SIZE_BYTES  # 52 bytes
+# 52 bytes
+TEST_PUBKEY_SIZE_BYTES = (TEST_CONFIG.HASH_LEN_FE + TEST_CONFIG.PARAMETER_LEN) * FE_SIZE_BYTES
+"""Size in bytes of the test public key representation."""
 
 PROD_SIGNATURE_SIZE_BYTES = (
-    PROD_LOG_LIFETIME * PROD_HASH_LEN_FE * FE_SIZE_BYTES  # path siblings
-    + PROD_RAND_LEN_FE * FE_SIZE_BYTES  # rho
-    + PROD_DIMENSION * PROD_HASH_LEN_FE * FE_SIZE_BYTES  # hashes
+    PROD_CONFIG.LOG_LIFETIME * PROD_CONFIG.HASH_LEN_FE * FE_SIZE_BYTES
+    + PROD_CONFIG.RAND_LEN_FE * FE_SIZE_BYTES
+    + PROD_CONFIG.DIMENSION * PROD_CONFIG.HASH_LEN_FE * FE_SIZE_BYTES
 )  # 3100 bytes
+"""Total size in bytes of a production signature, including all components."""
 
 TEST_SIGNATURE_SIZE_BYTES = (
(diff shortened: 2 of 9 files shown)


tests/lean_spec/__init__.py
```diff
@@ -0,0 +1 @@
+"""Test package for lean_spec subspec modules."""
```
tcoratger, line 1: I think we can remove this that is not linked to the PR in order to keep changes minimals

tests/lean_spec/subspecs/__init__.py
```diff
@@ -0,0 +1 @@
+"""Test package containing subspec test suites."""
```
tcoratger, line 1: I think we can remove this that is not linked to the PR in order to keep changes minimals

tests/lean_spec/subspecs/containers/__init__.py
```diff
@@ -0,0 +1 @@
+"""Container-related test package."""
```
tcoratger, line 1: I think we can remove this that is not linked to the PR in order to keep changes minimals

src/lean_spec/subspecs/containers/signature.py
```diff
@@ -2,20 +2,70 @@
...
+        """
+        Create a protocol Signature from an XMSS signature.
+
+        Args:
+            xmss_sig: The XMSS signature to convert.
+
+        Returns:
+            A protocol Signature containing the serialized XMSS signature.
+        """
+        return cls(xmss_sig.to_bytes())
```
tcoratger, line 30: That should not be needed, this is an helper that is nice to have for clients in Rust for example but we should avoid complexify things in the Python spec and just do a simple class initialization everywhere needed.

src/lean_spec/subspecs/containers/signature.py
```diff
@@ -2,20 +2,70 @@
...
+        """
+        Convert this protocol Signature to an XMSS signature.
+
+        Returns:
+            The XMSS signature.
+
+        Raises:
+            ValueError: If the signature cannot be deserialized.
+        """
+        return XmssSignature.from_bytes(bytes(self))
```
tcoratger, line 42: Here I'm pretty sure there is a better way of doing things for serialization/deserialization, you can check in other places of the code as Pydantic is used. Should be a good way of doing things here what do you think @fselmo ?

src/lean_spec/subspecs/containers/signature.py
```diff
@@ -2,20 +2,70 @@
...
+        Returns:
+            True if the signature is valid, False otherwise.
+        """
+        try:
+            xmss_sig = self.to_xmss()
+            # check padding
+            if bytes(self)[-SIGNATURE_PADDING_LENGTH:] == b"\x00" * SIGNATURE_PADDING_LENGTH:
+                signature_scheme = TEST_SIGNATURE_SCHEME
+            else:
+                signature_scheme = PROD_SIGNATURE_SCHEME
```
tcoratger, line 67: This way of choosing between the two is strange no? I think we should have something more pythonic, via a parameter or something but this seems pretty strange

src/lean_spec/subspecs/containers/validator.py
```diff
@@ -1,10 +1,42 @@
...
+        """
+        Get the XMSS public key from this validator.
+
+        Returns:
+            The XMSS public key.
+
+        Raises:
+            ValueError: If the public key cannot be deserialized.
+        """
+        return XmssPublicKey.from_bytes(bytes(self.pubkey))
```
tcoratger, line 42: Same not sure sure here we are doing the best we can, at the end what we are looking for here is serialization/deserialization. I'm pretty sure there is a better way, what do you think @fselmo? Using pydantic as well?

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -211,9 +211,69 @@ def _validate_block_signatures(
...
+        attestations = list(block.body.attestations)
+
+        # +1 is for the proposer signature
+        if len(signatures) != len(attestations) + 1:
+            return False
+
+        # Validate each signature
+        for index, attestation in enumerate(attestations):
+            signature = signatures[index]
+            validator_id = int(attestation.validator_id)
```
tcoratger, line 233:
```suggestion
            validator_id = attestation.validator_id.as_int()
```

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -211,9 +211,69 @@ def _validate_block_signatures(
...
+
+            validator = state.validators[validator_id]
+
+            try:
+                xmss_public_key = validator.get_xmss_pubkey()
+            except (ValueError, Exception):
+                return False
+
+            message = bytes(hash_tree_root(attestation))
+            epoch = int(attestation.data.slot)
```
tcoratger, line 248:
```suggestion
            epoch = attestation.data.slot.as_int()
```

src/lean_spec/subspecs/xmss/constants.py
```diff
@@ -138,3 +138,42 @@ def POS_OUTPUT_LEN_FE(self) -> int:  # noqa: N802
...
+"""Maximum public key size across all XMSS presets."""
+
+SIGNATURE_SIZE_BYTES = max(PROD_SIGNATURE_SIZE_BYTES, TEST_SIGNATURE_SIZE_BYTES)  # 3100 bytes
+"""Maximum signature size across all XMSS presets."""
+
+PUBKEY_PADDING_LENGTH = abs(PROD_PUBKEY_SIZE_BYTES - TEST_PUBKEY_SIZE_BYTES)  # 0 bytes
+"""Padding required to align public key sizes between presets."""
+
+SIGNATURE_PADDING_LENGTH = abs(PROD_SIGNATURE_SIZE_BYTES - TEST_SIGNATURE_SIZE_BYTES)  # 2304 bytes
+"""Padding required to align signature sizes between presets."""
```
tcoratger, line 179: I don't think what we're doing here is correct because we're mixing two completely unrelated configurations:

- The production configuration, which corresponds to what's actually intended for production use, but whose key generation takes a very long time and is therefore very difficult (impossible) to use for CI or testing.

- The test configuration, which is very small, random, and not trusted from a security standpoint. It's only used to run CI unit tests, nothing more.

So I think that if we need parameters like the size of public keys or something like that, we should consider the best way to do it in a clean and Pythonic way. Probably some methods as part of the implementation of `XmssConfig` seem reasonable to me. But the mixing we are doing here looks confusing and not secure at all.

src/lean_spec/subspecs/xmss/containers.py
```diff
@@ -118,6 +217,111 @@ class Signature(StrictBaseModel):
...
+        hashes: List[HashDigest] = []
+        for _ in range(dimension):
+            hash_digest: HashDigest = []
+            for _ in range(hash_len_fe):
+                value = int.from_bytes(data[offset : offset + FE_SIZE_BYTES], byteorder="little")
+                hash_digest.append(Fp(value=value))
+                offset += FE_SIZE_BYTES
+            hashes.append(hash_digest)
+
+        return cls(path=path, rho=rho, hashes=hashes)
```
tcoratger, line 323: Here I see two problems:

- For serialization and deserialization, I think we can find a better, more Pythonic framework, but if it's simpler, I'm totally open to keeping things as they are for this pull request.

- The biggest problem I see here is the mix of constants. It seems to me that we should have `XmssConfig` as part of the `Signature` class, or something similar, and use that within the function. Having the instantiation constants (with a mix of production and testing constants) inside the functions is really confusing to me.

tcoratger (commented):
I left some comments, can you solve conflicts? 

My recommendation for this to be merged quickly would be to split it into several PRs. 

For example a division which seems relevant to me to have for a smoother flux:
1. Serialization methods for the public key (+1/2 methods if needed inside the xmss config if this is needed for the public key serialization)
2. Same for the signature container.
3. A PR just to introduce the `XmssKeyManager`.
4. A PR to add the verify method to the signature container
5. ...

The goal would be to have super minimal PRs and fast merging process. I guess this is easier for you as well but as you prefer since this one is already done, we can start from there and handle the fixes progressively.

packages/testing/src/consensus_testing/keys.py
```diff
@@ -0,0 +1,89 @@
...
+        Args:
+            attestation: The attestation to sign.
+
+        Returns:
+            A signature for the given attestation.
+        """
+        validator_id = attestation.validator_id
+
+        _, sk = self.get_or_create_key(validator_id)
+        message = bytes(hash_tree_root(attestation))
```
g11tech, line 70: should we mix in domain? @tcoratger I don't see a reason why though, since these keys sign only once per slot

src/lean_spec/subspecs/containers/signature.py
g11tech: ideally one should instantiate a global signature scheme which could be done at the start of the process (of running tests or a client)

src/lean_spec/subspecs/containers/validator.py
g11tech: I agree we don't need these extra methods, one can just set Validator or get instantiate a XmssPublicKey directly using validator's property

although it would make sense to cache the loaded `XmssPublicKey` in some cache, but I think thats a client level optimization and spec doesn't need to worry about that

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -211,9 +211,69 @@ def _validate_block_signatures(
...
+
+            # Verify the signature
+            if not signature.verify(xmss_public_key, epoch, message):
+                return False
+
+        # Does not validate the proposer signature
+        # The proposer signature should be validated in the process_block method
```
g11tech, line 255: no this needs to be validated here, proposer signature will not be validated in process_block separately as all the signatures will be validated in one go

@anshalshukla

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -211,9 +211,69 @@ def _validate_block_signatures(
...
+
+        return True
+
+    def _validate_proposer_attestation(
```
g11tech, line 259: validate along with others in validate block signatures

src/lean_spec/subspecs/xmss/containers.py
g11tech: the better way is to have a singleton instantiation of signature scheme at program load time, for python it will always be the test one but this can abstract out this decision anyway

PR #109, opened 2025-11-02
feat: add de/serialization for xmss

anshalshukla:
## 🗒️ Description
Add serialization/ deserialization function for Public Key and Signature class. 

## 🔗 Related Issues or PRs
Splits PR #107

## ✅ Checklist
    ```console
    uvx tox
    ```


anshalshukla:
@tcoratger I think it's right to just split the PR #107 into multiple. I'm keeping the original PR open as it can serve as a reference.

I'm not sure if there is a more pythonic way to handle this.

Issue #113, opened 2025-11-03
tests: add tests that lexicographically higher block hash is chosen when fork weight is equal


unnawut:
Ref: https://github.com/leanEthereum/leanSpec/blob/c5796fd16b2abc10047f0b0d37f2dc5c6df05d29/src/lean_spec/subspecs/forkchoice/helpers.py#L70-L71

How can we guarantee that the label `"fork_a_2"` is "lexicographically" higher than `"fork_b_2"`?

_Originally posted by @syjn99 in https://github.com/leanEthereum/leanSpec/pull/105#discussion_r2485292881_

PR #117, opened 2025-11-04
Integrate xmss

anshalshukla:
## 🗒️ Description
Integrates XMSS Signature scheme.
- Fixes fork choice tests to avoid multiple blocks on same slot, as the proper_index is slot dependent
- Changed validator count in failing tests as the default validator count if unspecified is 4 which was leading the validators to overwrite their votes in prior slot, leading to nondeterministic results.

## 🔗 Related Issues or PRs
Superseeds #107 

## ✅ Checklist
    ```console
    uvx tox
    ```
- [X] Ran `uv` checks to validate forkchoice changes:
    ```
    uv run fill --fork=Devnet --clean
    ```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,4 +76,4 @@ jobs:
         run: uv sync --no-progress
 
       - name: Fill test fixtures
-        run: uv run fill --fork=Devnet --clean
+        run: uv run fill --fork=Devnet --clean -n auto
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -16,15 +16,17 @@ subspecifications that the Lean Ethereum protocol relies on.
 ## Development Workflow
 
 ### Running Tests
+
 ```bash
 uv sync                           # Install dependencies
 uv run pytest                     # Run unit tests
-uv run fill --fork=devnet --clean # Generate test vectors
+uv run fill --fork=devnet --clean -n auto # Generate test vectors
 # Note: execution layer support is planned for future, infrastructure is ready
 # for now, `--layer=consensus` is default and the only value used.
 ```
 
 ### Code Quality
+
 ```bash
 uv run ruff format src tests packages      # Format code
 uv run ruff check --fix src tests packages # Lint and fix
@@ -34,6 +36,7 @@ uvx tox                                    # Everything (checks + tests + docs)
 ```
 
 ### Common Tasks
+
 - **Main specs**: `src/lean_spec/`
 - **Subspecs**: `src/lean_spec/subspecs/{subspec}/`
 - **Unit tests**: `tests/lean_spec/` (mirrors source structure)
@@ -41,6 +44,7 @@ uvx tox                                    # Everything (checks + tests + docs)
 - **Execution spec tests**: `tests/execution/` (future - infrastructure ready)
 
 ## Code Style
+
 - Line length: 100 characters, type hints everywhere
 - Google docstring style (no docstrings for `__init__`)
 - Test files/functions must start with `test_`
@@ -54,15 +58,17 @@ uvx tox                                    # Everything (checks + tests + docs)
    - *Note: `tests/execution/` infrastructure is ready for future execution layer work*
 
 **Test Filling Framework:**
+
 - Layer-agnostic pytest plugin in `packages/testing/src/framework/pytest_plugins/filler.py`
 - Layer-specific packages: `consensus_testing` (active) and `execution_testing` (future)
 - Write consensus spec tests using `state_transition_test` or `fork_choice_test` fixtures
 - These fixtures are type aliases that create test vectors when called
-- Run `uv run fill --fork=Devnet --clean` to generate consensus fixtures
+- Run `uv run fill --fork=Devnet --clean -n auto` to generate consensus fixtures
 - Use `--layer=execution` flag when execution layer is implemented
 - Output goes to `fixtures/{layer}/{format}/{test_path}/...`
 
 **Example spec test:**
+
 ```python
 def test_block(state_transition_test: StateTransitionTestFiller) -> None:
     state_transition_test(
@@ -73,6 +79,7 @@ def test_block(state_transition_test: StateTransitionTestFiller) -> None:
 ```
 
 **How it works:**
+
 1. Test function receives a fixture class (not instance) as parameter
 2. Calling it creates a `FixtureWrapper` that runs `make_fixture()`
 3. `make_fixture()` executes the spec code (state transitions, fork choice steps)
@@ -81,12 +88,14 @@ def test_block(state_transition_test: StateTransitionTestFiller) -> None:
 6. Writes fixtures at session end to `fixtures/{layer}/{format}/{test_path}/...`
 
 **Layer-specific architecture:**
+
 - `framework/` - Shared infrastructure (base classes, pytest plugin, CLI)
 - `consensus_testing/` - Consensus layer fixtures, forks, builders
 - `execution_testing/` - Execution layer fixtures, forks, builders
 - Regular pytest runs (`uv run pytest`) ignore spec tests - they only run via `fill` command
 
 **Serialization requirements:**
+
 - All spec types (State, Block, Uint64, etc.) must be Pydantic models
 - Custom types need `@field_serializer` or `model_serializer` for JSON output
 - SSZ types typically serialize to hex strings (e.g., `"0x1234..."`)
@@ -97,6 +106,7 @@ def test_block(state_transition_test: StateTransitionTestFiller) -> None:
 - Test the serialization: `fixture.model_dump(mode="json")` must produce valid JSON
 
 **Key fixture types:**
+
 - `StateTransitionTest` - Tests state transitions with blocks
 - `ForkChoiceTest` - Tests fork choice with steps (tick/block/attestation)
 - Selective validation via `StateExpectation` and `StoreChecks` (only validates fields you specify)
@@ -113,6 +123,7 @@ def test_block(state_transition_test: StateTransitionTestFiller) -> None:
 When creating SSZ types, follow these established patterns:
 
 ### Domain-Specific Types (Preferred)
+
 - Use meaningful names that describe the purpose: `JustificationValidators`, `HistoricalBlockHashes`, `Attestations`
 - Define domain-specific types in modular structure (see Architecture section below)
 - Avoid generic names with numbers like `Bitlist68719476736` or `SignedAttestationList4096`
@@ -120,11 +131,13 @@ When creating SSZ types, follow these established patterns:
 ### SSZType vs SSZModel Design Decision
 
 **SSZType (IS-A pattern)**: Use for types that *are* data
+
 - Primitive scalars: `Uint64`, `Boolean`, `Bytes32`
 - These inherit directly from their underlying Python types
 - Example: `Uint64(42)` *is* the integer 42 with SSZ serialization
 
 **SSZModel (HAS-A pattern)**: Use for types that *have* data
+
 - Collections: `SSZList`, `SSZVector`, bitfields
 - Containers: `State`, `Block`, etc.
 - These use Pydantic models with a `data` field for contents
@@ -150,6 +163,7 @@ src/lean_spec/subspecs/containers/
 ```
 
 **Key principles:**
+
 - **Base types** (BaseBitlist, SSZList, etc.) stay in general scope (`src/lean_spec/types/`)
 - **Spec-specific types** go in their respective modules (`state/types.py`, `block/types.py`)
 - **Public API** exposed through `__init__.py` files for backward compatibility
@@ -158,6 +172,7 @@ src/lean_spec/subspecs/containers/
 ### Examples
 
 **Good domain-specific types:**
+
 ```python
 # In state/types.py
 HISTORICAL_ROOTS_LIMIT = 262144
@@ -174,6 +189,7 @@ class Attestations(SSZList):
 ```
 
 **Avoid generic types:**
+
 ```python
 # Don't do this:
 class Bitlist68719476736(BaseBitlist): ...
@@ -183,5 +199,6 @@ class SignedAttestationList4096(SSZList): ...
 ### API Compatibility
 
 When refactoring, maintain backward compatibility:
+
 - Keep existing import paths working through `__init__.py` exports
 - Preserve method signatures and behavior


packages/testing/src/consensus_testing/keys.py
```diff
@@ -0,0 +1,102 @@
...
+        if validator_index not in self.public_keys:
+            # Use max_slot + 1 as num_active_epochs since slots are used as epochs in the spec.
+            # +1 to include genesis slot
+            num_active_epochs = self.max_slot.as_int() + 1
+            self.public_keys[validator_index], self.secret_keys[validator_index] = (
+                self.schema.key_gen(0, num_active_epochs)
+            )
+        return self.public_keys[validator_index], self.secret_keys[validator_index]
+
+    def sign_attestation(self, attestation: Attestation) -> Signature:
```
g11tech, line 65:
```suggestion
    def sign_message(self, validator_id, message_bytes) -> Signature:
```

is the better API

packages/testing/src/consensus_testing/keys.py
```diff
@@ -0,0 +1,102 @@
...
+
+        Args:
+            attestation: The attestation to sign.
+
+        Returns:
+            A signature for the given attestation.
+        """
+        validator_id = attestation.validator_id
+
+        _, sk = self.get_or_create_key(validator_id)
```
g11tech: it shouldn't do creation here, it should assume the keys are there in manager

packages/testing/src/consensus_testing/keys.py
```diff
@@ -0,0 +1,102 @@
...
+                generated with enough capacity to sign messages up to this slot.
+                Defaults to 100 slots.
+            schema: The signature scheme to use. Defaults to TEST_SIGNATURE_SCHEME.
+        """
+        self.schema = schema
+        self.max_slot = max_slot
+        self.public_keys: dict[ValidatorIndex, PublicKey] = {}
+        self.secret_keys: dict[ValidatorIndex, SecretKey] = {}
+
+    def get_or_create_key(self, validator_index: ValidatorIndex) -> tuple[PublicKey, SecretKey]:
```
g11tech: just make this create so that one does the creation at the right time and at explicit places

packages/testing/src/consensus_testing/keys.py
anshalshukla: I need epoch number as well to sign the attestation

src/lean_spec/subspecs/containers/signature.py
```diff
@@ -2,20 +2,29 @@
...
-        return signature == Signature.zero()
-
-    @classmethod
-    def zero(cls) -> Signature:
-        """Return the zero (placeholder) signature."""
-        return cls(Bytes3100.zero())
+    def verify(self, public_key: PublicKey, epoch: int, message: bytes) -> bool:
+        """Verify the signature using XMSS verification algorithm."""
+        try:
+            if USE_TEST_SCHEME:
```
g11tech: instead of using if else, create a scheme instance in XmssSignature declaration file and directly import and use it here

packages/testing/src/consensus_testing/keys.py
```diff
@@ -0,0 +1,102 @@
...
+from lean_spec.subspecs.xmss.interface import TEST_SIGNATURE_SCHEME, GeneralizedXmssScheme
+from lean_spec.types import ValidatorIndex
+
+DEFAULT_MAX_SLOT = Slot(100)
+"""Default maximum slot for key generation."""
+DEFAULT_SCHEMA = TEST_SIGNATURE_SCHEME
+"""Default XMSS signature scheme."""
+
+
+class XmssKeyManager:
```
tcoratger, line 13: I guess this is cleaner to have something like this with the fields in the class directly

```python
DEFAULT_MAX_SLOT = Slot(100)
"""Default maximum slot for key generation."""

DEFAULT_SCHEMA = TEST_SIGNATURE_SCHEME
"""Default XMSS signature scheme."""


@dataclass(slots=True)
class XmssKeyManager:
    """
    Manage XMSS keys for test validators.

    Keys are generated lazily per validator and sized to be valid up to `max_slot`.
    """

    max_slot: Slot = DEFAULT_MAX_SLOT
    schema: GeneralizedXmssScheme = DEFAULT_SCHEMA
    _keypairs: Dict[ValidatorIndex, Tuple[PublicKey, SecretKey]] = field(
        default_factory=dict,
        init=False,
        repr=False,
    )

    def _derive_num_active_epochs(self) -> int:
        """
        Compute the number of active epochs for key generation.

        Slots are treated as epochs in the spec. We use max_slot + 1 to
        include the genesis slot.
        """
        return self.max_slot.as_int() + 1

    def get_or_create_key(
        self,
        validator_index: ValidatorIndex,
    ) -> Tuple[PublicKey, SecretKey]:
        """
        Return the XMSS key pair for the given validator, generating it if needed.
        """
        if validator_index not in self._keypairs:
            num_active_epochs = self._derive_num_active_epochs()
            pubkey, seckey = self.schema.key_gen(0, num_active_epochs)
            self._keypairs[validator_index] = (pubkey, seckey)

        return self._keypairs[validator_index]

    def sign_attestation(self, attestation: Attestation) -> Signature:
        """
        Sign an attestation with the validator's XMSS key.

        The XMSS epoch is taken from `attestation.data.slot`.
        """
        _, secret_key = self.get_or_create_key(attestation.validator_id)

        message = bytes(hash_tree_root(attestation))
        epoch = int(attestation.data.slot)
        xmss_signature = self.schema.sign(secret_key, epoch, message)

        signature_bytes = xmss_signature.to_bytes(self.schema.config)

        # Pad to 3100 bytes (Signature.LENGTH) with zeros on the right.
        # Padding only occurs with TEST_CONFIG (796 bytes) and not PROD_CONFIG (3100 bytes).
        padded_bytes = signature_bytes.ljust(Signature.LENGTH, b"\x00")

        return Signature(padded_bytes)

    def __contains__(self, validator_index: ValidatorIndex) -> bool:
        """
        Return True if the validator already has a registered key.
        """
        return validator_index in self._keypairs

    def __len__(self) -> int:
        """
        Return the number of validators with registered keys.
        """
        return len(self._keypairs)
```

src/lean_spec/runtime_config.py
```diff
@@ -0,0 +1,17 @@
...
+"""
+Global flag to control which XMSS signature scheme is used.
+
+When False (default): Uses production XMSS parameters (PROD_CONFIG)
+When True: Uses test XMSS parameters (TEST_CONFIG) for faster execution
+
+This flag is set by test frameworks in conftest.py to enable lighter-weight
+cryptographic operations during testing while keeping the spec code clean
+and independent of test-specific logic.
+"""
```
tcoratger: I think that we can do better than this. Having a general runtime thing like this is in general pretty bad. I guess the best option if to have something that the user can tune on demand.

src/lean_spec/subspecs/containers/signature.py
```diff
@@ -2,20 +2,29 @@
...
+                scheme = TEST_SIGNATURE_SCHEME
+                # TEST_CONFIG expects 796 bytes, but Signature is always 3100 bytes.
+                # Slice to the expected size for test config, assumes padding to the right.
+                signature_data = bytes(self)[: scheme.config.SIGNATURE_LEN_BYTES]
+                signature = XmssSignature.from_bytes(signature_data, scheme.config)
+            else:
+                scheme = PROD_SIGNATURE_SCHEME
+                signature = XmssSignature.from_bytes(bytes(self), scheme.config)
+            return bool(scheme.verify(public_key, epoch, message, signature))
```
tcoratger: Why do we need to convert to bool, is the result not a bool already?

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -261,23 +261,57 @@ def on_attestation(
...
+            except (ValueError, Exception):
+                return False
+
+            message = bytes(hash_tree_root(attestation))
+            epoch = attestation.data.slot.as_int()
+
+            if not signature.verify(pubkey, epoch, message):
+                return False
+
+        return True
```
tcoratger, line 310: We need unit tests via test vectors for this. If this is too complex, we can just have classic pytests and then open another PR to have test vectors.

tcoratger (commented):
In general it seems correct. Two points to be solved:
- We need to have unit tests for all the methods to ensure their correctness. Ideally we want test vectors when possible but if this is too complex due to lack of tools, I recommend doing classic pytest and opening one or more issues to have test vectors in the future.
- I think the run time constant for the configuration is in general a bad idea and we could do better (probably check how I did it inside the XMSS spec since I had to confront the same problems). That being said, I don't want it to be a blocker for this one to be merged as it should be a pretty important one for devnet1 so we can merge and create an issue to solve this (even if I preferred to do something clean from day 1).

src/lean_spec/runtime_config.py
anshalshukla: I'm handling it as per @g11tech suggestion

src/lean_spec/subspecs/forkchoice/store.py
anshalshukla: This function is already tested in `uv run fill --fork=Devnet --clean`, as the fork choice tests call this function. Earlier it was just making a naive zero byte comparison of signatures, now verification happens when `process_block` is called.

tcoratger: Ah yes that is true I forgot we did it like this in the past. Then no need for new tests here.

packages/testing/src/consensus_testing/keys.py
g11tech: hmm ok then sign_attestation makes sense

packages/testing/src/consensus_testing/keys.py
```diff
@@ -0,0 +1,101 @@
...
+        sk = self.secret_keys[validator_id]
+        message = bytes(hash_tree_root(attestation))
+        epoch = int(attestation.data.slot)
+        xmss_sig = DEFAULT_SIGNATURE_SCHEME.sign(sk, epoch, message)
+
+        signature_bytes = xmss_sig.to_bytes(DEFAULT_SIGNATURE_SCHEME.config)
+        # Pad to 3100 bytes (Signature.LENGTH) with zeros on the right
+        # Padding only occurs with TEST_CONFIG(796 bytes) and not PROD_CONFIG(3100 bytes).
+        padded_bytes = signature_bytes.ljust(Signature.LENGTH, b"\x00")
+        signature = Signature(padded_bytes)
```
tcoratger, line 86: Mhhhh here that is strange, I guess there is probably something not ok in the test config since we should not need to do this. We don't need to handle this here but I'll open an issue for this.

tcoratger:
I think that we just need to fix the conflicts and ready to merge for me. We can improve things later if needed, the important thing is that the features are here. Is it ok for both of you @g11tech and @unnawut?

packages/testing/src/consensus_testing/keys.py
```diff
@@ -0,0 +1,101 @@
...
+        Returns:
+            A tuple containing the public and secret keys for the given validator index.
+        """
+        if validator_index not in self.public_keys:
+            # Use max_slot + 1 as num_active_epochs since slots are used as epochs in the spec.
+            # +1 to include genesis slot
+            num_active_epochs = self.max_slot.as_int() + 1
+            self.public_keys[validator_index], self.secret_keys[validator_index] = (
+                DEFAULT_SIGNATURE_SCHEME.key_gen(0, num_active_epochs)
+            )
```
unnawut: Not sure do we need to introduce the max_slot concept if it's just num_active_epochs + 1? But not strong opinion here

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -103,7 +117,29 @@ def set_anchor_block_default(self) -> "ForkChoiceTest":
             )
         return self
 
-    def make_fixture(self) -> "ForkChoiceTest":
+    @model_validator(mode="after")
+    def set_max_slot_default(self) -> ForkChoiceTest:
```
unnawut, line 138: Not sure if this `max_slot` concept will work if we want to start varying the `activation_epoch` and `num_active_epochs` e.g. for bad validator config.

But not strong opinion and I guess we can deal with it when it comes, just noting down for now.

packages/testing/src/consensus_testing/keys.py
unnawut, line 62: Should the generated privkeys be saved to file somewhere along with the test vectors? I think otherwise clients cannot reproduce and match the state and block roots in the test vectors if they cannot recover the same generated keys.

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -125,6 +161,26 @@ def make_fixture(self) -> "ForkChoiceTest":
...
+        assert self.max_slot is not None, "max_slot must be set before make_fixture"
+
+        # Initialize XMSS key manager with max_slot for key generation
+        # Avoid adding excess buffer as it slows down test execution
+        key_manager = XmssKeyManager(max_slot=self.max_slot)
+
+        # Update validator pubkeys to match key_manager's generated keys
+        updated_validators = []
+        for i, validator in enumerate(self.anchor_state.validators):
+            pubkey, _ = key_manager.create_key_pair(ValidatorIndex(i))
```
unnawut: Nit: I'm not sure if it's just me but I think the name implies it's pure function (stateless, no side effects) while behind the scene it's actually storing the key pair in the manager before returning. Can we make this more visible? E.g.

```suggestion
            key_manager.create_key_pair(ValidatorIndex(i))
            pubkey, _ = key_manager.get_key_pair(ValidatorIndex(i))
```

or


```suggestion
            pubkey, _ = key_manager.create_and_store_key_pair(ValidatorIndex(i))
```

packages/testing/src/consensus_testing/keys.py
unnawut: Because of the much smaller [`LOG_LIFETIME = 8`](https://github.com/leanEthereum/leanSpec/blob/1e38ab2086be08da0f6e12fc08ae469635d9e616/src/lean_spec/subspecs/xmss/constants.py#L129) for testing config?

It would have worked for unit tests but for test vectors maybe we need prod xmss config but with very small `num_active_epochs` (which we're already doing)?

src/lean_spec/subspecs/containers/signature.py
```diff
@@ -4,18 +4,22 @@
...
+        """Verify the signature using XMSS verification algorithm."""
+        try:
+            scheme = DEFAULT_SIGNATURE_SCHEME
+            # Signature container is always 3100 bytes, but scheme config may expect less.
+            # Slice to the expected size if needed, assumes padding to the right.
+            signature_data = bytes(self)[: scheme.config.SIGNATURE_LEN_BYTES]
+            signature = XmssSignature.from_bytes(signature_data, scheme.config)
+            return scheme.verify(public_key, epoch, message, signature)
+        except Exception:
+            return False
```
unnawut, line 25: This is more severe than https://github.com/leanEthereum/leanSpec/pull/117/files#r2491285231 because it's part of the specs. But let's address separately.

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -261,23 +261,57 @@ def on_attestation(
             }
         )
 
-    def _validate_block_signatures(self, block: Block, signatures: BlockSignatures) -> bool:
+    def _validate_block_signatures(
+        self,
+        block: Block,
+        signatures: BlockSignatures,
+        proposer_attestation: Attestation,
```
```suggestion
        block: Block,
        proposer_attestation: Attestation,
        signatures: BlockSignatures,
```

or `signatures, block, proposer_attestation` but not in the middle

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -261,23 +261,57 @@ def on_attestation(
...
+    ) -> bool:
         """
         Validate block signatures.
 
-        Temporary stub for aggregated signature validation.
-
         Args:
             block: Block to validate.
             signatures: Block signatures to validate.
+            proposer_attestation: Attestation from the block proposer.
```
unnawut: Nit: just for consistency. Also rearrange if above suggestion is accepted

```suggestion
            proposer_attestation: Proposer's attestation to validate.
```

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -261,23 +261,57 @@ def on_attestation(
...
-        Note:
-            TODO: Integrate actual aggregated signature verification.
+            True if block signatures are valid, False otherwise.
         """
-        return all(Signature.is_valid(signature) for signature in signatures)
+        attestations = list(block.body.attestations)
+        attestations.append(proposer_attestation)
+
+        if len(signatures) != len(attestations):
+            return False
```
unnawut: Can we use assertions instead? Otherwise it'll be really hard for debugging especially when dealing with signatures

```suggestion
        assert len(signatures) == len(attestations), f"Signatures length ({len(signatures)}) does not match attestations length ({len(attestations)})"
```

Also applies to the rest of `return False`. There're way too many scenarios that `_validate_block_signatures()` can return False.

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -261,23 +261,57 @@ def on_attestation(
...
+            validator_id = attestation.validator_id.as_int()
+
+            if validator_id >= len(validators):
+                return False
+
+            validator = validators[validator_id]
+
+            try:
+                pubkey = validator.get_pubkey()
+            except (ValueError, Exception):
```
unnawut: My python is not strong but is this just simply `except:`?

Also this exception handling is needed because we're doing `xmss.PublicKey.from_bytes()` for every call to get the public key. We should sanitize this only once we created the validator container so we don't need to deal with this exception every call.

tests/consensus/devnet/fc/test_fork_choice_reorgs.py
```diff
@@ -421,9 +428,9 @@ def test_reorg_with_slot_gaps(
                     head_root_label="fork_a_3",
                 ),
             ),
-            # Fork B at slot 3 (competing, missed slot 2)
+            # Fork B at slot 4 (competing, missed slot 2-3)
             BlockStep(
-                block=BlockSpec(slot=Slot(3), parent_label="base", label="fork_b_3"),
+                block=BlockSpec(slot=Slot(4), parent_label="base", label="fork_b_4"),
```
unnawut, line 433: Hmm should signature integration change the behavior of fork choice tests?

unnawut (commented):
Some nits and maybe my own lack of python experience.

Most can go as new issues but if possible, I'd like to resolve the comment on the test changes in this PR if it's not related to this PR.

packages/testing/src/consensus_testing/keys.py
tcoratger: Yes I agree that this is a good idea, if too complex for this PR we can open an issue, but this is something we should handle.

tests/consensus/devnet/fc/test_fork_choice_reorgs.py
anshalshukla: Ideally no, but the issue here was that we were signing two different blocks at the same slot and treating them as a separate forks. Another testcase that I've modified was signing blocks till 9-10th slot and expecting the chain weights to be 6,3 but in default genesis config there are just 4 validators, so these validators were overwriting their own signature which was conceptually wrong.
There might be more test cases with such issues but these two are the ones that were failing sporadically.

packages/testing/src/consensus_testing/keys.py
tcoratger: Or maybe this is just a problem of some parameters which are not set correctly in the test configuration. I should check this.

anshalshukla: I had initially tried using PROD config but takes forever on my system even with small active epochs, I've specifically added function to calculate max epoch for this purpose, it shows considerable improvement even with TEST config

tests/consensus/devnet/fc/test_fork_choice_reorgs.py
tcoratger: Yes I'm a bit surprised by this, the old test didn't work when you added the feature with the signature?

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
anshalshukla: It's nothing essential as such because it will work fine even with default max slot, it's there as having a minimum active epochs helping me run the test in lesser time

packages/testing/src/consensus_testing/keys.py
tcoratger: Yes, that's why I had set up a very small test config to be able to run the tests, the CI and be able to check things. Otherwise I know that the PROD config key generation is already extra long in Rust so it would be totally impossible to run in Python in a reasonable amount of time.

tests/consensus/devnet/fc/test_fork_choice_reorgs.py
anshalshukla: It does, but sporadically and I think we need to revisit all the fork choice related testcases. I can take it up as separate issue. There are multiple issues that I have noticed, and I've fixed only the ones that were failing for me.

Broadly there are two issues in fork choice tests:
1. We try to check for head when the weights are equal. When the weights are equal the best child is determined using lexicographic ordering of hashing, it was consistent earlier but with signatures it becomes nondeterministic.
2. We don't have anchor_state in any of these tests, meaning the default genesis with 4 validators is generated. This is leading to wrong determination of chain weights in the tests as per the comments as most of the validator votes are being overwritten if tests go over slot 4. I have added `anchor_state` with 12 validators to handle the condition in `test_reorg_prevention_heavy_fork_resists_light_competition`. I did it because it was failing but I suspect there are multiple test cases were we have not handled this and have signed beyond slot 4 overwriting the validator signature to newer slot.

anshalshukla: I tried again without my changes and it works fine. I think it was failing earlier because at slot 8 (step-4) there is a tie and we try to check for best block there and it is dependent on hashing.

unnawut: > Ideally no, but the issue here was that we were signing two different blocks at the same slot and treating them as a separate forks.

Good catch!

> I tried again without my changes and it works fine. I think it was failing earlier because at slot 8 (step-4) there is a tie and we try to check for best block there and it is dependent on hashing.

If it's not required for this PR can we revert the test changes? The lexical order issue is tracked at #113

anshalshukla: No, I mean it passes but not consistently, it depends on how the signature is generated and eventually the block hash. If we revert my changes then we will have a nondeterministic CI which will depend on the hashes that are generated.

unnawut: Right got it!

packages/testing/src/consensus_testing/keys.py
g11tech: yes call be done in a followup, we can also accept an option seed to make it purely determistic
cc @anshalshukla

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
g11tech: I am good with the first suggestion i.e. create and then get because then create could totally be done at startup

src/lean_spec/subspecs/forkchoice/store.py
g11tech: why not send blockWithAttestation rather than separating, or may be signedBlockWithAttestation itself

unnawut: > why not send blockWithAttestation rather than separating, or may be signedBlockWithAttestation itself

Sgtm too since this function has a very specific use case

anshalshukla:
ugh, I just hope it doesn't timeout before it completes. Also, instead of generating keys for every test separately I think I should refactor it to reuse the keys. Adding another issue for this.

unnawut:
Reusing the keys within this PR sounds good. 10 minutes runtime will definitely impact productivity also locally

Maybe also cache the keys into the repo? Or a separate artifact repo that gets downloaded as part of testing? Then clients can also download this artifact for their test vectors execution too.

g11tech (approved):
yes we can probably add a pre_generated_key_manager instance as well, but this looks good for the moment

unnawut:
My local run of the PR branch uses about 12 minutes:

```
➜  leanSpec git:(integrate-xmss) git remote -v
origin  https://github.com/anshalshukla/leanSpec.git (fetch)
origin  https://github.com/anshalshukla/leanSpec.git (push)
➜  leanSpec git:(integrate-xmss) git branch --show-current
integrate-xmss
➜  leanSpec git:(integrate-xmss) uv run fill --fork=Devnet --clean
============================================================= test session starts =============================================================
platform darwin -- Python 3.14.0, pytest-8.4.2, pluggy-1.6.0
rootdir: /Users/unnawut/Development/sandbox/leanSpec
configfile: packages/testing/src/framework/cli/pytest_ini_files/pytest-fill.ini
testpaths: tests
plugins: xdist-3.8.0, cov-6.3.0, hypothesis-6.142.4
collected 29 items

tests/consensus/devnet/fc/test_attestation_processing.py ......
tests/consensus/devnet/fc/test_attestation_target_selection.py .....
tests/consensus/devnet/fc/test_fork_choice_head.py ......
tests/consensus/devnet/fc/test_fork_choice_reorgs.py ........
tests/consensus/devnet/stf_examples/test_blocks.py ...
tests/consensus/devnet/stf_examples/test_invalid.py .

======================================================= 29 passed in 705.04s (0:11:45) ========================================================
```

`-n auto` gives quite promising result (~5 mins) so personally I can work with this if we we're tackling pregen keys immediately after.

```
➜  leanSpec git:(integrate-xmss) uv run fill --fork=Devnet --clean -n auto
============================================================= test session starts =============================================================
platform darwin -- Python 3.14.0, pytest-8.4.2, pluggy-1.6.0
rootdir: /Users/unnawut/Development/sandbox/leanSpec
configfile: packages/testing/src/framework/cli/pytest_ini_files/pytest-fill.ini
testpaths: tests
plugins: xdist-3.8.0, cov-6.3.0, hypothesis-6.142.4
10 workers [29 items]
.............................
======================================================= 29 passed in 282.96s (0:04:42) ========================================================
```

Maybe we try parallelize in the CI and see the numbers one more time?

anshalshukla:
I think it is machine dependent, I can see in you `-n auto` run you have 10 workers but in case of CI there were only 4

tcoratger:
> > I think it is machine dependent, I can see in you `-n auto` run you have 10 workers but in case of CI there were only 4
> 
> Ah sorry I overlooked that `-n auto` is already in the PR so ~13 minutes is the number for CI parallelized.
> 
> I support merging this knowing that I can do full test suite in local in 5 minutes and pregen is coming up next. So we can start working on a few other improvements as well.

Ok no problem, if this is ok to you I'm fine, let's merge and fix the other thing then.

PR #118, opened 2025-11-04
stf: add `test_block_processing` test vector file

tcoratger:
## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/tests/consensus/devnet/state_transition/test_block_processing.py
+++ b/tests/consensus/devnet/state_transition/test_block_processing.py
@@ -83,7 +83,12 @@ def test_process_first_block_after_genesis(
         blocks=[
             BlockSpec(slot=Slot(1)),
         ],
-        post=StateExpectation(slot=Slot(1)),
+        post=StateExpectation(
+            slot=Slot(1),
+            latest_block_header_slot=Slot(1),
+            latest_block_header_state_root=Bytes32(b"\x00" * 32),
+            historical_block_hashes_count=1,
+        ),
     )
 
 
@@ -157,7 +162,12 @@ def test_blocks_with_gaps(
             BlockSpec(slot=Slot(4), parent_label="block_1", label="block_4"),
             BlockSpec(slot=Slot(8), parent_label="block_4", label="block_8"),
         ],
-        post=StateExpectation(slot=Slot(8)),
+        post=StateExpectation(
+            slot=Slot(8),
+            latest_block_header_slot=Slot(8),
+            latest_block_header_state_root=Bytes32(b"\x00" * 32),
+            historical_block_hashes_count=8,
+        ),
     )
 
 
--- a/packages/testing/src/consensus_testing/test_types/state_expectation.py
+++ b/packages/testing/src/consensus_testing/test_types/state_expectation.py
@@ -45,6 +45,15 @@ class StateExpectation(CamelModel):
     validator_count: int | None = None
     """Expected number of validators."""
 
+    latest_block_header_slot: Slot | None = None
+    """Expected slot of the latest block header."""
+
+    latest_block_header_state_root: Bytes32 | None = None
+    """Expected state root in the latest block header."""
+
+    historical_block_hashes_count: int | None = None
+    """Expected number of historical block hashes."""
+
     def validate_against_state(self, state: "State") -> None:
         """
         Validate this expectation against actual State.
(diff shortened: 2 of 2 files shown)


tests/consensus/devnet/state_transition/__init__.py
unnawut, line 1: Maybe good opportunity to start using full names? :)

tests/consensus/devnet/state_transition/...

tests/consensus/devnet/state_transition/test_block_processing.py
```diff
@@ -0,0 +1,341 @@
...
+
+Overview
+--------
+Tests for block header validation and block processing through the state transition
+function. Block processing is the core mechanism by which the blockchain state evolves,
+incorporating new transactions, attestations, and consensus decisions into the canonical state.
+
+What is Block Processing?
+--------------------------
+Block processing validates and integrates a proposed block into the beacon state:
```
unnawut: Just to avoid spec confusion

```suggestion
Block processing validates and integrates a proposed block into the chain's state:
```

tests/consensus/devnet/state_transition/test_block_processing.py
```diff
@@ -0,0 +1,341 @@
...
+**Phase 3: Body Processing**
+   Execute block body operations:
+   - Process attestations included in block
+   - Update justification/finalization
+   - Apply any other block operations
+
+**Phase 4: Root Verification**
+   Compute post-state root and verify it matches block state root.
+   This cryptographic commitment proves the state transition was
+   executed correctly.
```
unnawut, line 44: Was thinking the past few days how feasible we can maintain these docstrings in the long run? Not sure can we maintain this as part of the spec's docstrings instead of in the tests?

Also if we eventually implement some docgen, extracting from the specs is also easier to manage and structure than consolidating from all the tests.

Just an open question for discussion!

unnawut: But the docstrings for each test case, i.e. scenario description, expected behavior, why it matters, make sense to be in the test though

tests/consensus/devnet/state_transition/test_block_processing.py
```diff
@@ -0,0 +1,341 @@
...
+    5. Post-state at slot 1
+
+    This is the foundation for all subsequent blocks.
+    """
+    state_transition_test(
+        pre=generate_pre_state(),
+        blocks=[
+            BlockSpec(slot=Slot(1)),
+        ],
+        post=StateExpectation(slot=Slot(1)),
```
unnawut: Since it's listed in the Expected Behavior, right now it's only checking for "1. State advances from slot 0 to slot 1". Should we also assert these? 

    2. Block header is validated and processed
    3. Latest block header updated to new block
    4. Historical roots updated with genesis

Or at least check state.latest_block_header in addition to slot number to confirm the block has been consumed.

tests/consensus/devnet/state_transition/test_block_processing.py
```diff
@@ -0,0 +1,341 @@
...
+    This validates resilience to gaps.
+    """
+    state_transition_test(
+        pre=generate_pre_state(),
+        blocks=[
+            BlockSpec(slot=Slot(1), label="block_1"),
+            BlockSpec(slot=Slot(4), parent_label="block_1", label="block_4"),
+            BlockSpec(slot=Slot(8), parent_label="block_4", label="block_8"),
+        ],
+        post=StateExpectation(slot=Slot(8)),
```
unnawut: Similar to above, a quick latest_block_header and/or post-state root output would be nice. Currently progressing slots without any blocks would also pass this test?

tests/consensus/devnet/state_transition/test_block_processing.py
```diff
@@ -0,0 +1,341 @@
...
+    #
+    # For slot 1:
+    # - expected proposer is ValidatorIndex(1),
+    # - we'll try to use ValidatorIndex(5) instead
+    state_transition_test(
+        pre=generate_pre_state(),
+        blocks=[
+            BlockSpec(
+                slot=Slot(1),
+                proposer_index=ValidatorIndex(5),  # Wrong proposer
```
unnawut: If the test is configured to 4 validators, can we use something in range just to be sure the error is not because the validator input is out of bound?

```suggestion
                proposer_index=ValidatorIndex(3),  # Wrong proposer
```

tests/consensus/devnet/state_transition/__init__.py
tcoratger: Yep fixed. Once I'll remove the pytest forkchoice folder, I'll be able to rename the other one properly as well otherwise we have some conflicts unfortunately.

tests/consensus/devnet/state_transition/test_block_processing.py
tcoratger: Fixed, just added very basic tests, just counting the length of `state.historical_block_hashes` for example. We can go much deeper in tests in followup PRs when needed, building stronger testing tools for this.

tests/consensus/devnet/state_transition/test_block_processing.py
tcoratger: fixed

tests/consensus/devnet/state_transition/test_block_processing.py
tcoratger: Yes, I agree. I was initially aiming for one or two files to be very precise, but I realize there could be many, creating duplicates with the existing documentation in the specification.

In a follow-up pull request, I'll clean this up, especially the header documentation for each file, keeping only the bare minimum necessary for understanding. The same goes for the test functions—just the essentials for client teams to understand things after the deserialization process, but nothing more, to avoid excessive maintenance work on this documentation when we change a single test or item.

What's important is to be very precise and comprehensive with the documentation within the specification itself. For this, I usually check with AI using in-depth technical content, blog posts, and research articles to ensure the best possible technical documentation and that nothing is overlooked.

unnawut (approved):
Thank you!

PR #125, opened 2025-11-04
test: rm old test slot

tcoratger:
## 🗒️ Description

@unnawut can you confirm that after https://github.com/leanEthereum/leanSpec/pull/123 we can remove this file? seems like a duplicate of your complete suite of test vectors and we want to rely only on the test vectors.

## ✅ Checklist

    ```console
    uvx tox
    ```


unnawut:
> @unnawut can you confirm that after https://github.com/leanEthereum/leanSpec/pull/123 we can remove this file? seems like a duplicate of your complete suite of test vectors and we want to rely only on the test vectors.

The test vectors in #123 covered at least a single occurrence of all 3 rules, but seeing this file I think it's covering too few occurrences to be confident. I'll expand that test vector today but agree we can close this one anyway.

unnawut (approved):

PR #127, opened 2025-11-05
Fix floating-point precision bug

bomanaps:
## 🗒️ Description
Replaces floating-point arithmetic with `math.isqrt()`

## ✅ Checklist

    ```console
    uvx tox
    ```


src/lean_spec/subspecs/containers/slot.py
```diff
@@ -44,13 +45,16 @@ def is_justifiable_after(self, finalized_slot: Slot) -> bool:
...
             # Pronic numbers have the form n(n+1): 2, 6, 12, 20, 30, 42, 56, ...
             # Mathematical insight: For pronic delta = n(n+1), we have:
-            #   delta + 0.25 = n^2 + n + 0.25 = (n + 0.5)^2
-            #   Therefore sqrt(delta + 0.25) = n + 0.5
-            or ((delta + 0.25) ** 0.5) % 1 == 0.5
+            #   4*delta + 1 = 4n(n+1) + 1 = (2n+1)^2
+            # Check: 4*delta+1 is an odd perfect square
+            or (
+                math.isqrt(4 * delta + 1) ** 2 == 4 * delta + 1
+                and math.isqrt(4 * delta + 1) % 2 == 1
```
unnawut, line 58: The perfect square looks good to me! But the pronic number check is quite different from the previous version. Please allow me some time to think more on this within the next hour!

tcoratger: I guess a solution for the whole function should be like

```python
# Ensure the candidate slot is not before the finalized slot.
assert self >= finalized_slot, "Candidate slot must not be before finalized slot"

# Calculate the distance in slots from the last finalized slot.
delta = (self - finalized_slot).as_int()


# Rule 1: The first 5 slots after finalization are always justifiable.
#
# Examples: delta = 0, 1, 2, 3, 4, 5
if delta <= 5:
    return True

def _is_perfect_square(n: int) -> bool:
    """Checks if n is a perfect square using only integer arithmetic."""
    if n < 0:
        return False
    # Computes the integer part of the square root
    root = math.isqrt(n)
    # If n is a perfect square, its root squared will equal n
    return root * root == n

return (
    # Rule 2: Slots at perfect square distances are justifiable.
    #
    # Examples: delta = 1, 4, 9, 16, 25, 36, 49, 64, ...
    # Check: The integer square root of delta, when squared,
    #        must equal delta itself.
    _is_perfect_square(delta)
    # Rule 3: Slots at pronic number distances are justifiable.
    #
    # Pronic numbers have the form n(n+1): 2, 6, 12, 20, 30, ...
    # Mathematics:
    #   delta = n^2 + n
    #   4*delta + 1 = 4n^2 + 4n + 1 = (2n + 1)^2
    # Thus, delta is pronic if and only if (4*delta + 1) is
    # a perfect square.
    or _is_perfect_square(4 * delta + 1)
)
```

unnawut: @tcoratger Looks like we need the _odd_ perfect square check as well which is the 2nd part: `math.isqrt(4 * delta + 1) % 2 == 1` So actually the original PR proposal lgtm!

unnawut (approved):
Lgtm!

PR #131, opened 2025-11-06
chore: use JUSTIFICATION_LOOKBACK_SLOTS

unnawut:
## 🗒️ Description

[`JUSTIFICATION_LOOKBACK_SLOTS`](https://github.com/leanEthereum/leanSpec/blob/main/src/lean_spec/subspecs/chain/config.py#L26) was defined since devnet0 but the spec still uses the hard coded value `range(3)` for justification lookback. This PR updates it to use `JUSTIFICATION_LOOKBACK_SLOTS`

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/src/lean_spec/subspecs/forkchoice/store.py
+++ b/src/lean_spec/subspecs/forkchoice/store.py
@@ -726,7 +726,8 @@ class Store(Container):
         ensuring the target is in a justifiable slot range:
 
         1. **Start at Head**: Begin with the current head block
-        2. **Walk Toward Safe**: Move backward (up to 3 steps) if safe target is newer
+        2. **Walk Toward Safe**: Move backward (up to `JUSTIFICATION_LOOKBACK_SLOTS` steps)
+           if safe target is newer
         3. **Ensure Justifiable**: Continue walking back until slot is justifiable
         4. **Return Checkpoint**: Create checkpoint from selected block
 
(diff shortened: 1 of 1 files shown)


varun-doshi (approved):
Nice catch

syjn99 (approved):

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -749,7 +750,7 @@ def get_attestation_target(self) -> Checkpoint:
         #
         # This ensures the target doesn't advance too far ahead of safe target,
         # providing a balance between liveness and safety.
-        for _ in range(3):
+        for _ in range(JUSTIFICATION_LOOKBACK_SLOTS):
```
tcoratger, line 754: Just two lines before we should have `JUSTIFICATION_LOOKBACK_SLOTS` in the comment as well no?

```python
# Walk back toward safe target (up to `JUSTIFICATION_LOOKBACK_SLOTS` steps)
#
# This ensures the target doesn't advance too far ahead of safe target,
# providing a balance between liveness and safety.
for _ in range(JUSTIFICATION_LOOKBACK_SLOTS):
```

tcoratger (approved):

src/lean_spec/subspecs/forkchoice/store.py
unnawut: Ah good catch! I also spotted another occurrence in the docstring. Fixed

PR #137, opened 2025-11-06
test: add tests for genesis

tcoratger:
## 🗒️ Description

I realized that the state transition function tests lacked a simple test to test all the fields of the genesis setup.

So I added the necessary tools and a couple of tests to properly test this basic element.

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/packages/testing/src/consensus_testing/test_types/state_expectation.py
+++ b/packages/testing/src/consensus_testing/test_types/state_expectation.py
@@ -3,6 +3,12 @@
 from typing import TYPE_CHECKING
 
 from lean_spec.subspecs.containers.slot import Slot
+from lean_spec.subspecs.containers.state.types import (
+    HistoricalBlockHashes,
+    JustificationRoots,
+    JustificationValidators,
+    JustifiedSlots,
+)
 from lean_spec.types import Bytes32, CamelModel
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


unnawut:
Thanks for this! I have some questions around the usage of these test vectors. Let me know if I misunderstood something though.

I checked the state transition test fixture and it [looks like a no-op if blocks are not provided](https://github.com/leanEthereum/leanSpec/blob/28b8d61e85f0c37827229f0cf0f14d19360ff7d1/packages/testing/src/consensus_testing/test_fixtures/state_transition.py#L134). So it looks to me the tests are more of internal tests for `generate_pre_state()` than client test vectors. We can't test a state_transition without running a block, and if we run a block then we aren't at genesis.

Maybe instead we test how clients are expected to consume external inputs to create their genesis state? E.g. the clients are expected to consume some [common files from the genesis generator](https://github.com/ethpandaops/eth-beacon-genesis/pull/36#issuecomment-3263264459) to build the genesis state. So it would be nice to test genesis by testing a generated `config.yml` against the expected state instead?

unnawut:
> However, when the test is initialized via `generate_pre_state`, it actually calls our `State.generate_genesis`, which is what I want to test here:
> 
> [https://github.com/leanEthereum/leanSpec/blob/28b8d61e85f0c37827229f0cf0f14d19360ff7d1/packages/testing/src/consensus_testing/test_types/genesis.py#L31](https://github.com/leanEthereum/leanSpec/blob/28b8d61e85f0c37827229f0cf0f14d19360ff7d1/packages/testing/src/consensus_testing/test_types/genesis.py#L31)

Ahh I see I missed that important last line in the code.

Then I think my only other question here is how should the client consume this test vector because triggering State.generate_genesis() is happening inside the test fixture. In the test vector, the client will already see the generated genesis state as the pre-state, not the raw genesis_time and validators as input?

I’m not in front of my computer right now but let me try generate the vectors in a few minutes

jihoonsong:
I think not all tests should produce test vectors for clients to consume. For instance, the CL spec has two types of test: a unit tests that checks the spec itself and a test that produces test vectors (or reference tests according to the CL spec parlance). To my understanding, this PR adds the prior and I think it's quite significant given its context. I also think we would have and appreciate more tests like this down the road.

> Maybe instead we test how clients are expected to consume external inputs to create their genesis state?

That being said, +1 for this point. Probably the best way to go about this could be letting clients to implement such tests at their discretion. They can set up input and output according to the agreed-upon data. The spec doesn't need to aware of such tests as it's implementation details.

unnawut (approved):
From pqinterop call on Nov 12 I think we're all cool with this. Thanks @tcoratger!

PR #145, opened 2025-11-07
forkchoice tests: small touchup

tcoratger:
## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
+++ b/packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
@@ -215,7 +215,6 @@ class ForkChoiceTest(BaseConsensusFixture):
         )
 
         # Block registry for label-based fork creation
-        #
         # Register genesis/anchor block with implicit label
         self._block_registry: dict[str, Block] = {"genesis": self.anchor_block}
 


scripts/pregenerate_keys.py
```diff
@@ -0,0 +1,82 @@
...
+
+    # Write keys to JSON file
+    output_path = fixtures_dir / "pregenerated_xmss_keys.json"
+    with open(output_path, "w") as f:
+        json.dump(keys_data, f, indent=2)
+
+    print(f"\n✓ Successfully generated {num_validators} keys")
+    print(f"✓ Saved to: {output_path}")
+    print(f"✓ File size: {output_path.stat().st_size / 1024 / 1024:.2f} MB")
+    print("\nKeys will be automatically loaded during test execution.")
```
unnawut: Just for my own learning. How is the automatic key loading done? I couldn't find the relationship in the PR

unnawut (approved):
🙇‍♂️🙇‍♂️🙇‍♂️

scripts/pregenerate_keys.py
tcoratger: Oops, sorry, that was for another PR I'm working on (I mixed up the branches). It involves having a script that generates the key to store in a file so they can be reused and avoid the current huge time commitment of the CI.

But I still need to revisit this in a follow-up PR; I'm not happy with my current solution.

unnawut (approved):

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -211,10 +215,9 @@ def make_fixture(self) -> ForkChoiceTest:
         )
 
         # Block registry for label-based fork creation
-        self._block_registry: dict[str, Block] = {}
-
+        #
```
unnawut:
```suggestion
```

PR #146, opened 2025-11-07
store: cleanup and more doc for `_validate_block_signatures`

tcoratger:
## ✅ Checklist

    ```console
    uvx tox
    ```


unnawut (approved):
Quick note that we also want to move validate_block_signatures out of the fork choice store but we can also do that as a separate issue.

PR #148, opened 2025-11-07
test: add test_empty_block_processing tests

### LINKED ISSUE #121: tests: add `test_empty_block_processing` tests
After https://github.com/leanEthereum/leanSpec/pull/118 is merged we should add a test like

```python
def test_empty_block_processing(
    state_transition_test: StateTransitionTestFiller,
) -> None:
    """
    Test processing blocks with empty body (no attestations).

    Scenario
    --------
    Create blocks with no attestations in the body, only header.

    """
```

varun-doshi:
## 🗒️ Description

Adds tests to check that the chain progresses even with blocks with empty body

## 🔗 Related Issues or PRs

Closes #121 

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/tests/consensus/devnet/state_transition/test_block_processing.py
+++ b/tests/consensus/devnet/state_transition/test_block_processing.py
@@ -307,35 +307,40 @@ def test_block_extends_deep_chain(
     )
 
 
-def test_empty_blocks_in_the_middle(
+def test_empty_blocks(
     state_transition_test: StateTransitionTestFiller,
 ) -> None:
     """
-        Test processing blocks with empty body (no attestations).
+    Test processing blocks with empty body (no attestations).
 
-        Scenario
-    #     --------
-    #     Build chain of blocks with empty body:
-    #     - Slot 1: Block, Empty body
-    #     - Slot 2: Block, Empty body
-    #     - Slot 3: BLock, Empty body
-    #     - Slot 4: Block, Empty body
-    #     - Slot 5: Block, Empty body
-    #     - Slot 6: Block, Empty body
+    Scenario
+    --------
+    Build chain of blocks with empty body:
+    - Slot 1: Block, Empty body
+    - Slot 2: Block, Empty body
+    - Slot 3: Block, Empty body
+    - Slot 4: Block, Empty body
+    - Slot 5: Block, Empty body
+    - Slot 6: Block, Empty body
 
-    #     Expected Behavior
-    #     -----------------
-    #     1. Blocks process as expected
-    #     2. State advances to slot 6
+    Expected Behavior
+    -----------------
+    1. Blocks process as expected
+    2. State advances to slot 6
 
     """
-    blocks = [BlockSpec(slot=Slot(1), body=None, label="block_1")]
-    blocks.extend(
-        [
-            BlockSpec(slot=Slot(i), body=None, parent_label=f"block_{i - 1}", label=f"block_{i}")
-            for i in range(2, 7)
-        ]
-    )
+    blocks = [
+        BlockSpec(slot=Slot(1), body=None, label="block_1"),
+        *[
+            BlockSpec(
+                slot=Slot(slot),
+                body=None,
+                parent_label=f"block_{slot - 1}",
+                label=f"block_{slot}",
+            )
+            for slot in range(2, 7)
+        ],
+    ]
 
     state_transition_test(
         pre=generate_pre_state(),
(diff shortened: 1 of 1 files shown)


tests/consensus/devnet/state_transition/test_block_processing.py
```diff
@@ -305,3 +305,81 @@ def test_block_extends_deep_chain(
...
+    #     - Slot 2: Block, Empty body
+    #     - Slot 3: BLock, Empty body
+    #     - Slot 4: Block, Empty body
+    #     - Slot 5: Block, Empty body
+    #     - Slot 6: Block, Empty body
+
+    #     Expected Behavior
+    #     -----------------
+    #     1. Blocks process as expected
+    #     2. State advances to slot 6
```
tcoratger:
```suggestion
        --------
        Build chain of blocks with empty body:
        - Slot 1: Block, Empty body
        - Slot 2: Block, Empty body
        - Slot 3: Block, Empty body
        - Slot 4: Block, Empty body
        - Slot 5: Block, Empty body
        - Slot 6: Block, Empty body

        Expected Behavior
        -----------------
        1. Blocks process as expected
        2. State advances to slot 6
```

tests/consensus/devnet/state_transition/test_block_processing.py
```diff
@@ -305,3 +305,81 @@ def test_block_extends_deep_chain(
...
+
+    """
+    blocks = [BlockSpec(slot=Slot(1), body=None, label="block_1")]
+    blocks.extend(
+        [
+            BlockSpec(slot=Slot(i), body=None, parent_label=f"block_{i - 1}", label=f"block_{i}")
+            for i in range(2, 7)
+        ]
+    )
```
tcoratger:
```suggestion
    blocks = [
        BlockSpec(slot=Slot(1), body=None, label="block_1"),
        *[
            BlockSpec(
                slot=Slot(slot),
                body=None,
                parent_label=f"block_{slot - 1}",
                label=f"block_{slot}",
            )
            for slot in range(2, 7)
        ],
    ]
```

tests/consensus/devnet/state_transition/test_block_processing.py
```diff
@@ -305,3 +305,81 @@ def test_block_extends_deep_chain(
...
+    #     - Slot 3: BLock, Empty body
+    #     - Slot 4: Missed
+    #     - Slot 5: Block, Empty body
+    #     - Slot 6: Block
+
+    #     Expected Behavior
+    #     -----------------
+    #     1. Blocks process at specified slots
+    #     2. Empty slots handled automatically
+    #     3. State advances to slot 6
```
tcoratger:
```suggestion
        Scenario
        --------
        Build chain of blocks with empty body + missed slot:
        - Slot 1: Block
        - Slot 2: Block, Empty body
        - Slot 3: BLock, Empty body
        - Slot 4: Missed
        - Slot 5: Block, Empty body
        - Slot 6: Block

        Expected Behavior
        -----------------
        1. Blocks process at specified slots
        2. Empty slots handled automatically
        3. State advances to slot 6
```

tests/consensus/devnet/state_transition/test_block_processing.py
```diff
@@ -305,3 +305,86 @@ def test_block_extends_deep_chain(
...
+        blocks=[
+            BlockSpec(slot=Slot(1), label="block_1"),
+            BlockSpec(slot=Slot(2), body=None, parent_label="block_1", label="block_2"),
+            BlockSpec(slot=Slot(3), body=None, parent_label="block_2", label="block_3"),
+            # slot = 4 missed
+            BlockSpec(slot=Slot(5), body=None, parent_label="block_3", label="block_5"),
+            BlockSpec(slot=Slot(6), parent_label="block_5", label="block_6"),
+        ],
+        post=StateExpectation(
+            slot=Slot(6), latest_block_header_slot=Slot(6), historical_block_hashes_count=6
```
unnawut, line 388: I notice both this test and the previous both go up to `Slot(6)`, and both have `historical_block_hashes_count=6`

But this test is skipping a block so I'm not sure why the count is 6 (I don't have an answer though)

varun-doshi: Interestingly something similar is happening here as well
https://github.com/leanEthereum/leanSpec/blob/a0842043b1fb588abc925269955cb9935488d418/tests/consensus/devnet/state_transition/test_block_processing.py#L83

varun-doshi: I think this is what is happening:
If there were empty slots between the previous block and this block, a number of ZERO_HASH entries are also appended to represent those skipped slots.
Here:
https://github.com/leanEthereum/leanSpec/blob/a0842043b1fb588abc925269955cb9935488d418/src/lean_spec/subspecs/containers/state/state.py#L355

jihoonsong (approved):

tests/consensus/devnet/state_transition/test_block_processing.py
tcoratger: Yep, in the tests we currently have a bunch of count checks, and I don't really like that. I implemented it as a quick fix to easily test certain behaviors, but it's not ideal, and we should try to fix it in the future (it's not at all the fault of this PR).

Indeed, this PR is a very good example because relying on the counter hides the zeros. It would be more powerful and interesting, for example, to test the hashes directly. But that's another story, and this type of behavior needs to be addressed in the future; I need to think about the best way to do it.

PR #150, opened 2025-11-08
fix(tests): Missing blocks in generated test fixtures

### LINKED ISSUE #144: Missing `blocks` in generated test fixtures
The generated `test_block_with_invalid_proposer.json` and `test_block_with_invalid_parent_root.json` missing `blocks` value, currently it is an empty array which is inconsistent with python test.

varun-doshi:
## 🗒️ Description

The generated `test_block_with_invalid_proposer.json` and `test_block_with_invalid_parent_root.json` missing blocks value, currently it is an empty array which is inconsistent with python test.
The reason was the `make_fixtures()` internally calls `process_block()` from the actual specs. When the functions hits the `invalid_proposer` or `invalid_parent_root` it will correctly reject the block. Hence, it was not creating any block data in the json files.
The fix: Check if we are using an `invalid_proposer` or `invalid_parent_root` and if so return the `temp_block` before running `process_block()`

## 🔗 Related Issues or PRs

Fixes #144 

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/packages/testing/src/consensus_testing/test_fixtures/state_transition.py
+++ b/packages/testing/src/consensus_testing/test_fixtures/state_transition.py
@@ -197,25 +197,18 @@ class StateTransitionTest(BaseConsensusFixture):
         Block
             A complete block ready for state_transition.
         """
-        is_correct_proposer_index = True
         # Use provided proposer_index or compute it
         if spec.proposer_index is not None:
             proposer_index = spec.proposer_index
-            is_correct_proposer_index = spec.proposer_index == ValidatorIndex(
-                int(spec.slot) % int(state.validators.count)
-            )
         else:
             proposer_index = ValidatorIndex(int(spec.slot) % int(state.validators.count))
 
-        is_corect_parent_root = True
-        temp_state = state.process_slots(spec.slot)
-        computed_parent_root = hash_tree_root(temp_state.latest_block_header)
         # Use provided parent_root or compute it
         if spec.parent_root is not None:
             parent_root = spec.parent_root
-            is_corect_parent_root = parent_root == computed_parent_root
         else:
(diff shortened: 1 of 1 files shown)


packages/testing/src/consensus_testing/test_fixtures/state_transition.py
```diff
@@ -198,18 +197,25 @@ def _build_block_from_spec(self, spec: BlockSpec, state: State) -> Block:
...
         # Use provided proposer_index or compute it
         if spec.proposer_index is not None:
             proposer_index = spec.proposer_index
+            is_correct_proposer_index = spec.proposer_index == ValidatorIndex(
+                int(spec.slot) % int(state.validators.count)
+            )
         else:
             proposer_index = ValidatorIndex(int(spec.slot) % int(state.validators.count))
 
+        is_corect_parent_root = True
```
unnawut: Also

```suggestion
        is_correct_parent_root = True
```

jihoonsong:
`ForkChoiceTest` also has `_build_block_from_spec()`. Do we need to add the same code there as well?

varun-doshi:
> `ForkChoiceTest` also has `_build_block_from_spec()`. Do we need to add the same code there as well?

`ForkChoiceTest` actually doesn't have a `expect_exception` in its own structure unlike `StateTransitionTest` so if we want to do this we can add it and then make the same check.
However, I'd  recommend if we were to do this, we should do it in a separate PR since the missing data issue is currently only in `StateTransitionTest` so we should get that patch in so that client teams can work on consuming the vectors. And then we can add the same validation to `ForkChoiceTest`.
What do you think?

jihoonsong (approved):
Sounds good, we can focus on current scope.

PR #151, opened 2025-11-08
test: fix validators in fork choice reorg tests

tcoratger:
## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
+++ b/packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
@@ -239,11 +239,11 @@ class ForkChoiceTest(BaseConsensusFixture):
                     block = signed_block.message.block
                     step._filled_block = block
 
-                    # Validate that the same validator doesn't propose competing blocks
-                    # at the same slot
                     if block.slot not in slot_proposers:
                         slot_proposers[block.slot] = set()
 
+                    # Validate that the same validator doesn't propose competing blocks
+                    # at the same slot
                     if block.proposer_index in slot_proposers[block.slot]:
                         raise ValueError(
                             f"Step {i}: Validator {block.proposer_index} is equivocating "


packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -235,6 +239,23 @@ def make_fixture(self) -> ForkChoiceTest:
                     block = signed_block.message.block
                     step._filled_block = block
 
+                    # Validate that the same validator doesn't propose competing blocks
+                    # at the same slot
+                    if block.slot not in slot_proposers:
+                        slot_proposers[block.slot] = set()
+
+                    if block.proposer_index in slot_proposers[block.slot]:
```
jihoonsong:
```suggestion
                    if block.slot not in slot_proposers:
                        slot_proposers[block.slot] = set()

                    # Validate that the same validator doesn't propose competing blocks
                    # at the same slot
                    if block.proposer_index in slot_proposers[block.slot]:
```

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -235,6 +239,23 @@ def make_fixture(self) -> ForkChoiceTest:
...
+                        raise ValueError(
+                            f"Step {i}: Validator {block.proposer_index} is proposing multiple "
```
jihoonsong: We could make this description more concise, something like "Step {i}: Validator {block.proposer_index} is equivocating at slot {block.slot}.\n"

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -235,6 +239,23 @@ def make_fixture(self) -> ForkChoiceTest:
...
+                            f"blocks at slot {block.slot}. This creates competing forks from "
+                            f"the same validator, which is invalid.\n"
+                            f"Solution: Either add more validators to the test or explicitly "
```
jihoonsong: Could you elaborate how two solutions laid out here can actually prevent equivocation from happening?

tests/consensus/devnet/fc/test_fork_choice_reorgs.py
```diff
@@ -247,6 +253,9 @@ def test_three_block_deep_reorg(
     about chain history, ensuring safety and liveness even in adversarial scenarios.
     """
     fork_choice_test(
+        # anchor_state=generate_pre_state(
```
jihoonsong, line 256: Any reason this is commented out?

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
tcoratger: fix

Issue #128, opened 2025-11-05
refactor: fork choice store


unnawut:
I'm not too confident I get this right at all but creating this issue for tracking & discussion...

Fork choice store is carrying a lot of responsibilities that are beyond fork choice decision and we should move these out of it:

- [x] `forkchoice.store._validate_block_signatures()` -> should already be validated before entering fork choice
  - Done by #147
- [x] `forkchoice.store._process_block_body_attestations()` -> fork choice should only know of attestations through `store.on_attestation()`, maybe triggered by `state.process_attestation()`
    - Done by #153
- [x] `forkchoice.store._process_proposer_attestation()` -> fork choice should only know of attestations through `store.on_attestation()`, maybe triggered by `state.process_block()`
   - Done by #155 
- [ ] `produce_block_with_signatures()` -> fork choice should only need to provide `get_proposal_head()`
- [x] `produce_attestation()` -> fork choice should only need to provide `get_attestation_target()`
    - Done by #188

Thanks to @jihoonsong for raising this earlier in group chats

PR #153, opened 2025-11-09
`SignedBlockWithAttestation`: add `process_block_body_attestations`

tcoratger:
## 🔗 Related Issues or PRs

Related https://github.com/leanEthereum/leanSpec/issues/128

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/src/lean_spec/subspecs/containers/block/block.py
+++ b/src/lean_spec/subspecs/containers/block/block.py
@@ -16,11 +16,10 @@ from lean_spec.subspecs.ssz.hash import hash_tree_root
 from lean_spec.types import Bytes32, Uint64
 from lean_spec.types.container import Container
 
-from ..attestation import Attestation, SignedAttestation
+from ..attestation import Attestation
 from .types import Attestations, BlockSignatures
 
 if TYPE_CHECKING:
-    from ...forkchoice.store import Store
     from ..state import State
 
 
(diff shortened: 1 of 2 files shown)


src/lean_spec/subspecs/containers/block/block.py
```diff
@@ -182,3 +183,36 @@ def verify_signatures(self, parent_state: "State") -> bool:
...
+            self.message.block.body.attestations, self.signature, strict=False
+        ):
+            # Process as on-chain attestation (immediately becomes "known")
+            store = store.on_attestation(
+                signed_attestation=SignedAttestation(
+                    message=attestation,
+                    signature=signature,
+                ),
+                is_from_block=True,
+            )
```
unnawut: Sorry I'm second-thinking this. Since it's only job is to loop `store.on_attestation()` and only used by the `store.on_block`... Maybe we don't need the docstring and inlining makes a lot more sense. Wdyt? Proposed change below.

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -419,8 +383,8 @@ def on_block(self, signed_block_with_attestation: SignedBlockWithAttestation) ->
             }
         )
 
-        # Process attestations immutably (chain method calls)
-        store = store._process_block_body_attestations(block, signatures)
+        # Process block body attestations
+        store = signed_block_with_attestation.process_block_body_attestations(store)
```
```suggestion
        # Iterate over attestations and their corresponding signatures.
        for attestation, signature in zip(
            block.message.block.body.attestations, block.signature, strict=False
        ):
            # Process as on-chain attestation (immediately becomes "known")
            store = store.on_attestation(
                signed_attestation=SignedAttestation(
                    message=attestation,
                    signature=signature,
                ),
                is_from_block=True,
            )
```

src/lean_spec/subspecs/containers/block/block.py
tcoratger: Ah yes, you're right, I refactored the function `process_block_body_attestations` maximally with the zip thing, so it's true that we don't actually need it anymore. It's simpler this way, and we don't have to add unnecessary functions. So much better like this, I agree.

unnawut (approved):

KolbyML:
@tcoratger should the length of signatures always be equal to the length of attestations?

```
        for attestation, signature in zip(
            signed_block_with_attestation.message.block.body.attestations,
            signed_block_with_attestation.signature,
            strict=False,
        ):
 ```
 
^ if this code contains a duplicate attestation wouldn't that be considered a bug?

tcoratger:
> @tcoratger should the length of signatures always be equal to the length of attestations?
> 
> ```
>         for attestation, signature in zip(
>             signed_block_with_attestation.message.block.body.attestations,
>             signed_block_with_attestation.signature,
>             strict=False,
>         ):
> ```
> 
> ^ if this code contains a duplicate attestation wouldn't that be considered a bug?

Yes, I thought of that. I think we could use `strict=True`, normally it should be true all the time.

PR #158, opened 2025-11-11
optimize: reduce time complexity from O(t^2) to O(t)

anshalshukla:
## 🗒️ Description
Efficiently calculates `internal_linear_layer` for poseidon2, reducing time complexity from O(t^2) to O(t). Helps considerably reduce key generation and verification times.
- `pytest` runs within 9s down from 19s earlier
- forkchoice tests within 1min down from 3.5mins earlier
- key generation time reduces to 2s down from 11s earlier

## ✅ Checklist
    ```console
    uvx tox
    ```


g11tech (commented):
looks cool opt, I don't think 12 sec makes any diff in CI after this, the keys generated are always same right for deterministic fixtures

PR #160, opened 2025-11-13
fix: align process_attestations() with intended specs

unnawut:
## 🗒️ Description

The finalization logic in the python spec was unintentionally modified, causing implementation that aligned with the latest spec to fail to finalize.

The latest agreed finalization spec is the [markdown spec](https://github.com/leanEthereum/leanSpec/blob/4b750f2748a3718fe3e1e9cdb3c65e3a7ddabff5/docs/client/chain.md#operations), so this PR aligns back to the logic written in the markdown.

To check the correctness of this PR, the proposed [changes to `State.process_attestations()`](https://github.com/leanEthereum/leanSpec/pull/160/files#diff-c33ac3d9a3d230286b9b58b4608c6452a177cc99780ab25d661aef495114b147) must logically match [markdown spec](https://github.com/leanEthereum/leanSpec/blob/4b750f2748a3718fe3e1e9cdb3c65e3a7ddabff5/docs/client/chain.md#operations) fully.

## 🔗 Related Issues or PRs

- https://github.com/ReamLabs/ream/issues/902

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/tests/consensus/devnet/fc/test_fork_choice_reorgs.py
+++ b/tests/consensus/devnet/fc/test_fork_choice_reorgs.py
@@ -778,7 +778,7 @@ def test_reorg_on_newly_justified_slot(
     Scenario
     --------
     Two forks compete. Fork A is heavier and longer, but Fork B manages to
-    become Justified. Fork choice must switch to the justified fork regardless
+    become justified. Fork choice must switch to the justified fork regardless
     of weight/length.
 
     - Slot 1: Base


tests/consensus/devnet/fc/test_fork_choice_reorgs.py
```diff
@@ -816,63 +820,99 @@ def test_reorg_on_newly_justified_slot(
...
+                        SignedAttestationSpec(
+                            validator_id=ValidatorIndex(1),
+                            slot=Slot(5),
+                            target_slot=Slot(5),
+                            target_root_label="fork_b_1",
+                        ),
+                        # fork_b_1 should be able to justify without extra attestations
+                        # from validator 5 and 6 but the test is failing without these
+                        # two attestations below because block proposer's attestations
+                        # are not being counted towards justification
```
unnawut, line 887: Tracking as https://github.com/leanEthereum/leanSpec/issues/174

packages/testing/src/consensus_testing/test_types/store_checks.py
```diff
@@ -140,9 +137,21 @@ class StoreChecks(CamelModel):
...
     latest_finalized_root: Bytes32 | None = None
     """Expected latest finalized checkpoint root."""
 
+    latest_finalized_root_label: str | None = None
+    """
+    Expected latest finalized checkpoint root by label reference.
+
+    Alternative to latest_justified_root that uses the block label system.
+    The framework will resolve this label to the actual block root
+    and validate the latest finalized checkpointroot matches.
```
tcoratger:
```suggestion
    Alternative to `latest_justified_root` that uses the block label system.
    The framework will resolve this label to the actual block root
    and validate the latest finalized checkpoint root matches.
```

tests/consensus/devnet/fc/test_fork_choice_reorgs.py
tcoratger, line 1: I think we should modify the docstring of `test_reorg_on_newly_justified_slot` to reflect the reality of the test, what do you think about something like this:

```python
"""
    Reorg occurs correctly when forks cross justification boundaries.

    Scenario
    --------
    Two forks compete. Fork A is heavier and longer, but Fork B manages to
    become Justified. Fork choice must switch to the justified fork regardless
    of weight/length.

    - Slot 1: Base
    - Slots 2-4: Fork A extends (becomes head with depth 3)
    - Slot 5: Fork B appears (descending from Base, skipping slots 2-4)
    - Slot 6: Fork B extends. This block contains enough attestations to 
              justify Fork B at Slot 5.

    Expected Behavior
    -----------------
    1. Fork A takes the lead initially (Slots 2-4) as the heaviest chain.
    2. Fork B appears at Slot 5 but is initially lighter.
    3. At Slot 6, the new block includes attestations that justify Fork B at Slot 5.
    4. The Justified Checkpoint updates to Slot 5 (fork_b_1).
    5. Fork A is immediately discarded because it does not descend from the new 
       justified checkpoint (Fork A is on a branch from Slot 1).
    6. Fork B becomes the canonical head.
 ```

unnawut: I overlooked this, and our suggestion looks great. Just one note that I lowercased "Justified" and "Justified Checkpoint" for consistency.

Issue #120, opened 2025-11-04
tests: add `test_block_with_wrong_slot` test vector


tcoratger:
After https://github.com/leanEthereum/leanSpec/pull/118 is merged, we should add a test like this (I think for now we don't have enough testing tools so that we need to build new tools as well in the meantime, not exactly sure):

```python
def test_block_with_wrong_slot(
    state_transition_test: StateTransitionTestFiller,
) -> None:
    """
    Test that blocks with mismatched slot are rejected.

    Scenario
    --------
    Attempt to process a block at slot 1, but the block claims to be
    at slot 2.

    Expected Behavior
    -----------------
    Block processing fails with AssertionError: "Block slot mismatch"

    Why This Matters
    ----------------
    Ensures temporal consistency:
    - Blocks can't lie about their slot
    - Prevents time manipulation attacks
    - Maintains protocol timing integrity
    - Essential for slot-based consensus
    """
```

tcoratger:
> Please can I take this [@tcoratger](https://github.com/tcoratger)

Sure assigned

PR #161, opened 2025-11-13
Add wrong-slot STF test with filler support

bomanaps:
## 🗒️ Description
Add skip slot flag to state transition filler and implement consensus test rejecting blocks with mismatched slots.

## 🔗 Related Issues or PRs
https://github.com/leanEthereum/leanSpec/issues/120
## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/packages/testing/src/consensus_testing/test_fixtures/state_transition.py
+++ b/packages/testing/src/consensus_testing/test_fixtures/state_transition.py
@@ -141,7 +141,6 @@ class StateTransitionTest(BaseConsensusFixture):
                 # Store the filled Block for serialization
                 filled_blocks.append(block)
 
-                # Process block through state transition
                 # Use cached state if available, otherwise run state transition
                 if cached_state is not None:
                     state = cached_state
@@ -194,15 +193,8 @@ class StateTransitionTest(BaseConsensusFixture):
         """
         Build a Block from a BlockSpec, optionally caching the post-state.
 
-        Uses provided fields from spec, computes any missing fields.
-        This mimics what a local block builder would do.
-
         Returns both the block and the cached post-state (if computed) to avoid
         redundant state transitions.
-
-        TODO: If the spec implements a State.produce_block() method in the
-        future, we should use that instead of manually computing fields here.
-        Until then, this manual approach is necessary.
 
         Parameters
         ----------
@@ -219,7 +211,6 @@ class StateTransitionTest(BaseConsensusFixture):
         # Use provided proposer index or compute it
         proposer_index = spec.proposer_index or Uint64(int(spec.slot) % len(state.validators))
 
-        # Optionally advance a temporary state to the block's slot
         temp_state: State | None = None
         if not spec.skip_slot_processing:
             temp_state = state.process_slots(spec.slot)
@@ -247,7 +238,6 @@ class StateTransitionTest(BaseConsensusFixture):
             )
             return block, None
 
-        # Create temporary block with zero state root (will be computed)
         temp_block = Block(
             slot=spec.slot,
             proposer_index=proposer_index,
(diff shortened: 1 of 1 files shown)


packages/testing/src/consensus_testing/test_fixtures/state_transition.py
```diff
@@ -60,8 +60,8 @@ class StateTransitionTest(BaseConsensusFixture):
     """
     The filled Blocks, processed through the specs.
 
-    This is a private attribute not part of the model schema. Tests cannot set this.
-    The framework populates it during make_fixture().
+    This is a private attribute not part of the model schema.
+    Tests cannot set this. The framework populates it during make_fixture().
```
tcoratger: I think there is no change here, can we revert this to minimize git diff? I guess this is the same at other places in this PR :)

packages/testing/src/consensus_testing/test_fixtures/state_transition.py
```diff
@@ -110,7 +113,8 @@ def make_fixture(self) -> "StateTransitionTest":
         """
         Generate the fixture by running the spec.
 
-        Builds blocks from BlockSpec if needed, then processes them through state_transition.
+        Builds blocks from BlockSpec if needed, then processes them through
+        state_transition.
```
tcoratger: Same comment about minimizing git diff

packages/testing/src/consensus_testing/test_fixtures/state_transition.py
```diff
@@ -181,9 +194,9 @@ def _build_block_from_spec(self, spec: BlockSpec, state: State) -> Block:
         Uses provided fields from spec, computes any missing fields.
         This mimics what a local block builder would do.
 
-        TODO: If the spec implements a State.produce_block() method in the future,
-        we should use that instead of manually computing fields here. Until then,
-        this manual approach is necessary.
+        TODO: If the spec implements a State.produce_block() method in the
+        future, we should use that instead of manually computing fields here.
+        Until then, this manual approach is necessary.
```
tcoratger: Same message about minimizing git diff

packages/testing/src/consensus_testing/test_fixtures/state_transition.py
```diff
@@ -138,10 +142,13 @@ def make_fixture(self) -> "StateTransitionTest":
...
-                    block=block,
-                    valid_signatures=True,
-                )
+                if getattr(block_spec, "skip_slot_processing", False):
+                    state = state.process_block(block)
+                else:
+                    state = state.state_transition(
+                        block=block,
+                        valid_signatures=True,
+                    )
```
tcoratger: I'm not sure why we need all this here and why we need this new `skip_slot_processing`, looks like confusing and I guess we want to test the classical `state_transition` processing here because the block processing is the first step of it

https://github.com/leanEthereum/leanSpec/blob/715276e11ea7e1bfb61060aeff37af2c9740decf/src/lean_spec/subspecs/containers/state/state.py#L521-L522

so that we should see the exact same error that if we do `state.process_block(block)` because the only additional step which happens before this in the state transition processing is to check that `valid_signatures` is true but this is the case here in this call so there should be no problem.

bomanaps: The only reason the `skip_slot_processing` flag exists is to make the new “wrong slot” vector possible. `state_transition()` always advances through `process_slots(block.slot)` before running `process_block`, so by the time the header check runs the state slot already equals the block slot and `assert block.slot == self.slot` can never trip. The flag simply lets an individual test opt out of that automatic advancement so we can feed the block to `process_block` while the state is still at the earlier slot. Everything else continues through the normal `state_transition path`; no existing tests are touched. If you’d prefer an alternative hook (e.g., a dedicated helper to run process_block directly), I’m happy to rework it the key requirement is just some way to bypass that initial `process_slots` call for this negative test.

packages/testing/src/consensus_testing/test_fixtures/state_transition.py
bomanaps: Sorry about this I got a Ruff’s line-length lint so had to split to a new line

packages/testing/src/consensus_testing/test_fixtures/state_transition.py
tcoratger: Ah yes true I get it, can you solve conflicts?

tcoratger:
Hey @bomanaps I think that we have some conflicts with recent updates and then looks good, we should just try to minimize git diff here especially with the comments if possible.

PR #162, opened 2025-11-13
forkchoice helpers: rm `get_latest_justified`

tcoratger:
## 🗒️ Description

There is a single `get_latest_justified` call for now that can be replaced by an even simpler one line thing so that this makes no sense to have a `get_latest_justified` helper for this. Removing this simplify the reading experience.

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/src/lean_spec/subspecs/forkchoice/store.py
+++ b/src/lean_spec/subspecs/forkchoice/store.py
@@ -398,22 +398,27 @@ class Store(Container):
             New Store with updated head, latest_justified, and latest_finalized.
 
         """
-        # Compute latest justified checkpoint from all known states
+        # Find the Latest Justified Checkpoint
         #
-        # Scans every state in the store to find the checkpoint with the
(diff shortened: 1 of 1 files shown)


unnawut:
This was also to mimic beacon chain structure so I'll defer to @g11tech

g11tech:
we do this on on_block, with justified being the new post state's justified on onblock

https://github.com/blockblaz/zeam/blob/main/pkgs/node/src/forkchoice.zig#L187

tcoratger:
@g11tech I just merge this one as is because this PR was only dedicated to removing `get_latest_justified` and so to avoid overloading the PR, I just merge.

About your comment, `store.update_head()` with this logic is also used in `accept_new_attestations` method so I'm not sure about your proposal. Can you open a dedicated issue to illustrate what you propose in details so that we can tackle it in a follow up PR?

PR #170, opened 2025-11-14
Add devnet forkchoice lexicographic tie-break test vector (with StoreChecks safeguards)

### LINKED ISSUE #113: tests: add tests that lexicographically higher block hash is chosen when fork weight is equal

vyakart:
Closes #113

## Summary

This PR adds a Devnet forkchoice **test vector** that validates the deterministic
lexicographic tie-break: when two sibling forks have equal weight and share the
same slot, the forkchoice algorithm must select the block whose **SSZ root is
lexicographically highest**.

The implementation now follows the shared StoreChecks pattern used across the
spec.

---

## ✔ New Test Vector

The dedicated scenario lives in:

tests/consensus/devnet/fc/test_lexicographic_tiebreaker.py

It uses:

lexicographic_head_among=["fork_a_3", "fork_b_3"]

StoreChecks validates:

- each label exists  
- all referenced blocks share the same slot  
- the head matches the lexicographically highest SSZ root  

This ensures the tie-break is tested correctly and cannot pass with incomplete
or misleading inputs.

---

## ✔ Reorgs Test Restored

The file:
tests/consensus/devnet/fc/test_fork_choice_reorgs.py

has been restored to its original content.  
All previously added logic there has been removed to avoid duplicate or misplaced
coverage.

---

## ✔ StoreChecks Enhancements

`lexicographic_head_among` in StoreChecks now provides hardened
guardrails:

- resolves each label to its block root  
- enforces equal-slot requirement  
- verifies head == lexicographic max(root_i)  

These protections fix the issue where earlier tests could pass even with a
single label or mismatched forks.

---

## Testing

All checks pass:

uvx tox -e pytest
uvx tox -e all-checks
uv run fill --fork=Devnet --clean -n auto

This confirms both the Devnet vector and the StoreChecks logic are correct.

---

## Notes

Thanks to @tcoratger for the guidance and the clean rework of the tie-break
store-check logic — this PR now fully aligns with the migration toward
consumable, deterministic test vectors.

### DIFF
--- a/packages/testing/src/consensus_testing/test_types/store_checks.py
+++ b/packages/testing/src/consensus_testing/test_types/store_checks.py
@@ -385,12 +385,19 @@ class StoreChecks(CamelModel):
                         f"but block_registry not provided to validate_against_store()"
                     )
 
+                # Require at least 2 forks to test tiebreaker
+                if len(expected_value) < 2:
+                    raise ValueError(
+                        f"Step {step_index}: lexicographic_head_among requires at least 2 forks "
+                        f"to test tiebreaker behavior, got {len(expected_value)}: {expected_value}"
+                    )
+
                 # Import hash_tree_root locally to avoid circular import
                 from lean_spec.subspecs.ssz import hash_tree_root
 
-                # Resolve all fork labels to roots
-                fork_roots: dict[str, Bytes32] = {}
-                fork_slots: dict[str, Slot] = {}
+                # Resolve all fork labels to roots and compute their weights
+                # Map: label -> (root, slot, weight)
+                fork_data: dict[str, tuple[Bytes32, Slot, int]] = {}
                 for label in expected_value:
                     if label not in block_registry:
                         available = list(block_registry.keys())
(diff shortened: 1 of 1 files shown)


vyakart:
Thanks for the pointer!
`test_equal_weight_forks_use_lexicographic_tiebreaker` now lives under `tests/consensus/devnet/fc/test_fork_choice_reorgs.py`, so fixtures/clients will capture the deterministic tie-break behaviour.

I re-ran:
- `uvx tox -e pytest`
- `uvx tox -e all-checks`

tcoratger (commented):
I think that we should remove everything that you added in the `test_helpers.py` file because this will create a pytest but in our case we want only a test vector (as part of our migration process to test vectors aimed at creating something consumable by client teams).

Also in your `test_equal_weight_forks_use_lexicographic_tiebreaker`, I'm not sure about the behavior, like ideally we should confirm that at the end, the head points to proper thing according to the rule of the lexicographic ordering of hashes so that in the test we want detailed documentation like in the other tests and we also want to be sure that the one we pick at the end really has the highest hash in lexicographic order.

Your idea in the other place to have this

```python
root_hash = _byte32(0) 
low_child_hash = _byte32(1) 
high_child_hash = _byte32(2)
```

was a good method because we could hardcoded this into the blocks so that we can check that the picked head really is the highest hash. But in our case with test vectors it's different because we rely on a real world scenario so that we should find a clean way to check this probably via something inside the test framework or something nice to think about.

tests/consensus/devnet/fc/test_fork_choice_reorgs.py
tcoratger, line 1: You can delete everything added here basically

vyakart:
Thanks for the guidance! I've applied the full set of changes:
Merged your PR into my branch.
Restored test_fork_choice_reorgs.py to its original content.

Added the new lexicographic tie-break scenario under
tests/consensus/devnet/fc/test_lexicographic_tiebreaker.py using the new
lexicographic_head_among StoreChecks field.

Updated StoreChecks with your implementation and synced the test accordingly.

All checks are green (uvx tox -e pytest, uvx tox -e all-checks, and the Devnet filler).

tests/consensus/devnet/fc/test_lexicographic_tiebreaker.py
```diff
@@ -0,0 +1,91 @@
...
+                ),
+            ),
+            BlockStep(
+                block=BlockSpec(slot=Slot(3), parent_label="fork_b_2", label="fork_b_3"),
+                checks=StoreChecks(
+                    head_slot=Slot(3),
+                    # Both forks now have equal weight (2 blocks each)
+                    #
+                    # Tiebreaker determines the winner
+                    lexicographic_head_among=["fork_a_3", "fork_b_3"],
```
tcoratger, line 87: I've just been thinking about it again, and there's a problem here because the test still passed whether we used:

```python
lexicographic_head_among=["fork_a_3", "fork_b_3"]
```

or

```python
lexicographic_head_among=["fork_b_3"]
```

or

```python
lexicographic_head_among=["fork_a_3"]
```

So clearly there was a problem because we weren't testing the right thing. I think I've solved the problem here: https://github.com/vyakart/leanSpec/pull/2. Please double-check it but I've added other safeguards, such as ensuring the weights are equal or that we have two or more fork labels that are also in the same slot.

vyakart:
Thanks for the careful review and the follow-up fix — I pulled in your PR,
validated the updated StoreChecks safeguards, and confirmed that the test now
behaves correctly under all the edge cases you mentioned.

The same-slot requirement, explicit label resolution, and strict lexicographic
comparison fully address the issue where the test previously passed regardless
of the input.

Everything now works end-to-end, including fixture generation.  
Really appreciate the clear guidance — this made the final vector much cleaner
and fully aligned with the spec’s testing pattern.

tcoratger (approved):
Looks good to me, @unnawut can you confirm?

packages/testing/src/consensus_testing/test_types/store_checks.py
```diff
@@ -158,6 +158,20 @@ class StoreChecks(CamelModel):
     attestation_checks: list[AttestationCheck] | None = None
     """Optional list of attestation content checks for specific validators."""
 
+    lexicographic_head_among: list[str] | None = None
```
unnawut, line 161: I'm not sure if we'll ever need to explicitly use this check again outside of `test_lexicographic_tiebreaker.py` unlike head_slot, latest_finalized_root_label, etc. etc.

tests/consensus/devnet/fc/test_lexicographic_tiebreaker.py
```diff
@@ -0,0 +1,91 @@
...
+
+    Expected Behavior
+    -----------------
+    The competing forks have identical attestation weight. The head is chosen
+    via lexicographic ordering of the block roots. The framework automatically
+    verifies that:
+    1. Both forks are at the same slot (equal depth)
+    2. The head is the lexicographically highest root among them
+    """
+    fork_choice_test(
```
unnawut, line 46: Because we're not specifying the block root, we are still relying on multiple runs to be confident that no matter how block roots are ordered before entering the fork choice the result is correct.

vyakart:
Thanks for the review and the notes!
Happy to revisit the StoreChecks field or root handling in a follow-up if we end up reusing this pattern elsewhere.

Issue #167, opened 2025-11-14
testing framework: use spec instead of `_build_block_from_spec` for forkchoice


tcoratger:
Right now we have

https://github.com/leanEthereum/leanSpec/blob/0354f5da6db7e0dad610bfbbe1be4a37a465fcc7/packages/testing/src/consensus_testing/test_fixtures/fork_choice.py#L285-L421

As highlighted in the TODO, we can't use `Store.produce_block_with_signatures` directly here to do side effects. However similarly to https://github.com/leanEthereum/leanSpec/issues/166, this is much better to call the spec here instead of doing manual manipulation. In https://github.com/leanEthereum/leanSpec/issues/128, there is also talk of a refactoring of the `Store` which includes `produce_block_with_signatures` so that we should probably do this first or group the two together but in any case it should be nice to remove this manual block processing here in the test framework and just rely on the spec for this.

vyakart:
I’d like to volunteer for taking up this issue (#167). 

From what I understand, the goal here is to replace the manual block-building logic in the forkchoice fixtures with a more canonical spec-level API (likely derived from Store.produce_block_with_signatures) so that tests no longer assemble blocks manually.

A few questions before I begin:
	1.	Do you view this issue as depending on the forkchoice store refactor in issue #128 (i.e., splitting block-build vs store-insert) or is an incremental change acceptable for now?
	2.	Is there an existing preferred API shape for block production/attestation assembly (for example State.produce_block(), or BlockBuilder + signatures) that you’d like me to adhere to, or should I propose a draft?

I’m happy to start with a short design doc or draft PR to refine the approach.

PR #171, opened 2025-11-14
testing framework: replace _build_block_from_spec by a call to a spec method

### LINKED ISSUE #166: testing framework: replace state `_build_block_from_spec` by a call to a spec method
Currently here 

https://github.com/leanEthereum/leanSpec/blob/0354f5da6db7e0dad610bfbbe1be4a37a465fcc7/packages/testing/src/consensus_testing/test_fixtures/state_transition.py#L177-L249

we have a manual block builder inside the test framework. As mentioned in the TODO here

https://github.com/leanEthereum/leanSpec/blob/0354f5da6db7e0dad610bfbbe1be4a37a465fcc7/packages/testing/src/consensus_testing/test_fixtures/state_transition.py#L184-L186

we should probably have the equivalent method directly available in the spec and call it from there instead of manually building the block. Indeed, the ideal situation for testing things properly is for this type of building to be done via the spec and not via manual manipulations that do not call the spec.

varun-doshi:
## 🗒️ Description

- introduces a `produce_block` function in the spec
- this takes in an optional boolean parameter which signifies whether to return an "unprocessed" block or not
- "unprocessed" here means not running `process_block` on the new block. this is needed for cases where we are testing invalid proposer/slot/state root etc.
- the block being returned from here is now updated according to the `BlockSpec` in our testing framework.

So essentially, the actual "build a block" logic stays in the spec under State, while testing specific code where we alter fields in a block remain under the testing framework.

## 🔗 Related Issues or PRs

Fixes #166 

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/packages/testing/src/consensus_testing/test_fixtures/state_transition.py
+++ b/packages/testing/src/consensus_testing/test_fixtures/state_transition.py
@@ -173,7 +173,8 @@ class StateTransitionTest(BaseConsensusFixture):
 
     def fill_block_from_spec(self, spec: BlockSpec, state: State) -> Block:
         """
-        Fills the created block from `State.produce_block()` based on fields in the spec for state transition tests.
+        Fills the created block from `State.produce_block()`
+        based on fields in the spec for state transition tests.
 
         Parameters
         ----------


packages/testing/src/consensus_testing/test_fixtures/state_transition.py
```diff
@@ -197,53 +188,25 @@ def _build_block_from_spec(self, spec: BlockSpec, state: State) -> Block:
...
+            temp_block = state.produce_block(spec.slot, True)
+        else:
+            temp_block = state.produce_block(spec.slot, False)
+
         # Use provided proposer_index or compute it
         if spec.proposer_index is not None:
-            proposer_index = spec.proposer_index
-        else:
-            proposer_index = ValidatorIndex(int(spec.slot) % int(state.validators.count))
+            temp_block.model_copy(update={"proposer_index": spec.proposer_index})
```
unnawut, line 198: I think model_copy is making a copy of the block, and its return is not captured so the 4 new model_copy are basically a no-op. 

Also temp_block remains unmodified but I have no idea why tests are all passing.

packages/testing/src/consensus_testing/test_fixtures/state_transition.py
```diff
@@ -197,53 +188,25 @@ def _build_block_from_spec(self, spec: BlockSpec, state: State) -> Block:
         Block
             A complete block ready for state_transition.
         """
+        if self.expect_exception:
+            temp_block = state.produce_block(spec.slot, True)
+        else:
+            temp_block = state.produce_block(spec.slot, False)
```
unnawut, line 194: I don't think we can generalize that all exceptions produce empty blocks and all non-exceptions produce non-empty blocks.

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -440,3 +440,54 @@ def state_transition(self, block: Block, valid_signatures: bool = True) -> "Stat
...
+            state_root=Bytes32.zero(),
+            body=BlockBody(attestations=Attestations(data=[])),
+        )
+
+        if is_empty:
+            return block
+
+        # 3. Process the state with the block to get the final state_root
+        post_state = pre_state.process_block(block)
+        state_root = hash_tree_root(post_state)
```
unnawut, line 484: I'm also not sure we can treat that when the block body is empty, we don't need to process_block nor the state_root

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -440,3 +440,54 @@ def state_transition(self, block: Block, valid_signatures: bool = True) -> "Stat
             raise AssertionError("Invalid block state root")
 
         return new_state
+
+    def produce_block(self, target_slot: Slot, is_empty: bool = False) -> Block:
```
unnawut, line 444: Since this is added to the spec, when would we be using this function instead of [`produce_block_with_signatures()`](https://github.com/leanEthereum/leanSpec/blob/3e812e0b77eebcae446ee59a8384f9fdf5b60fd6/src/lean_spec/subspecs/forkchoice/store.py#L691)? It's still a test helper function to me.

packages/testing/src/consensus_testing/test_fixtures/state_transition.py
```diff
@@ -197,53 +188,25 @@ def _build_block_from_spec(self, spec: BlockSpec, state: State) -> Block:
...
-        else:
-            body = BlockBody(attestations=Attestations(data=[]))
+            temp_block.model_copy(update={"body": spec.body})
 
         # Use provided state_root or compute it via dry-run
         if spec.state_root is not None:
-            state_root = spec.state_root
-        else:
-            # Need to dry-run to compute state_root
-            temp_state = state.process_slots(spec.slot)
```
unnawut, line 224: Because this function is called by `BlockStep` in the tests, which can skip blocks for some slots, I think we still need to retain `process_slots()` to follow the STF flow

tcoratger (commented):
I'd like to know something before doing anything with this. I opened the corresponding issue because it's generally not something we want to have methods inside test vector fixtures with dedicated logic that could be handled by the specification itself.

However, on the other hand, the associated function, if it's ultimately migrated to the specification, must be useful to the specification itself because it's also not good practice to include unnecessary functions in the core of the specification; that's not the goal at all.

In this specific case, I feel like we might need a general method to produce a block in the state, but I'm not sure about that. Can you confirm or not? If it's not useful and never will be, then we can just close the pull request and close https://github.com/leanEthereum/leanSpec/issues/166 as well as https://github.com/leanEthereum/leanSpec/issues/167.

PR #173, opened 2025-11-17
framework: add configurable seeds/activation to XmssKeyManager (#129)

### LINKED ISSUE #129: Upgrade XmssKeyManager utility
Currently the `XmssKeyManager` isn't completely configurable. Let's make it more generic, few pointer to include (non exhaustive list):
- `create_and_store_key_pair` offer very little control to user. Add activation_epoch, active_epochs, seed, etc to give complete control over key generation
- save the private keys along with test vectors to help client teams verify results
- we are padding the signature_bytes before writing them to Signature object, this look kinda ugly, see if there is a way this can be better handled (`sign_attestation`)
- anything else to make it generic and give users complete control over key generation

vyakart:
## Description

This pull request makes XMSS key handling in the testing framework more configurable and deterministic. It extends `XmssKeyManager` with explicit control over activation epochs, active lifetimes, and per-key seeds, so tests can request reproducible key material tailored to a given scenario while keeping the existing defaults for callers that do not care.

A new `create_and_store_key_pair` entry point is introduced to generate and cache keys using a deterministic tuple `(validator_index, activation_epoch, num_active_epochs, seed)`. The legacy behaviour of reusing a single key per validator backed by a shared manager is preserved when no additional configuration is provided, so current fixtures remain unchanged.

To reduce duplication around XMSS signatures, the PR adds `Signature.from_xmss`, which centralises the conversion from XMSS containers to the consensus `Signature` type (including padding and length validation). Fork-choice fixtures are wired to these capabilities via new optional fields (`key_manager_seed` and `key_manager_activation_epoch`), allowing tests to opt into a dedicated, keyed `XmssKeyManager` when needed while defaulting to the shared manager path otherwise.

## Related Issues or PRs

Should close #129

## Checklist

  ```bash
  uvx tox -e all-checks
  uvx tox -e pytest

### DIFF
--- a/packages/testing/src/consensus_testing/keys.py
+++ b/packages/testing/src/consensus_testing/keys.py
@@ -195,7 +195,7 @@ class XmssKeyManager:
         Returns:
         -------
         KeyPair
-            The validator's XMSS key pair.
+            XMSS key pair associated with the validator.
 
         Notes:
         -----
@@ -273,11 +273,14 @@ class XmssKeyManager:
         # Convert to the consensus Signature container (handles padding internally).
         return Signature.from_xmss(xmss_sig, self.scheme)
 
-        # Ensure the signature meets the consensus spec length (3116 bytes).
-        #
-        # This is necessary when using TEST_CONFIG (796 bytes) vs PROD_CONFIG.
-        # Padding with zeros on the right maintains compatibility.
-        padded_bytes = signature_bytes.ljust(Signature.LENGTH, b"\x00")
+    def export_test_vectors(self, include_private_keys: bool = False) -> list[dict[str, Any]]:
+        """
+        Export generated keys as dictionaries suitable for JSON test vectors.
+
+        Parameters
+        ----------
+        include_private_keys : bool, optional
+            When True, include SecretKey contents for debugging fixtures.
 
         Returns:
         -------
@@ -313,7 +316,7 @@ class XmssKeyManager:
         Returns:
         -------
         PublicKey
-            The validator's public key.
+            Public key for the validator.
         """
         return self[validator_index].public
 


packages/testing/src/consensus_testing/keys.py
```diff
@@ -23,24 +30,63 @@ class KeyPair(NamedTuple):
...
 """
 
 
+def _to_int(value: int | Slot | Uint64 | None, default: int = 0) -> int:
+    """Normalize Slot/Uint64/int to int with an optional default."""
+    if value is None:
+        return default
+    if isinstance(value, Slot):
+        return value.as_int()
+    return int(value)
```
tcoratger: We should not have this, this is not clean to have. We should probably rely only on Uint64 or int and not have this. This is a primitive conversion stuff that should not be inside the testing framework folder.

packages/testing/src/consensus_testing/keys.py
```diff
@@ -23,24 +30,63 @@ class KeyPair(NamedTuple):
...
+        # Use a deterministic stream rather than os.urandom for repeatability in tests.
+        return self._rng.randbytes(PRF_KEY_LENGTH)
+
+
 class XmssKeyManager:
     """Lazy key manager for test validators using XMSS signatures."""
 
     DEFAULT_MAX_SLOT = Slot(100)
     """Default maximum slot horizon if not specified."""
+    DEFAULT_ACTIVATION_EPOCH = 0
```
tcoratger: We need a doc string to explain

packages/testing/src/consensus_testing/keys.py
```diff
@@ -23,24 +30,63 @@ class KeyPair(NamedTuple):
...
 
     def __init__(
         self,
         max_slot: Optional[Slot] = None,
         scheme: GeneralizedXmssScheme = TEST_SIGNATURE_SCHEME,
+        default_activation_epoch: int | Slot | Uint64 = DEFAULT_ACTIVATION_EPOCH,
```
tcoratger: We should probably have only Uint64 here to avoid additional complexities

packages/testing/src/consensus_testing/keys.py
```diff
@@ -61,7 +112,92 @@ def __init__(
...
+            self._schemes_by_seed[seed] = GeneralizedXmssScheme(
+                config=self.scheme.config,
+                prf=SeededPrf(self.scheme.config, seed),
+                hasher=self.scheme.hasher,
+                merkle_tree=self.scheme.merkle_tree,
+                encoder=self.scheme.encoder,
+                rand=SeededRand(self.scheme.config, seed),
+            )
+
+        return self._schemes_by_seed[seed]
```
tcoratger: I don't think that we need to use this

packages/testing/src/consensus_testing/keys.py
```diff
@@ -61,7 +112,92 @@ def __init__(
...
+            key_pair = KeyPair(public=pk, secret=sk)
+            _KEY_CACHE[cache_key] = key_pair
+
+        self._key_pairs[validator_index] = key_pair
+        self._key_metadata[validator_index] = {
+            "activation_epoch": activation_epoch_int,
+            "num_active_epochs": num_active_epochs_int,
+            "seed": key_seed,
+        }
+        return key_pair
```
tcoratger, line 184: Same not sure that is useful, `__getitem__` was much more idiomatic in my opinion

packages/testing/src/consensus_testing/keys.py
```diff
@@ -23,24 +30,63 @@ class KeyPair(NamedTuple):
...
+
+    def __init__(self, config: XmssConfig, seed: int) -> None:
+        """Initialize with a deterministic seed."""
+        super().__init__(config)
+        self._rng = random.Random(seed)
+
+    def key_gen(self) -> bytes:
+        """Generate a deterministic PRF key for repeatable tests."""
+        # Use a deterministic stream rather than os.urandom for repeatability in tests.
+        return self._rng.randbytes(PRF_KEY_LENGTH)
```
tcoratger: This should not be allowed because we want the PRF and RAND to be based on the XMSS paper, so we can't just put whatever we want there. Here, I've created a pull request to prevent this behavior: https://github.com/leanEthereum/leanSpec/pull/175

Now this should error out

packages/testing/src/consensus_testing/keys.py
```diff
@@ -36,11 +38,13 @@ class XmssKeyManager:
 
     DEFAULT_MAX_SLOT = Slot(100)
     """Default maximum slot horizon if not specified."""
+    DEFAULT_ACTIVATION_EPOCH = Uint64(0)
```
tcoratger, line 41:
```suggestion
    DEFAULT_ACTIVATION_EPOCH = Uint64(0)
    """Default activation epoch if not specified"""
```

packages/testing/src/consensus_testing/keys.py
```diff
@@ -225,3 +256,29 @@ def __contains__(self, validator_index: ValidatorIndex) -> bool:
...
+                "num_active_epochs": meta.get("num_active_epochs"),
+                "public_key": key_pair.public.to_bytes(self.scheme.config).hex(),
+            }
+            if include_private_keys:
+                # Pydantic models are JSON-serializable; keep the raw dump for full fidelity.
+                entry["secret_key"] = key_pair.secret.model_dump(mode="json")
+
+            vectors.append(entry)
+
+        return vectors
```
tcoratger: Is it used somewhere?

tests/lean_spec/subspecs/containers/test_signature.py
```diff
@@ -0,0 +1,123 @@
...
+        ]
+        xmss_sig = XmssSignature(path=path, rho=rho, hashes=hashes)
+
+        # Convert to consensus signature
+        consensus_sig = Signature.from_xmss(xmss_sig, TEST_SIGNATURE_SCHEME)
+
+        # Verify it's the correct type and length
+        assert isinstance(consensus_sig, Signature)
+        assert len(consensus_sig) == Signature.LENGTH
+        assert len(consensus_sig) == 3100
```
tcoratger: I don't recommend to put hardcoded numbers like this in tests, it is then a nightmare to debug if we change the config for example. Asserting this `assert len(consensus_sig) == Signature.LENGTH` is enough in my opinion.

```suggestion
```

tests/lean_spec/subspecs/containers/test_signature.py
```diff
@@ -0,0 +1,123 @@
...
+        )
+        rho = [Fp(value=0)] * TEST_CONFIG.RAND_LEN_FE
+        hashes = [[Fp(value=0)] * TEST_CONFIG.HASH_LEN_FE] * TEST_CONFIG.DIMENSION
+        xmss_sig = XmssSignature(path=path, rho=rho, hashes=hashes)
+
+        consensus_sig = Signature.from_xmss(xmss_sig, TEST_SIGNATURE_SCHEME)
+
+        # The signature should be padded with zeros
+        raw_xmss = xmss_sig.to_bytes(TEST_CONFIG)
+        expected_padding = 3100 - len(raw_xmss)
```
tcoratger: The 3100 should come from the production config no? we should not have hardcoded numbers but instead computations

src/lean_spec/subspecs/containers/signature.py
```diff
@@ -28,3 +28,22 @@ def verify(
...
+
+    @classmethod
+    def from_xmss(
+        cls, xmss_signature: XmssSignature, scheme: GeneralizedXmssScheme = TEST_SIGNATURE_SCHEME
+    ) -> "Signature":
+        """
+        Create a consensus `Signature` container from an XMSS signature object.
+
+        Handles padding to the fixed 3100-byte length required by the consensus layer,
+        delegating all encoding details to the XMSS container itself.
```
tcoratger: That is confusing to put the raw 3100 number here directly, we don't really understand from where it comes from

```suggestion
```

tests/lean_spec/subspecs/containers/test_signature.py
```diff
@@ -0,0 +1,123 @@
...
+        consensus_sig = Signature.from_xmss(xmss_sig, TEST_SIGNATURE_SCHEME)
+
+        # The beginning of the consensus signature should match the XMSS bytes
+        raw_xmss = xmss_sig.to_bytes(TEST_CONFIG)
+        assert bytes(consensus_sig)[: len(raw_xmss)] == raw_xmss
+
+    def test_from_xmss_roundtrip_with_verify(self) -> None:
+        """Test that a signature created via from_xmss can be verified."""
+        from lean_spec.subspecs.xmss.interface import TEST_SIGNATURE_SCHEME
+        from lean_spec.types import Uint64
```
tcoratger: This should be already imported on the top of the file no?

```suggestion
```

tests/lean_spec/subspecs/containers/test_signature.py
```diff
@@ -0,0 +1,123 @@
...
+        hashes = [[Fp(value=0)] * TEST_CONFIG.HASH_LEN_FE] * TEST_CONFIG.DIMENSION
+
+        xmss_sig1 = XmssSignature(path=path1, rho=rho, hashes=hashes)
+        xmss_sig2 = XmssSignature(path=path2, rho=rho, hashes=hashes)
+
+        consensus_sig1 = Signature.from_xmss(xmss_sig1, TEST_SIGNATURE_SCHEME)
+        consensus_sig2 = Signature.from_xmss(xmss_sig2, TEST_SIGNATURE_SCHEME)
+
+        # Different XMSS signatures should produce different consensus signatures
+        assert bytes(consensus_sig1) != bytes(consensus_sig2)
```
tcoratger: This one is not needed, covered by all the XMSS tests

```suggestion
```

tests/lean_spec/subspecs/containers/test_signature.py
```diff
@@ -0,0 +1,123 @@
...
+        xmss_sig = XmssSignature(path=path, rho=rho, hashes=hashes)
+
+        consensus_sig = Signature.from_xmss(xmss_sig, TEST_SIGNATURE_SCHEME)
+
+        # The signature should be padded with zeros
+        raw_xmss = xmss_sig.to_bytes(TEST_CONFIG)
+        expected_padding = 3100 - len(raw_xmss)
+
+        # Verify the last bytes are zeros (padding)
+        assert consensus_sig[-expected_padding:] == b"\x00" * expected_padding
```
tcoratger: We don't need this test, the padding thing is not clean and we should rather try to remove this in a followup PR

```suggestion
```

vyakart:
Thanks again for the guidance!
I’ve made the requested changes across the commits:
	•	Fixed all style/lint issues and imported the missing docstrings.
	•	Updated the logic around XmssKeyManager and fork-choice fixtures to match the design we discussed.
	•	Ensured that existing defaults remain unchanged while exposing the new seed/activation parameters.
	•	Re-ran the full test suite (uvx tox -e pytest + uvx tox -e all-checks), and everything is passing.

I believe the PR now addresses all the reviewer’s comments. Happy to make further tweaks if you’d prefer any variant in the APIs or if any tests require additional coverage.

packages/testing/src/consensus_testing/keys.py
```diff
@@ -36,11 +38,14 @@ class XmssKeyManager:
...
     DEFAULT_MAX_SLOT = Slot(100)
     """Default maximum slot horizon if not specified."""
+    DEFAULT_ACTIVATION_EPOCH = Uint64(0)
+    """Default activation epoch if not specified"""
 
     def __init__(
         self,
         max_slot: Optional[Slot] = None,
         scheme: GeneralizedXmssScheme = TEST_SIGNATURE_SCHEME,
+        default_activation_epoch: Optional[Uint64] = None,
```
anshalshukla: nit: let's just name it `activation_epoch` here, it's a function param. Also I'll prefer the order to be maintained here so `activation_epoch` followed by `max_slot`

packages/testing/src/consensus_testing/keys.py
```diff
@@ -53,6 +58,8 @@ def __init__(
         scheme : GeneralizedXmssScheme, optional
             The XMSS scheme to use.
             Defaults to `TEST_SIGNATURE_SCHEME`.
+        default_activation_epoch : Uint64, optional
+            Activation epoch used when none is provided for key generation.
```
anshalshukla: rename to `activation_epoch`

packages/testing/src/consensus_testing/keys.py
```diff
@@ -61,7 +68,67 @@ def __init__(
         """
         self.max_slot = max_slot if max_slot is not None else self.DEFAULT_MAX_SLOT
         self.scheme = scheme
+        self.default_activation_epoch = (
+            default_activation_epoch
+            if default_activation_epoch is not None
+            else self.DEFAULT_ACTIVATION_EPOCH
+        )
```
anshalshukla: same as above

packages/testing/src/consensus_testing/keys.py
```diff
@@ -61,7 +68,67 @@ def __init__(
...
         self._key_pairs: dict[ValidatorIndex, KeyPair] = {}
+        self._key_metadata: dict[ValidatorIndex, dict[str, Any]] = {}
+
+    @property
+    def default_num_active_epochs(self) -> int:
+        """Default lifetime derived from the configured max_slot."""
+        return self.max_slot.as_int() + 1
```
anshalshukla, line 115: I think the function naming is misleading here. You can maybe name it `default_max_epoch` and use DEFAULT_MAX_SLOT here.

packages/testing/src/consensus_testing/keys.py
```diff
@@ -61,7 +68,67 @@ def __init__(
...
+            pk, sk = self.scheme.key_gen(activation_epoch_val, num_active_epochs_val)
+            key_pair = KeyPair(public=pk, secret=sk)
+            _KEY_CACHE[cache_key] = key_pair
+
+        self._key_pairs[validator_index] = key_pair
+        self._key_metadata[validator_index] = {
+            "activation_epoch": int(activation_epoch_val),
+            "num_active_epochs": int(num_active_epochs_val),
+        }
+        return key_pair
```
anshalshukla, line 184: This is okay, but ideally I would like to have multiple keys for same validator based on the activation epochs, I think having longer activation epochs affects the signing time. Maybe I'm asking too much here and we won't need it with pregen signatures.

@tcoratger  what are your thoughts on this?

tests/lean_spec/subspecs/containers/test_signature.py
```diff
@@ -0,0 +1,79 @@
...
+        # Convert to consensus format
+        consensus_sig = Signature.from_xmss(xmss_sig, TEST_SIGNATURE_SCHEME)
+
+        # The beginning of the consensus signature should match the XMSS bytes
+        raw_xmss = xmss_sig.to_bytes(TEST_CONFIG)
+        assert bytes(consensus_sig)[: len(raw_xmss)] == raw_xmss
+
+    def test_from_xmss_roundtrip_with_verify(self) -> None:
+        """Test that a signature created via from_xmss can be verified."""
+        from lean_spec.types import Uint64
```
anshalshukla: Isn't this test sufficient to test basic_conversion and preserves_data tests as well. Raw signature construction using path, rho, hashes, looks weird to me. We can have key generation and signature creation via a sign there itself if you think that is required for some reason.

src/lean_spec/subspecs/containers/signature.py
```diff
@@ -28,3 +28,22 @@ def verify(
             return scheme.verify(public_key, epoch, message, signature)
         except Exception:
             return False
+
+    @classmethod
+    def from_xmss(
+        cls, xmss_signature: XmssSignature, scheme: GeneralizedXmssScheme = TEST_SIGNATURE_SCHEME
+    ) -> "Signature":
```
anshalshukla: nit: change it to Signature by adding `from __future__ import annotations`. I know this isn't consistent across the repo but direct refs look cleaner.

vyakart:
I’ve incorporated all the requested changes:
	•	Added from __future__ import annotations and updated the return type of Signature.from_xmss to Signature.
	•	Removed the manual raw-signature test; kept only test_from_xmss_roundtrip_with_verify.
	•	Renamed default_activation_epoch → activation_epoch, moved it before max_slot, and updated corresponding doc-comments and constructor API.
	•	Renamed default_num_active_epochs to default_max_epoch and based it on DEFAULT_MAX_SLOT per your suggestion.
	•	Deleted the unused export_test_vectors API as discussed.
	•	Re-ran uvx tox -e pytest and uvx tox -e all-checks, everything passes, and CI is green.
	•	Resolved all review threads.

packages/testing/src/consensus_testing/keys.py
tcoratger: @vyakart Can you fix this?

packages/testing/src/consensus_testing/keys.py
tcoratger: @vyakart Can you fix this?

packages/testing/src/consensus_testing/keys.py
tcoratger: @vyakart ?

packages/testing/src/consensus_testing/keys.py
tcoratger: I think we can keep this for a followup PR

packages/testing/src/consensus_testing/keys.py
tcoratger: @vyakart ?

src/lean_spec/subspecs/containers/signature.py
tcoratger: @vyakart ?

PR #177, opened 2025-11-18
forkchoice store: more documentation and refactoring for `validate_attestation`

tcoratger:
## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/src/lean_spec/subspecs/forkchoice/store.py
+++ b/src/lean_spec/subspecs/forkchoice/store.py
@@ -141,7 +141,10 @@ class Store(Container):
         """
         Validate incoming attestation before processing.
 
-        Performs basic validation checks on attestation structure and timing.
+        Ensures the vote respects the basic laws of time and topology:
+            1. The blocks voted for must exist in our store.
+            2. A vote cannot span backwards in time (source > target).
+            3. A vote cannot be for a future slot.
 
         Args:
             signed_attestation: Attestation to validate.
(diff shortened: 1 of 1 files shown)


src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -141,36 +141,46 @@ def validate_attestation(self, signed_attestation: SignedAttestation) -> None:
...
 
-        # Validate slot relationships
+        # Topology Check
+        #
+        # History is linear and monotonic. Source must be an ancestor of Target.
         source_block = self.blocks[data.source.root]
         target_block = self.blocks[data.target.root]
 
         assert source_block.slot <= target_block.slot, "Source slot must not exceed target"
-        assert data.source.slot <= data.target.slot, "Source checkpoint slot must not exceed target"
```
tcoratger, line 165: Here I think we don't need to check this because in the next lines we are checking

```python
assert source_block.slot == data.source.slot, "Source checkpoint slot mismatch"
assert target_block.slot == data.target.slot, "Target checkpoint slot mismatch"
```

so that it is redundant.

unnawut: Apologies for late review and also a highly semantic nit. Can we remove the line above instead of this one? 

Because `source_block.slot <= target_block.slot` is highly derived from attestation source/target -> block -> slot, but `data.source.slot <= data.target.slot` is directly "A vote cannot span backwards in time (source > target)"

I'm happy to open a PR but asking first since I'm late to this

tcoratger: Yes sure I see your point, this should result in exactly the same check but probably your version is more explicit, can you open a small PR for this?

PR #180, opened 2025-11-18
bitfields: capability to return a list from a slice key

tcoratger:
## 🔗 Related Issues or PRs

Should be a followup after https://github.com/leanEthereum/leanSpec/pull/179

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/src/lean_spec/subspecs/containers/block/block.py
+++ b/src/lean_spec/subspecs/containers/block/block.py
@@ -145,7 +145,7 @@ class SignedBlockWithAttestation(Container):
         # This creates a single list containing both:
         # 1. Block body attestations (from other validators)
         # 2. Proposer attestation (from the block producer)
-        all_attestations = list(block.body.attestations) + [self.message.proposer_attestation]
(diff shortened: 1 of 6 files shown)


src/lean_spec/subspecs/xmss/constants.py
```diff
@@ -131,7 +131,7 @@ def SIGNATURE_LEN_BYTES(self) -> int:  # noqa: N802
     BASE=4,
     FINAL_LAYER=6,
     TARGET_SUM=6,
-    MAX_TRIES=1_000,
+    MAX_TRIES=100_000,
```
unnawut, line 134: Is this intentionally part of the PR?

tcoratger: Yes, that was a mistake I made when testing the configurations on another PR. Using `1_000` sometimes randomly causes the CI to fail, even though it doesn't save any time, so I put `100_000` back.

Issue #172, opened 2025-11-17
xmss: fix serialization/deserialization strategy and tests


tcoratger:
Currently, for the serialization and deserialization strategy of the signature, in the reference implementation, we are using serde derivation pipeline:

https://github.com/b-wagn/hash-sig/blob/23517cf35a4806c93a864f55ace5c4e4a67a08b3/src/signature/generalized_xmss.rs#L40

For consistency, we need to have the same strategy (meaning equivalent results because of course we don't have serde in Python) in the Python specification to be able to compare the results.

Therefore, we should add: 
- a unit test for serialization,
- a unit test for deserialization of the signature to the Python specification, 

comparing them to a result obtained with the Rust implementation to verify they are the same.

If there is a difference (which is expected currently because the strategy in Python looks different for now), then we should fix this difference to make it compatible with the Rust implementation.

tcoratger:
Complete after https://github.com/leanEthereum/leanSpec/pull/210

PR #181, opened 2025-11-18
serialization/deserialization: setup bincode

tcoratger:
## 🗒️ Description

This pull request introduces serialization and deserialization bincode for the public key and XMSS signature scheme to mimic the Rust implementation we have here (https://github.com/leanEthereum/leanSig/blob/2bd2b422dd225aa211fa37a1589b1fa55acb7992/src/signature/generalized_xmss.rs#L40) and here (https://github.com/leanEthereum/leanSig/blob/2bd2b422dd225aa211fa37a1589b1fa55acb7992/src/signature/generalized_xmss.rs#L50).

Numerous tests have been added to this pull request to compare the results obtained in Python with those obtained in Rust to ensure that the serialization behaves exactly the same way and that the results obtained are identical.

Note, however, that a slight difference remains, and it stems from how the finite field is represented in Python.

Here https://github.com/Plonky3/Plonky3/blob/fc1c39e03c20db7ba623bc0d4c4dbd7d789facc8/monty-31/src/monty_31.rs#L155-L168, it is clearly stated that the serialization and deserialization processes are faster when performed directly in Montgomery format so this is done this way in Rust. However, for simplicity, we don't use Montgomery format in Python, so we cannot expect to obtain the same result for the serialization of a given KoalaBear, for example.

However, if we modify the Rust function and remove the Montgomery form, we obtain the same results as those obtained in Python (which validates our serialization implementation). It is thanks to this workaround that the data tests were obtained in Rust for the serialization data of field elements and then copied to Python to compare.

The appropriate documentation is provided within the function.

## 🔗 Related Issues or PRs

Should close https://github.com/leanEthereum/leanSpec/issues/172

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/src/lean_spec/subspecs/xmss/containers.py
+++ b/src/lean_spec/subspecs/xmss/containers.py
@@ -432,8 +432,6 @@ class Signature(StrictBaseModel):
         Raises:
             ValueError: If any component has incorrect length.
         """
-        from . import bincode
-
         # Validate (same as to_bytes)
         if len(self.path.siblings) != config.LOG_LIFETIME:
             raise ValueError(
--- a/tests/lean_spec/subspecs/koalabear/test_field_bincode.py
+++ b/tests/lean_spec/subspecs/koalabear/test_field_bincode.py
@@ -1,14 +1,4 @@
-"""
-Bincode serialization tests for Fp field elements.
-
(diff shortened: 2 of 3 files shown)


tcoratger:
> Do we close this now?

Yes exactly we can close this one.

PR #183, opened 2025-11-19
forkchoice store: some touchups and better doc

tcoratger:
## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/src/lean_spec/subspecs/forkchoice/store.py
+++ b/src/lean_spec/subspecs/forkchoice/store.py
@@ -140,7 +140,7 @@ class Store(Container):
     """
 
     @classmethod
-    def from_state_and_block(cls, state: State, anchor_block: Block) -> "Store":
+    def get_forkchoice_store(cls, state: State, anchor_block: Block) -> "Store":
         """
         Initialize forkchoice store from an anchor state and block.
 
--- a/packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
+++ b/packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
@@ -209,7 +209,7 @@ class ForkChoiceTest(BaseConsensusFixture):
         )
 
         # Initialize Store from anchor
-        store = Store.from_state_and_block(
+        store = Store.get_forkchoice_store(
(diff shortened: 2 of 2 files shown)


src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -96,34 +140,53 @@ class Store(Container):
     """
 
     @classmethod
-    def get_forkchoice_store(cls, state: State, anchor_block: Block) -> "Store":
+    def from_state_and_block(cls, state: State, anchor_block: Block) -> "Store":
```
tcoratger: Looks like a more reasonable choice to name this function what do you think?

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -238,15 +319,25 @@ def on_attestation(
...
-                    del new_new[validator_id]
+            # Fetch any pending ("new") attestation for this validator.
+            existing_new = new_new.get(validator_id)
+
+            # Remove the pending attestation if:
+            # - it exists, and
+            # - it is from an equal or earlier slot than this on-chain attestation.
+            #
+            # In that case, the on-chain attestation supersedes it.
+            if existing_new is not None and existing_new.message.data.slot <= attestation_slot:
```
tcoratger, line 339: We can use `and` between the conditions instead of a nested `if` condition.

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -187,49 +250,67 @@ def on_attestation(
...
+        # Extract the validator index that produced this attestation.
         validator_id = ValidatorIndex(signed_attestation.message.validator_id)
+
+        # Extract the attestation's slot:
+        # - used to decide if this attestation is "newer" than a previous one.
         attestation_slot = signed_attestation.message.data.slot
 
-        # Create copies for immutable update
-        new_known = dict(self.latest_known_attestations)
-        new_new = dict(self.latest_new_attestations)
```
tcoratger, line 232: We don't need to wrap this inside a `dict`

PR #184, opened 2025-11-21
proposer attestation and checks hashes

turuslan:
## 🗒️ Description
- Replace `Block` with `BlockWithProposerAttestation` in fork choice test vectors.
  Allows to read proposer attestation from json, instead of manually recomputing it.
- Fill missing hashes from labels in fork choice test vectors.
  `StoreChecks` json contained labels, but no hashes.

## ✅ Checklist
- [x] Ran `tox` checks to avoid unnecessary CI fails

### DIFF
--- a/packages/testing/src/consensus_testing/test_types/store_checks.py
+++ b/packages/testing/src/consensus_testing/test_types/store_checks.py
@@ -182,9 +182,7 @@ class StoreChecks(CamelModel):
     """
 
     def fill_hash_from_label(self, block_registry: dict[str, "Block"]) -> None:
-        """
-        Resolves block labels to root hashes for head, justified, and finalized checkpoints.
-        """
+        """Resolves block labels to root hashes for head, justified, and finalized checkpoints."""
         from lean_spec.subspecs.ssz import hash_tree_root
 
         # Mapping of {target_field: source_label_field}


packages/testing/src/consensus_testing/test_types/store_checks.py
```diff
@@ -181,8 +181,26 @@ class StoreChecks(CamelModel):
...
+        if self.head_root is None and self.head_root_label is not None:
+            self.head_root = hash_tree_root(block_registry[self.head_root_label])
+        if self.latest_justified_root is None and self.latest_justified_root_label is not None:
+            self.latest_justified_root = hash_tree_root(
+                block_registry[self.latest_justified_root_label]
+            )
+        if self.latest_finalized_root is None and self.latest_finalized_root_label is not None:
+            self.latest_finalized_root = hash_tree_root(
+                block_registry[self.latest_finalized_root_label]
+            )
```
tcoratger: Can't we have something like this?

```suggestion
    def fill_hash_from_label(self, block_registry: dict[str, "Block"]) -> None:
        """
        Resolves block labels to root hashes for head, justified, and finalized checkpoints.
        """
        from lean_spec.subspecs.ssz import hash_tree_root

        # Mapping of {target_field: source_label_field}
        checkpoints = {
            "head_root": "head_root_label",
            "latest_justified_root": "latest_justified_root_label",
            "latest_finalized_root": "latest_finalized_root_label",
        }

        for root_attr, label_attr in checkpoints.items():
            # If root is missing but label exists, resolve it
            if getattr(self, root_attr) is None and (label := getattr(self, label_attr)):
                setattr(self, root_attr, hash_tree_root(block_registry[label]))
```

PR #187, opened 2025-11-23
refactor: change Signature size to 3116

Kayden-ML:
## 🗒️ Description
https://github.com/ReamLabs/ream/issues/922

because where using bincode the serialize size of the signature is actually a maximum of 3116


KolbyML (approved):
Looks good. It looks like Zeam still uses Bytes4000 https://github.com/blockblaz/zeam/issues/381

For Ream we use 3116

PR #188, opened 2025-11-24
containers: mv `produce_attestation` to `Validator`

tcoratger:
## 🔗 Related Issues or PRs

Part of https://github.com/leanEthereum/leanSpec/issues/128

## ✅ Checklist

    ```console
    uvx tox
    ```


KolbyML (commented):
This PR feels weird, as the function is more dependent on Store, then it does to the validator_index.

Anyways I think this function should remain in `Store`, but instead of returning an `Attestation`, it returns `AttestationData`, then a `validator_id`/u64 isn't needed at all.

Then the function in `Validator` can construct the `Attestation` and take `AttestationData` as an argument.

I think this would achieve the goals of this PR, well removing the `validator_id` context as apart of `Store`

tests/lean_spec/subspecs/forkchoice/test_validator.py.bak
KolbyML, line 1: Was this file meant to be committed?

tcoratger: Oups yes thanks, my backup file :)

PR #190, opened 2025-11-25
forkchoice store: add `_compute_lmd_ghost_head` in core store

tcoratger:
## 🗒️ Description

In my opinion it makes little sense to have this method in the helper file because this is the core of the logic and it relies on the blocks of the store, so it makes more sense to have this in the core of the store to maintain clarity.

## ✅ Checklist

    ```console
    uvx tox
    ```


src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -485,6 +484,69 @@ def on_block(self, signed_block_with_attestation: SignedBlockWithAttestation) ->
 
         return store
 
+    def _compute_lmd_ghost_head(
```
varun-doshi, line 487: Is there any reason to put _ at the start here?

tcoratger: Just because this is an internal method that should not be used anywhere else except inside the Store

varun-doshi: Understood
Thank you

PR #210, opened 2025-11-29
xmss: full ssz serialization

tcoratger:
## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/tests/lean_spec/subspecs/ssz/test_signed_attestation.py
+++ b/tests/lean_spec/subspecs/ssz/test_signed_attestation.py
@@ -34,7 +34,18 @@ def test_encode_decode_signed_attestation_roundtrip() -> None:
         ),
     )
 
-    # Test that encoding and decoding round-trips correctly
+    # Test that encoding produces the expected hardcoded value
     encode = signed_attestation.encode_bytes()
+    expected_value = (
+        "000000000000000000000000000000000000000000000000000000000000"
+        "000000000000000000000000000000000000000000000000000000000000"
+        "000000000000000000000000000000000000000000000000000000000000"
+        "000000000000000000000000000000000000000000000000000000000000"
+        "000000000000000000000000000000008c00000024000000000000000000"
+        "000000000000000000000000000000000000000000002800000004000000"
+    )
+    assert encode.hex() == expected_value, "Encoded value must match hardcoded expected value"
+
+    # Test that decoding round-trips correctly
     decoded = SignedAttestation.decode_bytes(encode)
     assert decoded == signed_attestation


tests/lean_spec/subspecs/ssz/test_signed_attestation.py
```diff
@@ -21,10 +27,14 @@ def test_encode_decode_signed_attestation_roundtrip() -> None:
...
+        ),
     )
 
+    # Test that encoding and decoding round-trips correctly
     encode = signed_attestation.encode_bytes()
-    expected_value = "0" * 6504
-    assert encode.hex() == expected_value
-    assert SignedAttestation.decode_bytes(encode) == signed_attestation
+    decoded = SignedAttestation.decode_bytes(encode)
+    assert decoded == signed_attestation
```
KolbyML, line 51: shouldn't we keep an assert against a hardcoded answer here. Just because we want other clients to implement these tests and make sure they get the same results, as the spec does

KolbyML (approved):
I think we should just merge this as is right now, this won't affect devnet1 because we already pinned the version of the spec for our interop here https://github.com/leanEthereum/pm/blob/main/breakout-rooms/leanConsensus/pq-interop/pq-devnet-1.md

tests/lean_spec/subspecs/ssz/test_signed_attestation.py
tcoratger: Yes good idea, just added this with the new ssz implementation results

src/lean_spec/subspecs/containers/signature.py
```diff
@@ -1,30 +0,0 @@
...
-    ) -> bool:
-        """Verify the signature using XMSS verification algorithm."""
-        try:
-            # Signature container is always 3116 bytes, but scheme config may expect less.
-            # Slice to the expected size if needed, assumes padding to the right.
-            signature_data = bytes(self)[: scheme.config.SIGNATURE_LEN_BYTES]
-            signature = XmssSignature.from_bytes(signature_data, scheme.config)
-            return scheme.verify(public_key, epoch, message, signature)
-        except Exception:
-            return False
```
KolbyML, line 30: Shouldn't we be keeping this container?

As from leanSig we will do "Signature(signature.to_bytes())"

"to_bytes()" can be any encodings, Signature is FixedBytes

I believe the hardcoded value for SignAttestation is wrong, because it should be around 6k chars. 

With prod leansig signatures the container should be 3112 bytes, 2kish for the test configuration

KolbyML (changes requested):

src/lean_spec/subspecs/containers/signature.py
tcoratger: For now I would like to avoid having this one because before when we had this it was very confusing because we had two signature classes: one in the XMSS folder and one here that was tied to a specific number of bytes associated to the production configuration.

But this was very messy in terms of flexibility and clarity (for example we had to pad with the testing configuration).

So I would say that we can see in a followup PR if someone has a better idea to make things clean but for now the current way of doing things is pretty minimalist and clear from a spec point of view.

KolbyML: One thing to note is won't the test cases which generate fixtures have invalid signatures, I guess that is fine for now as I assume a lot of changes to the spec need to be done for devnet2. So if we merge this I think we should have a tracking issue that signature serialization is broken, I will dismiss my request, feel free to merge this

tcoratger: Can you open an issue to describe the problem so that someone can start from here?

KolbyML: Sure

KolbyML: > With prod leansig signatures the container should be 3112 bytes, 2kish for the test configuration
Oh nevermind for some reason Ream had a test configuration for LeanSig, but I removed it

KolbyML: > specific number of bytes associated to the production configuration

Are we ever going to use anything other then the production configuration

KolbyML: > Can you open an issue to describe the problem so that someone can start from here?

I think from here we add the `Signature` struct back, and only use the production configuration. If it takes too long to generate keys, we can cache and reuse keys in tests or something.

PR #212, opened 2025-11-30
feat: remove redundant state root calculation

### LINKED ISSUE #211: Redundant process_slot and process_block
There's a redundant `process_slot` and `process_block` calculation in `produce_block_with_signatures`.
Specfifically, this line https://github.com/leanEthereum/leanSpec/blob/10c1ef5a596e47224c1e7e2ad5342cbe5ed22ce2/src/lean_spec/subspecs/forkchoice/store.py#L1019 does an extra call to `process_slot` and https://github.com/leanEthereum/leanSpec/blob/10c1ef5a596e47224c1e7e2ad5342cbe5ed22ce2/src/lean_spec/subspecs/forkchoice/store.py#L1029 does an extra call to `process_block`. 
Both of these can be avoided because the last `state` that we reach inside the loop in this function will be the actual `final post state`.
Essentially, the final post state created here https://github.com/leanEthereum/leanSpec/blob/10c1ef5a596e47224c1e7e2ad5342cbe5ed22ce2/src/lean_spec/subspecs/forkchoice/store.py#L988 will be the real post state after the block is applied.

 I've confirmed this as well using some local testing and the results seems to match this observation

varun-doshi:
## 🗒️ Description

There was an extra `process_slot` and `process_block` breing called from `produce_block_with_signatures`. This removes that redundant call and uses the final state from inside the loop.

## 🔗 Related Issues or PRs

Fixes #211

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/src/lean_spec/subspecs/forkchoice/store.py
+++ b/src/lean_spec/subspecs/forkchoice/store.py
@@ -949,7 +949,6 @@ class Store(Container):
         # Get parent block and state to build upon
         store, head_root = self.get_proposal_head(slot)
         head_state = store.states[head_root]
-        post_state = head_state
 
         # Validate proposer authorization for this slot
         num_validators = Uint64(head_state.validators.count)
@@ -1008,14 +1007,8 @@ class Store(Container):
             attestations.extend(new_attestations)
             signatures.extend(new_signatures)
 
-        # Create final block with all collected attestations
-        final_block = Block(
-            slot=slot,
-            proposer_index=validator_index,
-            parent_root=head_root,
-            state_root=hash_tree_root(post_state),
-            body=BlockBody(attestations=Attestations(data=attestations)),
-        )
+        # Store the post state root in the block
+        final_block = candidate_block.model_copy(update={"state_root": hash_tree_root(post_state)})
 
         # Store block and state immutably
         block_hash = hash_tree_root(final_block)


tcoratger (approved):
Thanks much clearer, the previous approach was really messy.

@unnawut @g11tech can you confirm this?

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -957,6 +949,7 @@ def produce_block_with_signatures(
         # Get parent block and state to build upon
         store, head_root = self.get_proposal_head(slot)
         head_state = store.states[head_root]
+        post_state = head_state
```
unnawut: This `post_state` gets replaced at https://github.com/leanEthereum/leanSpec/pull/212/files#diff-67c42b82824779fcb11cc28349098aa13e096ed5a4b759919fce36a185ffcbe4R981 so maybe it's not needed?


```suggestion
```

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -1016,28 +1009,21 @@ def produce_block_with_signatures(
...
         # Create final block with all collected attestations
-        final_state = head_state.process_slots(slot)
         final_block = Block(
             slot=slot,
             proposer_index=validator_index,
             parent_root=head_root,
-            state_root=Bytes32.zero(),  # Will be updated with computed hash
+            state_root=hash_tree_root(post_state),
             body=BlockBody(attestations=Attestations(data=attestations)),
         )
```
unnawut: We can do a model_copy instead of recreating entire block:


```suggestion
        final_block = candidate_block.model_copy(
            update={"state_root": hash_tree_root(final_post_state)}
        )
```

unnawut (commented):
minor comments but otherwise lgtm

unnawut (approved):

PR #217, opened 2025-12-03
validator: rm validator index alias

tcoratger:
## 🗒️ Description

For me, the fact of having `ValidatorIndex` just as an alias makes more of a mess than anything else because it doesn't enforce anything.

This is not a proper typing enforcement (for that we need a class) and so it brings more confusion than anything else because we can put either `ValidatorIndex` or `Uint64` without the linter complaining, making the alias useless.

So I propose to remove this for the moment and to keep this as an implementation detail that client teams should handle except if we want to enforce a proper class for it where we can implement properly the `is_proposer` method. But I think that this approach has been rejected in https://github.com/leanEthereum/leanSpec/pull/140

## ✅ Checklist

    ```console
    uvx tox
    ```


KolbyML (approved):
I am a big fan of not having an alias over a uint64

PR #213, opened 2025-11-30
feat: separate block production and storage

varun-doshi:
## 🗒️ Description

`produce_block_with_signatures` is a bit misleading since it does block production + storing of blocks. Ideally, we want these functionalities separated. 
This PR separates the functionalities into 2 separate internal functions call atomically by 1 external public function

Also on the path to solve this https://github.com/leanEthereum/leanSpec/blob/10c1ef5a596e47224c1e7e2ad5342cbe5ed22ce2/packages/testing/src/consensus_testing/test_fixtures/fork_choice.py#L300

## ✅ Checklist

- [ ] Ran `tox` checks to avoid unnecessary CI fails:
    ```console
    uvx tox
    ```
- [ ] Considered adding appropriate tests for the changes.
- [ ] Considered updating the online docs in the [./docs/](/leanEthereum/leanSpec/tree/main/docs/) directory.

PR #218, opened 2025-12-03
core: better block production and usage in fixture

tcoratger:
## 🗒️ Description

Now we can use the same block building function in the test fixture that we are using in the spec, making the logic more consistent and more concise.

## 🔗 Related Issues or PRs

Should close https://github.com/leanEthereum/leanSpec/issues/167

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/src/lean_spec/subspecs/containers/state/state.py
+++ b/src/lean_spec/subspecs/containers/state/state.py
@@ -6,6 +6,8 @@ recent blocks, and validator attestations. State also records which blocks are
 justified and finalized.
 """
 
+from typing import TYPE_CHECKING, AbstractSet, Iterable
+
 from lean_spec.subspecs.ssz.hash import hash_tree_root
 from lean_spec.types import (
     ZERO_HASH,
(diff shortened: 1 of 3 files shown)


tcoratger:
@varun-doshi What do you think about this?

We can rework this again in follow up PRs to cleanup the forkchoice store a bit but I think it solves the problems you had in https://github.com/leanEthereum/leanSpec/pull/213 no?

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -516,3 +517,53 @@ def state_transition(self, block: Block, valid_signatures: bool = True) -> "Stat
...
+        slot: Slot,
+        proposer_index: ValidatorIndex,
+        parent_root: Bytes32,
+        attestations: list[Attestation] | None = None,
+    ) -> tuple[Block, "State"]:
+        """
+        Build a valid block on top of this state.
+
+        This is a pure function that:
+        1. Advances the state to the target slot
```
KolbyML: Isn't this function kind of reverting this PR which was just merged https://github.com/leanEthereum/leanSpec/pull/212

In that we are doing redundant `process_slots(slot).process_block` again? In the `produce_block_with_signatures()` function

varun-doshi: Right 
So we are calculating the final post state in `produce_block_with_signatures` in the last iteration of the loop and then we do this calculation again in `build_block`

tcoratger: Yes good remark should be solved now

tcoratger:
@varun-doshi should be solved, let me know what you think

PR #193, opened 2025-11-25
test(signatures): verify signatures in SignedBlockWithAttestation

unnawut:
## 🗒️ Description

Adding test vectors for verifying signatures in `SignedBlockWithAttestation`

#### Some notes:

- Since it heavily resembles other fixtures, I used claude code heavily, but also took a lot of effort to clean it up.

#### Still pending:

- [] ~Production signatures~ Will be done in the next PR

#### Usage

Feed `signedBlockWithAttestation` and `anchorState` into [`SignedBlockWithAttestation.verify_signatures()`](https://github.com/leanEthereum/leanSpec/blob/6bfdbfc883de7f3785c2b6a08de6d5d6dd11e3ba/src/lean_spec/subspecs/containers/block/block.py#L118), the function should return `true`. When `expectedException` is provided, expect that the signature verification should throw an exception.

The generated vectors look like this:

```json
{
    "tests/consensus/devnet/verify_signature/test_valid_signatures.py::test_proposer_and_attester_signatures[fork_Devnet][fork_Devnet-verify_signature_test]": {
        "network": "Devnet",
        "anchorState": {
            "config": {
                "genesisTime": 0
            },
            "slot": 0,
            "latestBlockHeader": {
                "slot": 0,
                "proposerIndex": 0,
                "parentRoot": "0x0000000000000000000000000000000000000000000000000000000000000000",
                "stateRoot": "0x0000000000000000000000000000000000000000000000000000000000000000",
                "bodyRoot": "0xdba9671bac9513c9482f1416a53aabd2c6ce90d5a5f865ce5a55c775325c9136"
            },
            "latestJustified": {
                "root": "0x0000000000000000000000000000000000000000000000000000000000000000",
                "slot": 0
            },
            "latestFinalized": {
                "root": "0x0000000000000000000000000000000000000000000000000000000000000000",
                "slot": 0
            },
            "historicalBlockHashes": {
                "data": []
            },
            "justifiedSlots": {
                "data": []
            },
            "validators": {
                "data": [
                    {
                        "pubkey": "0xb40ec9783b56572d99965130ad1cec4f066df43a0fd89d22d420535c62b52711cdaf7f5f696c6b1e4df66a6bab48bb51abcdd77e",
                        "index": 0
                    },
                    {
                        "pubkey": "0xc5ed4b6fa46e83281e29201b878ad52b4a9b8a2f3c9d8017cfd9b81e8097e1372a097e62be37194c39e2f36f76e8906d2e818238",
                        "index": 1
[...]

PR #223, opened 2025-12-05
feat: make SSZVector of Fp serializable

unnawut:
## 🗒️ Description

I was having issue with getting #193 to output the signatures to test vectors. It looks like SSZVector didn't know how to serialize Fp. 3 main changes are done:

- ~`Fp.__get_pydantic_core_schema__()` is implemented similar to [`BaseUint.__get_pydantic_core_schema__()`](https://github.com/leanEthereum/leanSpec/blob/main/src/lean_spec/types/uint.py#L38-L66)~ Reverted. Turns out it's not needed anymore
- `SSZVector._serialize_data()` is implemented similar to [`SSZList._serialize_data()`](https://github.com/leanEthereum/leanSpec/blob/main/src/lean_spec/types/collections.py#L188-L200)
- `SSZList._serialize_data()` updated to support handling `Fp` items (only for consistency)

See how they are being used in #193.

## ✅ Checklist

    ```console
    uvx tox
    ```


unnawut:
@tcoratger Sorry I forgot to add. Would this be enough? https://github.com/leanEthereum/leanSpec/pull/223/commits/728f736abc70b7513e7cfd127bed0711514e63a8

PR #230, opened 2025-12-09
framework: ability to use test or prod signature scheme in tests

unnawut:
## 🗒️ Description

In this PR, there are no more independent references to `TEST_SIGNATURE_SCHEME` or `PROD_SIGNATURE_SCHEME` in the test vector framework nor the test cases. All usage go through `lean_spec.config.LEAN_ENV`. This should help ensure that the signature scheme setting is consistent across all tests.

Defaults around key managers and signature operations are removed so the configured scheme is always trickled down from the starting point than trying to default mid-way. The exception is in the client subspecs like `signed_block_with_attestation.verify_signatures()` and `store.on_block()` where they default to `PROD_SIGNATURE_SCHEME` for specs clarity.

### Key changes:

1. Consolidated `_get_shared_key_manager()` from each fixture into `consensus_testing/keys.py` so there's only one place to set the signature scheme for the key manager
2. Added `LEAN_ENV` env var that takes in either "test" or "prod". I avoid using `SIGNATURE_SCHEME` or similar because it gets really ambiguous when it's used in the xmss subspecs.
3. `LEAN_ENV` parsing:
    - For subspecs, `LEAN_ENV` is parsed only once by `xmss.constants`. This ensures the entire subspec uses a consistent value.
    - For tests, `LEAN_ENV` is defaulted to `LEAN_ENV="test"` in `tests/conftest.py`. It's then parsed and consumed mainly by `BaseFixture.lean_env()` as a computed field. There's an additional parsing by `test_types.genesis.generate_pre_state()` because the function is independent from the fixtures.
4. If `LEAN_ENV` is not set:
    - For subspecs, it defaults to `TARGET_CONFIG=PROD_CONFIG` and `TARGET_SIGNATURE_SCHEME=PROD_SIGNATURE_SCHEME`, so specs are consumed with prod config/scheme by default.
    - For tests, it defaults `LEAN_ENV="test"`, so tests are run in test config/scheme by default, which in turn would also configure subspecs to use test config.
5. Splitted `test-keys.json` into individual key files. This helps with managing prod keys that would've been ~200MB for 12 validators in a single file.
6. Added `consensus.keys --download` option in addition to the existing keygen so pre-generated prod keys can be downloaded.
    - The pre-generated keys currently live in https://github.com/leanEthereum/leansig-test-keys releases. Thinking this is the simplest solution without adding a new kind of infra.
7. Added CI workflow that triggers test filling with prod keys on a main branch commit, export the vectors as as a build artifact

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/src/lean_spec/subspecs/containers/block/block.py
+++ b/src/lean_spec/subspecs/containers/block/block.py
@@ -12,6 +12,7 @@ can propose.
 from typing import TYPE_CHECKING
 
 from lean_spec.subspecs.containers.slot import Slot
+from lean_spec.subspecs.xmss.interface import TARGET_SIGNATURE_SCHEME, GeneralizedXmssScheme
 from lean_spec.types import Bytes32, Uint64
 from lean_spec.types.container import Container
 
@@ -192,6 +193,7 @@ class SignedBlockWithAttestation(Container):
                     validator.get_pubkey(),
                     aggregated_attestation.data.slot,
                     attestation_root,
+                    scheme,
                 ), "Attestation signature verification failed"
 
         # Verify proposer attestation signature
@@ -206,6 +208,7 @@ class SignedBlockWithAttestation(Container):
             proposer.get_pubkey(),
             proposer_attestation.data.slot,
             proposer_attestation.data.data_root_bytes(),
+            scheme,
         ), "Proposer signature verification failed"
 
         return True


unnawut:
@fselmo @tcoratger Some updates to the PR:

1. I've replaced `--scheme` flag with `LEAN_ENV` env var that takes in either "test" or "prod". I avoid using `SIGNATURE_SCHEME` or similar because it gets really ambiguous when it's used in the xmss subspecs.
2. `LEAN_ENV` parsing:
    - For subspecs, `LEAN_ENV` is parsed only once by `xmss.constants`. This ensures the entire subspec uses a consistent value.
    - For tests, `LEAN_ENV` is defaulted to `LEAN_ENV="test"` in `tests/conftest.py`. It's then parsed and consumed mainly by `BaseFixture.lean_env()` as a computed field. There's an additional parsing by `test_types.genesis.generate_pre_state()` because the function is independent from the fixtures.
3. If `LEAN_ENV` is not set:
    - For subspecs, it defaults to `TARGET_CONFIG=PROD_CONFIG` and `TARGET_SIGNATURE_SCHEME=PROD_SIGNATURE_SCHEME`, so specs are consumed with prod config/scheme by default.
    - For tests, it defaults `LEAN_ENV="test"`, so tests are run in test config/scheme by default, which in turn would also configure subspecs to use test config.
4. Splitted `test-keys.json` into individual key files. This helps with managing prod keys that would've been ~200MB for 12 validators in a single file.
5. Added `consensus.keys --download` option in addition to the existing keygen so pre-generated prod keys can be downloaded. The pre-generated keys currently live in https://github.com/unnawut/leansig-test-keys releases for ease of
6. Added CI workflow that triggers test filling with prod keys on a main branch commit, export the vectors as as a build artifact

I havn't broken down to smaller PRs because the approach changes how different components interact. Happy to break down to smaller PRs if the overall approach looks okay.

packages/testing/src/consensus_testing/test_types/genesis.py
```diff
@@ -40,7 +24,15 @@ def generate_pre_state(**kwargs: Any) -> State:
...
     num_validators = kwargs.get("num_validators", 4)
-    key_manager = _get_shared_key_manager()
+
+    lean_env = os.environ.get("LEAN_ENV", "test").lower()
+    key_manager = get_shared_key_manager(lean_env)
+    available_keys = len(key_manager)
+
+    assert num_validators <= available_keys, (
+        "Not enough keys to generate state",
+        f"Expecting {num_validators} validators but the key manager has only {available_keys} keys",
```
tcoratger:
```suggestion
        f"Expecting a minimum of {num_validators} validators but the key manager has only {available_keys} keys",
```

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -186,14 +169,9 @@ def make_fixture(self) -> ForkChoiceTest:
...
-        # Use shared key manager if it has sufficient capacity, otherwise create a new one
-        # This optimizes performance by reusing keys across tests when possible
-        shared_key_manager = _get_shared_key_manager()
-        key_manager = (
-            shared_key_manager
-            if self.max_slot <= shared_key_manager.max_slot
-            else XmssKeyManager(max_slot=self.max_slot, scheme=TEST_SIGNATURE_SCHEME)
-        )
+        # Get shared key manager with the required max_slot
+        # This optimizes performance by reusing keys across tests with the same max_slot
```
tcoratger: Here could we replace max_slot by "maximum slot" or something without having explicitly the name of the variable in the doc so that we are free to rename things in the future?

tcoratger (approved):
Sounds super cool, thanks a lot, just two super small nits for the rest, looks good to me.

packages/testing/src/framework/test_fixtures/base.py
```diff
@@ -55,6 +59,20 @@ def __pydantic_init_subclass__(cls, **kwargs: Any) -> None:
         if cls.format_name:
             BaseFixture.formats[cls.format_name] = cls
 
+    @computed_field  # type: ignore[prop-decorator]
+    @cached_property
+    def lean_env(self) -> str:
+        """The target lean environment (e.g. 'test' or 'prod')."""
+        lean_env = os.environ.get("LEAN_ENV", "").lower()
```
fselmo: Should we default to `test` here? 👀 
```suggestion
        lean_env = os.environ.get("LEAN_ENV", "test").lower()
```

packages/testing/src/framework/test_fixtures/base.py
```diff
@@ -55,6 +59,20 @@ def __pydantic_init_subclass__(cls, **kwargs: Any) -> None:
         if cls.format_name:
             BaseFixture.formats[cls.format_name] = cls
 
+    @computed_field  # type: ignore[prop-decorator]
```
fselmo: I want to understand the goal here a bit more. I think I was originally a bit more uncertain about this, but the env var is set once at import time only right? So if we have this as cached _and_ as a computed field, it will still cache the very first time it's set, iiuc.

What if we made this a regular property and _didn't_ cache it? Something like:

```py
from lean_spec.subspecs.xmss import LEAN_ENV


class BaseFixture(CamelModel):
    lean_env: str = Field(default=LEAN_ENV)
    ...
```

I could very well be missing something. Lmk if that makes sense.

tests/conftest.py
```diff
@@ -1,7 +1,13 @@
 """Pytest configuration and shared fixtures."""
 
+import os
+
 from hypothesis import settings
 
+# Set LEAN_ENV=test before importing any modules that depend on it
+if "LEAN_ENV" not in os.environ:
```
fselmo, line 7: If we do default to `test`, we can remove this.

fselmo (commented):
This seems great 👌🏼. Very similar to what I was imagining as well. I just left some small things to think about.

I think also it could be nicer for DevEx if we had a flag for `fill`, something like `--scheme` that overrides the env var for that run perhaps. If you ran with `--scheme=prod` it would override any sort of `test` env var you had just for that run.

I think it would be as simple as adding a `scheme` flag to the click API for `fill` and doing something like:

```py
 if scheme:
        os.environ["LEAN_ENV"] = scheme.lower()
```

So the order of precedence would be:

- CLI flag (sets `os.environ["LEAN_ENV"]` which is only used on current run)
- If no CLI flag, reads from `LEAN_ENV` from global user settings
- If no `LEAN_ENV`, defaults to `test`

Thoughts? Either way looks good from my end 👍🏼

unnawut:
> edit: Ah, it looks like it was in there at one point but then removed in [6eb89e1](https://github.com/leanEthereum/leanSpec/commit/6eb89e1c5df5a10df5bb1295c70c647b44636cc8).
> 
> > Added --scheme flag to the test filler. This is where the scheme gets defined for usage
> 
> Curious on your thoughts on keeping it vs removing 👀

I also liked to use `--scheme` than setting the env var because it feels more natural to me. My thinking was that I didn't want to pre-emptively add multiple ways of achieving the same behavior unless suggested. And `LEAN_ENV` was the single way that could apply globally to both the subspecs and the tests and hence why I went with env var and removed the `--scheme` flag which is only applicable to tests.

But since you suggested + also my preference, let's add it back!

packages/testing/src/framework/test_fixtures/base.py
unnawut: Sounds good to me. I didn't this of this. Also I think it doesn't make sense for LEAN_ENV to be in xmss subspecs so I just moved it to `lean_spec.config.LEAN_ENV`. Also a bit of refactor now the env var is read in only 2 places: 1 for the specs and 1 for the tests.

Updated in https://github.com/leanEthereum/leanSpec/pull/230/changes/94c17a1d057aba87bd78117866cc2b3234b96d66

tests/conftest.py
fselmo: Ah yeah this is true of `pytest` where we need to set it before importing any modules. I think this needs to stay 👍🏼

tests/conftest.py
```diff
@@ -1,7 +1,12 @@
 """Pytest configuration and shared fixtures."""
 
+import os
+
 from hypothesis import settings
 
+if "LEAN_ENV" not in os.environ:
+    os.environ["LEAN_ENV"] = "test"
+
 import lean_spec.subspecs.xmss as xmss
```
fselmo: This can be removed I think as it's not used anywhere

fselmo:
Sounds good, yeah I approve :)

PR #231, opened 2025-12-10
Sig agg precursor

anshalshukla:
## 🗒️ Description
Updates containers and related functionality to enable signature aggregation.
- Changes validator signing to `AttestationData` signing from `Attestation` signing as `Attestation` contains `validator_id` making it incompatible for leanVm aggregation
- Updates `Block` to store naive `aggregatedAttestationList`, the `aggregation_bits` is set for all the `validator_id`s that have signed the `AttestationData`
- Introduces `BlockSignature` container to store `attestation_signatures` and `proposer_signature`. `attestation_signature` is currently a naive list of signatures corresponding to `aggregatedAttestationList` in block.
- It is expected that the `signatures` in `attestation_signatures` is in order of validator keys
- `attestation_signature` will be updated to store a list of of proofs in followup PR    

## ✅ Checklist
    ```console
    uvx tox
    ```

### DIFF
--- a/packages/testing/src/consensus_testing/keys.py
+++ b/packages/testing/src/consensus_testing/keys.py
@@ -36,7 +36,6 @@ from lean_spec.subspecs.containers.block.types import (
     AttestationSignatures,
 )
 from lean_spec.subspecs.containers.slot import Slot
-from lean_spec.subspecs.ssz.hash import hash_tree_root
 from lean_spec.subspecs.xmss.containers import PublicKey, SecretKey, Signature
 from lean_spec.subspecs.xmss.interface import TEST_SIGNATURE_SCHEME, GeneralizedXmssScheme
 from lean_spec.types import Uint64
@@ -210,7 +209,7 @@ class XmssKeyManager:
         self._state[validator_id] = kp.with_secret(sk)
 
         # Sign hash tree root of the attestation data
-        message = bytes(hash_tree_root(attestation_data))
+        message = attestation_data.data_root_bytes()
         return self.scheme.sign(sk, epoch, message)
 
     def build_attestation_signatures(
@@ -218,43 +217,20 @@ class XmssKeyManager:
         aggregated_attestations: AggregatedAttestations,
         signature_lookup: Mapping[tuple[Uint64, bytes], Signature] | None = None,
     ) -> AttestationSignatures:
-        """
-        Build `AttestationSignatures` for already-aggregated attestations.
-
-        This is a convenience helper for tests/fixtures that need to produce
-        `BlockSignatures.attestation_signatures` for a block.
-
-        Args:
-            aggregated_attestations: Iterable of aggregated attestation containers.
-                Each item is expected to have:
-                - `.data` (AttestationData)
-                - `.aggregation_bits.to_validator_indices()` (Iterable[Uint64])
-            signature_lookup: Optional override map keyed by
-                `(validator_id, bytes(hash_tree_root(attestation_data))) -> signature`.
-                When provided and a key exists, that signature is used instead of signing.
-
-        Returns:
-            AttestationSignatures matching the ordering of `aggregated_attestations`
-            and per-attestation validator index ordering.
-        """
+        """Build `AttestationSignatures` for already-aggregated attestations."""
+        lookup = signature_lookup or {}
         return AttestationSignatures(
             data=[
                 NaiveAggregatedSignature(
                     data=[
                         (
-                            signature_lookup.get(
-                                (validator_id, aggregated_attestation.data.data_root_bytes())
-                            )
-                            if signature_lookup is not None
-                            else None
+                            lookup.get((vid, agg.data.data_root_bytes()))
+                            or self.sign_attestation_data(vid, agg.data)
                         )
-                        or self.sign_attestation_data(validator_id, aggregated_attestation.data)
-                        for validator_id in (
-                            aggregated_attestation.aggregation_bits.to_validator_indices()
-                        )
+                        for vid in agg.aggregation_bits.to_validator_indices()
                     ]
                 )
-                for aggregated_attestation in aggregated_attestations
+                for agg in aggregated_attestations
             ]
         )
 
--- a/packages/testing/src/consensus_testing/test_fixtures/verify_signatures.py
+++ b/packages/testing/src/consensus_testing/test_fixtures/verify_signatures.py
@@ -210,7 +210,7 @@ class VerifySignaturesTest(BaseConsensusFixture):
         # for any intentionally-invalid signature from the input spec remains invalid
         # in the produced `SignedBlockWithAttestation`.
         signature_lookup: dict[tuple[Uint64, bytes], Signature] = {
-            (att.validator_id, bytes(hash_tree_root(att.data))): sig
+            (att.validator_id, att.data.data_root_bytes()): sig
             for att, sig in zip(attestations, attestation_signature_inputs, strict=True)
         }
         attestation_signatures = key_manager.build_attestation_signatures(
--- a/src/lean_spec/subspecs/containers/block/block.py
+++ b/src/lean_spec/subspecs/containers/block/block.py
@@ -12,7 +12,6 @@ can propose.
 from typing import TYPE_CHECKING, cast
 
 from lean_spec.subspecs.containers.slot import Slot
-from lean_spec.subspecs.ssz.hash import hash_tree_root
 from lean_spec.types import Bytes32, Uint64
 from lean_spec.types.container import Container
 
@@ -179,7 +178,7 @@ class SignedBlockWithAttestation(Container):
                 "Aggregated attestation signature count mismatch"
             )
 
-            attestation_root = bytes(hash_tree_root(aggregated_attestation.data))
+            attestation_root = aggregated_attestation.data.data_root_bytes()
 
             # Verify each validator's attestation signature
             for validator_id, signature in zip(validator_ids, aggregated_signature, strict=True):
@@ -204,7 +203,7 @@ class SignedBlockWithAttestation(Container):
         assert proposer_signature.verify(
             proposer.get_pubkey(),
             proposer_attestation.data.slot,
-            bytes(hash_tree_root(proposer_attestation.data)),
+            proposer_attestation.data.data_root_bytes(),
         ), "Proposer signature verification failed"
 
         return True


packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -13,13 +13,17 @@
...
+from lean_spec.subspecs.containers.block import (
     Block,
     BlockBody,
+    BlockSignatures,
     BlockWithAttestation,
     SignedBlockWithAttestation,
 )
-from lean_spec.subspecs.containers.block.types import Attestations, BlockSignatures
+from lean_spec.subspecs.containers.block.types import (
+    AggregatedAttestationsList,
```
g11tech:
```suggestion
    AggregatedAttestations,
```
we can just call list this of `AggregatedAttestation` (no `s` here)

src/lean_spec/subspecs/containers/attestation/attestation.py
```diff
@@ -82,5 +85,44 @@ class SignedAggregatedAttestations(Container):
...
+    - this will be replaced by a SNARK in future devnets.
+    - this will be aggregated by aggregators in future devnets.
+    """
+
+
+def _aggregation_bits_to_validator_index(bits: AggregationBits) -> Uint64:
+    """
+    Extract the single validator index encoded in aggregation bits.
+
+    Current devnets only support naive aggregation where every attestation
```
g11tech: this seems wrong, aggregation bits encode all validator indices witht he same aggregation data

src/lean_spec/subspecs/containers/attestation/attestation.py
```diff
@@ -82,5 +85,44 @@ class SignedAggregatedAttestations(Container):
...
+    return _aggregation_bits_to_validator_index(bits)
+
+
+def aggregated_attestation_to_plain(aggregated: AggregatedAttestations) -> Attestation:
+    """Convert aggregated attestation data to the plain Attestation container."""
+    validator_index = _aggregation_bits_to_validator_index(aggregated.aggregation_bits)
+    return Attestation(validator_id=validator_index, data=aggregated.data)
+
+
+def attestation_to_aggregated(attestation: Attestation) -> AggregatedAttestations:
```
g11tech:
```suggestion
def attestations_to_aggregated(attestation: Attestation) -> AggregatedAttestations:
```
for all the attestations with the same attestation data

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -457,18 +457,25 @@ def on_block(self, signed_block_with_attestation: SignedBlockWithAttestation) ->
...
+
+        for aggregated_attestation, signature in zip(
+            block_attestations, attestation_signatures, strict=True
         ):
-            # Process as on-chain attestation (immediately becomes "known")
+            validator_id = aggregation_bits_to_validator_index(
+                aggregated_attestation.aggregation_bits
+            )
+
             store = store.on_attestation(
```
g11tech: we will comment out this line, because we will no longer import attestation from the block to the forkchoice because signature will be aggregated one, and we will need recursive aggregation for the signature

also note the same as comment over here

src/lean_spec/subspecs/containers/attestation/attestation.py
anshalshukla: I've added a different function for that, this function can stay as it serves as a helper function

src/lean_spec/subspecs/containers/attestation/attestation.py
anshalshukla: In the earlier version I was not aggregating the attestations and was treating it as a naive list with single set bit, I've updated the logic to aggregate attestations now

src/lean_spec/subspecs/containers/attestation/attestation.py
```diff
@@ -82,5 +87,103 @@ class SignedAggregatedAttestations(Container):
...
+    """
+    validator_indices = [Uint64(index) for index, bit in enumerate(bits) if bool(bit)]
+    if not validator_indices:
+        raise AssertionError("Aggregated attestation must reference at least one validator")
+    return validator_indices
+
+
+def aggregated_attestations_to_plain(
+    aggregated: AggregatedAttestation,
+) -> list[Attestation]:
```
tcoratger: We should have this implemented directly in the `AggregatedAttestation` class like a `to_plain` method.

src/lean_spec/subspecs/containers/attestation/attestation.py
```diff
@@ -82,5 +87,103 @@ class SignedAggregatedAttestations(Container):
...
 
-    TODO: this will be replaced by a SNARK in future devnets.
+    TODO:
+    - signatures will be replaced by MegaBytes in next PR to include leanVM proof.
+    - this will be replaced by a SNARK in future devnets.
+    - this will be aggregated by aggregators in future devnets.
     """
+
+
+def aggregation_bits_to_validator_indices(bits: AggregationBits) -> list[Uint64]:
```
tcoratger: We should have this implemented in `AggregationBits` with a `to_validator_indices` method.

src/lean_spec/subspecs/containers/attestation/attestation.py
```diff
@@ -82,5 +87,103 @@ class SignedAggregatedAttestations(Container):
...
+    Args:
+        bits: Aggregation bitlist with participating validators.
+
+    Returns:
+        List of validator indices, sorted in ascending order.
+    """
+    validator_indices = [Uint64(index) for index, bit in enumerate(bits) if bool(bit)]
+    if not validator_indices:
+        raise AssertionError("Aggregated attestation must reference at least one validator")
+    return validator_indices
```
tcoratger:
```suggestion
    if not (indices := [Uint64(i) for i, bit in enumerate(bits) if bit]):
        raise AssertionError("Aggregated attestation must reference at least one validator")
    
    return indices
```

src/lean_spec/subspecs/containers/attestation/attestation.py
```diff
@@ -82,5 +87,103 @@ class SignedAggregatedAttestations(Container):
...
+        List of plain attestations, one per participating validator.
+    """
+    validator_indices = aggregation_bits_to_validator_indices(aggregated.aggregation_bits)
+    return [
+        Attestation(validator_id=validator_id, data=aggregated.data)
+        for validator_id in validator_indices
+    ]
+
+
+def attestation_to_aggregated(attestation: Attestation) -> AggregatedAttestation:
```
tcoratger: We should have this one implemented in `Attestation` directly as a `to_aggregated` method.

src/lean_spec/subspecs/containers/attestation/attestation.py
```diff
@@ -82,5 +87,103 @@ class SignedAggregatedAttestations(Container):
...
+
+    # Create aggregated attestations with all relevant bits set
+    result: list[AggregatedAttestation] = []
+
+    for data, validator_ids in data_to_validator_ids.items():
+        # Find the maximum validator index to determine bitlist size
+        max_validator_id = max(validator_ids)
+
+        # Create bitlist with all participating validators set to True
+        bits = [False] * (max_validator_id + 1)
```
g11tech: this is wrong, ~~this way all the bits will be flipped to true~~, you have to use max registry limit as the max validator count

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -352,16 +357,31 @@ def process_block(self, block: Block) -> "State":
...
+
+            attestations_data.add(aggregated_att.data)
+            attestations.extend(aggregated_attestations_to_plain(aggregated_att))
+
+        return state.process_attestations(attestations)
 
     def process_attestations(
         self,
-        attestations: Attestations,
+        attestations: Iterable[Attestation],
```
tcoratger: I think here that is better if we have a proper typing so that  we can implement `aggregate_attestations_by_data` inside this class (before it was `Attestations`), why do we need an Iterable here specifically?

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -687,7 +715,20 @@ def build_block(
             attestations.extend(new_attestations)
             signatures.extend(new_signatures)
 
+        # Sort attestations and signatures by validator_id to ensure
```
```suggestion
        # Sort attestations and signatures by validator id to ensure
```

src/lean_spec/subspecs/containers/attestation/attestation.py
g11tech: ahh ok, this is sort of an optimization where the list doesn't go high if the the max validator id in that list is low
add a comment to this effect: that on an average case for packing sparse attestations this size could be lower than full list size

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -687,7 +715,20 @@ def build_block(
...
+        # consistent ordering for signature validation.
+        # Only sort if we have signatures (i.e., attestation collection was performed)
+        if signatures:
+            attestations_with_sigs = sorted(
+                zip(attestations, signatures, strict=True), key=lambda x: x[0].validator_id
+            )
+            sorted_attestations = [att for att, _ in attestations_with_sigs]
+            sorted_signatures = [sig for _, sig in attestations_with_sigs]
```
tcoratger: Here we could group the iterators together no? this is about iterating over the same thing no?

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -457,18 +457,29 @@ def on_block(self, signed_block_with_attestation: SignedBlockWithAttestation) ->
...
+            plain_att
+            for aggregated_att in block_attestations
+            for plain_att in aggregated_attestations_to_plain(aggregated_att)
+        ]
+
+        # Sort attestations by validator_id to align with signature order
+        plain_attestations.sort(key=lambda att: att.validator_id)
+
+        assert len(plain_attestations) == len(attestation_signatures), (
+            "Attestation signature list must align with validator attestations"
```
```suggestion
            "Attestation signature list lenght must align with the number of validator attestations"
```

src/lean_spec/subspecs/containers/attestation/attestation.py
tcoratger, line 1: In general, in this file I think it would be better to implement the methods inside the various classes to continue our effort to build a strongly typed Python spec.

I would love to limit the number of isolated methods like this. It makes things heavier to read, less types and with more imports because we have to import all these methods in other files to be able to use them.

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -344,15 +348,23 @@ def _build_block_from_spec(
         )
 
         # Sign all attestations and the proposer attestation
-        signature_list = [key_manager.sign_attestation(att) for att in attestations]
-        signature_list.append(key_manager.sign_attestation(proposer_attestation))
+        attestation_signatures = [
```
g11tech: all attestations here have same data? since we are forming a single list?

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -344,15 +348,23 @@ def _build_block_from_spec(
...
+        )
 
         return SignedBlockWithAttestation(
             message=BlockWithAttestation(
                 block=final_block,
                 proposer_attestation=proposer_attestation,
             ),
-            signature=BlockSignatures(data=signature_list),
+            signature=BlockSignatures(
+                attestation_signatures=NaiveAggregatedSignature(data=attestation_signatures),
```
g11tech:
```suggestion
                attestation_signatures=NaiveAggregatedSignature(data=attestation_signatures),
```
this is a list of NaiveAggregatedSignature

src/lean_spec/subspecs/containers/attestation/attestation.py
```diff
@@ -45,18 +54,31 @@ class Attestation(Container):
     data: AttestationData
     """The attestation data produced by the validator."""
 
+    def to_aggregated(self) -> AggregatedAttestation:
```
g11tech: where is this used?

src/lean_spec/subspecs/containers/attestation/attestation.py
```diff
@@ -69,11 +91,52 @@ class AggregatedAttestations(Container):
...
+        return [
+            cls(
+                aggregation_bits=AggregationBits.from_validator_indices(validator_ids),
+                data=data,
+            )
+            for data, validator_ids in data_to_validator_ids.items()
+        ]
+
 
 class SignedAggregatedAttestations(Container):
```
g11tech:
```suggestion
class SignedAggregatedAttestation(Container):
```
since message is AggregatedAttestation

src/lean_spec/subspecs/containers/attestation/attestation.py
anshalshukla: I think i was using it earlier in one of the commit, will remove it as it is not being used anywhere now

src/lean_spec/subspecs/containers/attestation/attestation.py
```diff
@@ -69,11 +81,52 @@ class AggregatedAttestations(Container):
...
+
+
+class SignedAggregatedAttestation(Container):
     """Aggregated attestation bundled with aggregated signatures."""
 
-    message: AggregatedAttestations
+    message: AggregatedAttestation
     """Aggregated attestation data."""
 
     signature: AggregatedSignatures
```
g11tech: are we calling it `AggregatedSignatures` or `NaiveAggregatedSignature` (without s) because only one of these types are to be used?

src/lean_spec/subspecs/containers/block/__init__.py
```diff
@@ -4,17 +4,24 @@
...
     BlockHeader,
+    BlockSignatures,
     BlockWithAttestation,
     SignedBlockWithAttestation,
 )
-from .types import Attestations, BlockSignatures
+from .types import (
+    AggregatedAttestations,
+    AttestationSignatures,
+    NaiveAggregatedSignatures,
```
g11tech:
```suggestion
    NaiveAggregatedSignature,
```
lets remove `s` from it because its one aggregation (as a list)

src/lean_spec/subspecs/containers/attestation/attestation.py
anshalshukla: This is `SignedAggregatedAttestation` I'm not using it anywhere it is for future devents when the aggregator role is introduced.

Right now in `BlockSignatures` we have `AttestationSignatures` which is list of `NaiveAggregatedSignature`~s~ which is list of `XmssSignature`

packages/testing/src/consensus_testing/keys.py
```diff
@@ -198,12 +207,57 @@ def sign_attestation(self, attestation: Attestation) -> Signature:
             prepared = self.scheme.get_prepared_interval(sk)
 
         # Cache advanced state
-        self._state[idx] = kp.with_secret(sk)
+        self._state[validator_id] = kp.with_secret(sk)
 
-        # Sign hash tree root
-        message = bytes(hash_tree_root(attestation))
+        # Sign hash tree root of the attestation data
+        message = bytes(hash_tree_root(attestation_data))
```
tcoratger: I think we have a bunch of occurence where this could be used?

```suggestion
        message = attestation_data.data_root_bytes()
```

packages/testing/src/consensus_testing/keys.py
```diff
@@ -198,12 +207,57 @@ def sign_attestation(self, attestation: Attestation) -> Signature:
...
+                        )
+                        or self.sign_attestation_data(validator_id, aggregated_attestation.data)
+                        for validator_id in (
+                            aggregated_attestation.aggregation_bits.to_validator_indices()
+                        )
+                    ]
+                )
+                for aggregated_attestation in aggregated_attestations
+            ]
+        )
```
tcoratger, line 235: Something like this should be simpler and totally fine for tests

```suggestion
    def build_attestation_signatures(
        self,
        aggregated_attestations: AggregatedAttestations,
        signature_lookup: Mapping[tuple[Uint64, bytes], Signature] | None = None,
    ) -> AttestationSignatures:
        """
        Build `AttestationSignatures` for already-aggregated attestations.
        """
        lookup = signature_lookup or {}

        return AttestationSignatures(data=[
            NaiveAggregatedSignature(data=[
                lookup.get((vid, agg.data.data_root_bytes())) or self.sign_attestation_data(vid, agg.data)
                for vid in agg.aggregation_bits.to_validator_indices()
            ])
            for agg in aggregated_attestations
        ])
```

packages/testing/src/consensus_testing/keys.py
anshalshukla: nice catch, I had added the method specifically to address this but forgot to replace it all the places

tcoratger (approved):
Looks good to me! This PR is pretty big but looks very nice so let us merge and continue to fix things as we go in some followup PRs so that we don't block progress.

Issue #60, opened 2025-09-19
Look into replacing `mypy` with `ty` once released


fselmo:
Astral (builders of `uv` and `ruff`) is working on a type-checker, `ty`, that could replace `mypy` and utilize rust for performance. Look into using this here once it is released (currently in preview / alpha).

https://github.com/astral-sh/ty

tcoratger:
@fselmo As per https://github.com/astral-sh/ty/pull/1606, looks like we could use `ty` instead of `mypy` now no?

PR #235, opened 2025-12-15
python: replace mypy by ty

tcoratger:
## 🔗 Related Issues or PRs

Should close https://github.com/leanEthereum/leanSpec/issues/60

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/tests/lean_spec/types/test_bitfields.py
+++ b/tests/lean_spec/types/test_bitfields.py
@@ -1,14 +1,39 @@
 """ "Tests for the Bitvector and Bitlist types."""
 
 import io
-from typing import Any, cast
+from typing import Any
 
 import pytest
-from pydantic import ValidationError, create_model
+from pydantic import BaseModel, ValidationError
 from typing_extensions import Tuple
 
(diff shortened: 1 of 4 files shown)


fselmo (approved):
@tcoratger I added a cleanup commit and a nit commit directly after that replaces `cast(Any, ...)` logic with more explicit ignores and / or class definitions. These were only in the tests so I don't mind either way tbh. If you feel casting to `Any` is better or easier to read, feel free to drop that commit.

I noticed that the `mypy` pydantic plugin does a lot to help the code understand the type coercion that pydantic does. `ty` doesn't seem to have this and I'm not sure they ever plan on having it tbh. Things like `Boolean(True), Boolean(False)`, etc instead of being able to use `True, False` and have pydantic convert to `Boolean` internally.

I'm not sure that this is necessarily a bad thing, it just makes things more verbose. For specifications, this seems totally reasonable so either way I am good with it - just making an observation here 😃. 

It's amazing how fast `ty` is 🚀... this lgtm 👍🏼

PR #238, opened 2025-12-18
feat: add signature aggregation using python bindings

anshalshukla:
## 🗒️ Description
Supports aggregation as per our discussion for devnet2.

The current implementation of aggregation bindings don't actually aggregate the signatures and instead return a single 0 byte, same for the verification API. This will be updated when we have aggregation support for current signature encoding scheme on TEST_CONFIG.

## ✅ Checklist
    ```console
    uvx tox
    ```


src/lean_spec/types/byte_arrays.py
```diff
@@ -383,3 +383,9 @@ class ByteList2048(BaseByteList):
     """Variable-length byte list with a limit of 2048 bytes."""
 
     LIMIT = 2048
+
+
+class ByteListMib(BaseByteList):
```
g11tech:
```suggestion
class LeanAggregatedSignature(BaseByteList):
```

src/lean_spec/subspecs/xmss/aggregation.py
```diff
@@ -0,0 +1,132 @@
...
+            _coerce_epoch(epoch),
+        )
+        return aggregated_bytes
+    except Exception as exc:
+        raise LeanMultisigAggregationError(f"lean-multisig aggregation failed: {exc}") from exc
+
+
+def verify_aggregated_payload(
+    public_keys: Sequence[PublicKey],
+    payload: bytes,
```
g11tech: is this just a byte sequence and not the ssz serialized signature list?

src/lean_spec/subspecs/xmss/aggregation.py
```diff
@@ -0,0 +1,132 @@
...
+    message: bytes,
```
g11tech: similar question as above, message here is serialized attestation data?

src/lean_spec/subspecs/xmss/aggregation.py
```diff
@@ -0,0 +1,132 @@
...
+def _coerce_epoch(epoch: int | Uint64) -> int:
+    value = int(epoch)
+    if value < 0 or value >= 2**32:
+        raise ValueError("epoch must fit in uint32 for lean-multisig aggregation")
+    return value
+
+
+def aggregate_signatures(
+    public_keys: Sequence[PublicKey],
+    signatures: Sequence[Signature],
```
g11tech: this is XmssSignature sequene? I don't think we have the recusive signature signing as of now?

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -458,27 +458,20 @@ def on_block(self, signed_block_with_attestation: SignedBlockWithAttestation) ->
...
-            assert len(validator_ids) == len(aggregated_signature), (
-                "Aggregated attestation signature count mismatch"
-            )
-
-            for validator_id, signature in zip(validator_ids, aggregated_signature, strict=True):
+            for validator_id in validator_ids:
+                # Signature bytes are not needed for forkchoice once the block-level
+                # aggregated proof has been verified.
                 store = store.on_attestation(
                     signed_attestation=SignedAttestation(
```
g11tech: we need to separate this import as `on_block_attestation` where we pass Attestation and not SignedAttestation, and that function will not verify the signature as signature already been verified in on_block

also latest know and latest new need to be track attestation data, and while building the block we need to lookup from attestation data and validator id to SignedAttestation for packing the block, and later on from SignedAggregatedAttesation when we can recursively pack it

so we need now two maps , 

validator_id, data => SignedAttestation object references,
validator_id, data => SignedAggregatedAttestation references

g11tech: ok so what we can do is, we can have `on_attestation` as what we have now, but with Attestation as input

and then have `on_gossip_attestation` which will take signedattestation from gossip to call `on_attestation` after verifying the xmss signature as well as populating the validator_id, data => SignedAttestation object references, map

while on_block populates the validator_id, data => SignedAggregatedAttestation references and calls `on_attestation` as well

so store.latest_known_attestations and store, latest_new_attestations are dicts from validator id => attestation data

and in proposal while looking to pack signatures, we do a validator_id, attestation data lookup in the two maps we have been manitaining

src/lean_spec/subspecs/containers/block/block.py
```diff
@@ -172,29 +171,33 @@ def verify_signatures(
...
-                    attestation_root,
-                    scheme,
-                ), "Attestation signature verification failed"
+
+            public_keys = [validators[vid].get_pubkey() for vid in validator_ids]
+            try:
+                verify_aggregated_payload(
+                    public_keys=public_keys,
+                    payload=bytes(aggregated_signature),
+                    message=attestation_root,
```
g11tech:
```suggestion
                    message=attestation_data_root,
```
lets rename this var it for more clarity

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -626,27 +627,34 @@ def build_block(
         proposer_index: Uint64,
         parent_root: Bytes32,
         attestations: list[Attestation] | None = None,
-        available_signed_attestations: Iterable[SignedAttestation] | None = None,
+        available_attestations: Iterable[Attestation] | None = None,
         known_block_roots: AbstractSet[Bytes32] | None = None,
+        gossip_attestation_signatures: dict[tuple[Uint64, bytes], "Signature"] | None = None,
+        block_attestation_signatures: dict[tuple[Uint64, bytes], list[LeanAggregatedSignature]] | None = None,
```
g11tech: we also need to store the corresponding aggregation bits as well with LeanAggregatedSignature in the dict because we can't recursively aggregate them (later) without knowing which indices voted for it which might be required as hint in recursive agg

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -677,19 +685,37 @@ def build_block(
...
+                # First check gossip signatures, then block signatures (first entry if available)
+                signature = None
+                if gossip_attestation_signatures:
+                    signature = gossip_attestation_signatures.get((validator_id, data_root))
+
+                # If not found in gossip and block signatures available, try those
+                # Note: For now, we only use block signatures if gossip signature not found
+                # TODO: Implement recursive aggregation to combine multiple aggregated signatures
+                if signature is None and block_attestation_signatures:
+                    aggregated_sigs = block_attestation_signatures.get((validator_id, data_root))
```
g11tech: we will also hhave aggregated bits as the result of lookup which will be used in the hint, so we return that as well

But everything rest looking good

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -119,21 +122,41 @@ class Store(Container):
...
+    """
+
+    gossip_attestation_signatures: Dict[tuple[Uint64, bytes], Signature] = {}
+    """
+    Map of validator id and attestation root to the XMSS signature.
+    """
+
+    block_attestation_signatures: Dict[
+        tuple[Uint64, bytes], list[tuple[AggregationBits, LeanAggregatedSignature]]
+    ] = {}
```
anshalshukla: Here I've stored hash root of attestation data instead of `AttestationData`, I think we can prune all the maps together and can get hash root for data to be pruned from `latest_known_attestations` and `latest_new_attestations`. Slots in those maps will help us know which keys are safe to prune.

We don't have pruning yet, it will be added later but I'm describing the rational why I think we should be able to prune it without knowing the slot.

cc: @g11tech

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -325,8 +335,9 @@ def _build_block_from_spec(
         )
 
         # Sign all attestations and the proposer attestation
-        attestation_signatures = key_manager.build_attestation_signatures(
-            final_block.body.attestations
+        attestation_signatures_blob = key_manager.build_attestation_signatures(
+            final_block.body.attestations,
+            signature_lookup=attestation_signatures,
```
unnawut: nit: for consistency with the 1st arg

```suggestion
            attestation_signatures,
```

src/lean_spec/subspecs/containers/block/block.py
```diff
@@ -177,24 +180,25 @@ def verify_signatures(
...
+                    epoch=aggregated_attestation.data.slot,
+                )
+            except LeanMultisigError as exc:
+                raise AssertionError(
+                    "Attestation aggregated signature verification failed"
+                ) from exc
```
unnawut: minor: Should we just let the original error propagate up? Not sure if we need to wrap it into a more generic AssertionError type and also suppress the original exception object

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -620,40 +620,153 @@ def state_transition(self, block: Block, valid_signatures: bool = True) -> "Stat
...
+    def _aggregate_signatures_from_gossip(
+        self,
+        validator_ids: list[Uint64],
+        data_root: bytes,
+        epoch: Slot,
+        gossip_attestation_signatures: dict[tuple[Uint64, bytes], "Signature"] | None = None,
+    ) -> LeanAggregatedSignature | None:
+        """Aggregate per-validator XMSS signatures into a single payload, if available."""
+        if not gossip_attestation_signatures or not validator_ids:
+            return None
```
unnawut: I think this currently always return None and the code below is never reached, because [the caller always passes 3 arguments](https://github.com/anshalshukla/leanSpec/blob/13f7387230ad3283f39eca3b9be120b1676609ca/src/lean_spec/subspecs/containers/state/state.py#L849-L851) and so gossip_attestation_signatures always default to None

anshalshukla: Missed the 4th argument, adding it in next commit.

src/lean_spec/subspecs/xmss/aggregation.py
anshalshukla: I think that is the case but anyway currently its in `test_mode` so actual aggregation is not taking place.

src/lean_spec/subspecs/xmss/aggregation.py
anshalshukla: It's not recursive signature signing it is calling aggregation for the list of Signatures, it calls leanMultisig python bindings here, but it is in test mode so return dummy 0 byte

src/lean_spec/subspecs/containers/block/block.py
```diff
@@ -177,24 +181,25 @@ def verify_signatures(
         ):
             validator_ids = aggregated_attestation.aggregation_bits.to_validator_indices()
 
-            assert len(aggregated_signature) == len(validator_ids), (
-                "Aggregated attestation signature count mismatch"
-            )
-
-            attestation_root = aggregated_attestation.data.data_root_bytes()
+            attestation_data_root = aggregated_attestation.data.data_root_bytes()
```
unnawut, line 183: Quite minor and I should've raised in the previous PR. But just want to flag this for quick opinion @tcoratger spec-writing wise...

Since `data_root_bytes()` is doing just `bytes(hash_tree_root(self))`, should we do this instead and remove `data_root_bytes()`?

```suggestion
            attestation_data_root = hash_tree_root(aggregated_attestation.data)
```

also the `bytes(...)` can be handled inside `verify_aggregated_payload()` so we don't lose type information early

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -466,15 +482,21 @@ def _build_signed_attestation_from_spec(
...
+        if spec.signature is not None:
+            signature = spec.signature
+        elif spec.valid_signature:
+            signature = key_manager.sign_attestation_data(
+                attestation.validator_id, attestation.data
+            )
+        else:
+            signature = Signature(
+                path=HashTreeOpening(siblings=HashDigestList(data=[])),
+                rho=Randomness(data=[Fp(0) for _ in range(PROD_CONFIG.RAND_LEN_FE)]),
```
tcoratger: I don't think that we should have any `PROD_CONFIG` in the codebase, we should use the env variable right?

packages/testing/src/consensus_testing/keys.py
```diff
@@ -273,24 +275,40 @@ def sign_attestation_data(
...
-        signature_lookup: Mapping[tuple[Uint64, bytes], Signature] | None = None,
+        signature_lookup: Mapping[AttestationSignatureKey, Signature] | None = None,
     ) -> AttestationSignatures:
-        """Build `AttestationSignatures` for already-aggregated attestations."""
+        """
+        Build `AttestationSignatures` for already-aggregated attestations.
+
+        For each aggregated attestation, collect the participating validators' public keys and
+        signatures, then produce a single leanVM aggregated signature proof blob using
+        `xmss_aggregate_signatures` (via `aggregate_signatures`).
```
tcoratger: These implementation details should change in the future (especially when we will not rely on the bindings anylire) so no need to add them here in the doc.

```suggestion
        For each aggregated attestation, collect the participating validators' public keys and
        signatures, then produce a single leanVM aggregated signature proof.
```

src/lean_spec/subspecs/containers/block/block.py
tcoratger: I'm open to both; I don't have a preference. It seems to me that this pattern is used a lot in the code, so perhaps having the utility is simpler and less verbose.

Of course, if there are only one or two uses across the entire codebase, that is useless from a spec point of view.

src/lean_spec/subspecs/containers/block/types.py
```diff
@@ -12,18 +16,51 @@ class AggregatedAttestations(SSZList[AggregatedAttestation]):
...
 
-    def has_duplicate_data(self) -> bool:
-        """Check if any two attestations share the same AttestationData."""
-        seen: set[AttestationData] = set()
-        for attestation in self:
-            if attestation.data in seen:
-                return True
-            seen.add(attestation.data)
-        return False
+    def each_duplicate_attestation_has_unique_participant(self) -> bool:
```
tcoratger: Here maybe we could have a less verbose name such as `validate_unique_contributions` ? Or maybe another idea?

src/lean_spec/subspecs/containers/block/types.py
```diff
@@ -12,18 +16,51 @@ class AggregatedAttestations(SSZList[AggregatedAttestation]):
...
+            for bits in bits_list:
+                unique = False
+                for i, bit in enumerate(bits.data):
+                    if bit and counts[i] == 1:
+                        unique = True
+                        break
+                if not unique:
+                    return False
+
+        return True
```
tcoratger: Something like this should be more pythonic and compact no?

```python
groups = defaultdict(list)
for att in self:
    groups[att.data.data_root_bytes()].append(att.aggregation_bits)

for bits_list in groups.values():
    if len(bits_list) <= 1:
        continue

    # 1. Convert bitfields to sets of active indices
    sets = [{i for i, bit in enumerate(bits.data) if bit} for bits in bits_list]

    # 2. Count index occurrences across the entire group
    counts = Counter(idx for s in sets for idx in s)

    # 3. Verify EVERY attestation has ANY index that appears EXACTLY once
    if not all(any(counts[i] == 1 for i in s) for s in sets):
        return False

return True
```

src/lean_spec/subspecs/containers/block/types.py
```diff
@@ -12,18 +16,51 @@ class AggregatedAttestations(SSZList[AggregatedAttestation]):
...
 
+class AttestationSignatures(SSZList[LeanAggregatedSignature]):
+    """
+    List of per-attestation aggregated signature proof blobs.
 
-class AttestationSignatures(SSZList[NaiveAggregatedSignature]):
-    """List of per-attestation naive signature lists aligned with block body attestations."""
+    Each entry corresponds to an aggregated attestation from the block body and contains
+    the raw bytes of the leanVM XMSSAggregatedSignature proof produced by
+    `xmss_aggregate_signatures`.
```
tcoratger: To avoid mentioning stuffs that can change in the future, could be better to have

```python
Each entry corresponds to an aggregated attestation from the block body and contains
    the raw bytes of the leanVM signature aggregation proof.
```

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -620,40 +621,109 @@ def state_transition(self, block: Block, valid_signatures: bool = True) -> "Stat
...
+            sigs.append(sig)
+            pks.append(self.validators[vid].get_pubkey())
+
+        aggregated_signature = aggregate_signatures(
+            public_keys=pks,
+            signatures=sigs,
+            message=data_root,
+            epoch=epoch,
+        )
+        return aggregated_signature
```
```suggestion
    def _aggregate_signatures_from_gossip(
        self,
        validator_ids: list[Uint64],
        data_root: bytes,
        epoch: Slot,
        gossip_signatures: dict[AttestationSignatureKey, "Signature"] | None = None,
    ) -> LeanAggregatedSignature | None:
        """Aggregate per-validator XMSS signatures into a single payload, if available."""
        if not gossip_signatures or not validator_ids:
            return None

        signatures: list[Signature] = []
        public_keys: list[bytes] = []

        for validator_index in validator_ids:
            # Attempt to retrieve the signature; fail fast if any are missing.
            if (sig := gossip_signatures.get((validator_index, data_root))) is None:
                return None

            signatures.append(sig)
            public_keys.append(self.validators[validator_index].get_pubkey())

        return aggregate_signatures(
            public_keys=public_keys,
            signatures=signatures,
            message=data_root,
            epoch=epoch,
        )
```

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -699,20 +767,233 @@ def build_block(
...
+                    continue
+
+                # We can only include an attestation if we have some way to later provide
+                # an aggregated payload for its group:
+                # - either a per validator XMSS signature from gossip, or
+                # - at least one aggregated payload learned from a block that references
+                #   this validator+data.
+                has_gossip_sig = bool(
+                    gossip_signatures and gossip_signatures.get(attestation_key) is not None
+                )
```
tcoratger: This should work no?

```suggestion
                has_gossip_sig = bool(
                    gossip_signatures and attestation_key in gossip_signatures
                )
```

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -699,20 +767,233 @@ def build_block(
...
+            }
+        )
 
         # Store the post state root in the block
-        final_block = candidate_block.model_copy(update={"state_root": hash_tree_root(post_state)})
+        final_block = final_block.model_copy(
+            update={
+                "state_root": hash_tree_root(post_state),
+            }
+        )
```
tcoratger, line 792: We should do a single model copy for all this no?

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -699,20 +767,233 @@ def build_block(
...
+            if aggregated_signature is not None:
+                final_aggregated_attestations.append(completely_aggregated_attestation)
+                final_aggregated_signatures.append(aggregated_signature)
+                continue
+
+            # If we have not found any aggregated signatures, split the attestations to cover
+            # all validators with minimal splits.
+            (
+                splited_aggregated_attestations,
+                splited_aggregated_signatures,
```
```suggestion
                splitted_aggregated_attestations,
                splitted_aggregated_signatures,
```

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -699,20 +767,233 @@ def build_block(
...
+                completely_aggregated_attestation,
+                gossip_signatures,
+                aggregated_payloads,
+            )
+            final_aggregated_attestations.extend(splited_aggregated_attestations)
+            final_aggregated_signatures.extend(splited_aggregated_signatures)
+
+        return final_aggregated_attestations, final_aggregated_signatures
+
+    def split_aggregated_attestations(
```
tcoratger: This function is too complex to be readable in a spec environment, but I'm in favor of merging asap and we can check later how to make things better.

src/lean_spec/subspecs/containers/state/types.py
```diff
@@ -1,11 +1,44 @@
...
+from ..block import Block  # noqa: E402
+
+BlockLookup = dict[Bytes32, Block]
+"""Mapping from block root to Block objects."""
+
+StateLookup = dict[Bytes32, "State"]
+"""Mapping from state root to State objects."""
+
+AttestationsByValidator = dict[Uint64, AttestationData]
+"""Mapping from validator index to attestation data."""
```
tcoratger: The fact that you have to put `# noqa: E402` and stuffs like this probably means that this is not the right place to define this. 

For example `BlockLookup` should be defined in the same file as `Block` no?

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -462,33 +553,47 @@ def on_block(
...
         aggregated_attestations = signed_block_with_attestation.message.block.body.attestations
         attestation_signatures = signed_block_with_attestation.signature.attestation_signatures
 
         assert len(aggregated_attestations) == len(attestation_signatures), (
             "Attestation signature groups must match aggregated attestations"
         )
 
+        # Copy the aggregated signature payloads map for updates
+        # Must deep copy the lists to maintain immutability of previous store snapshots
+        new_block_sigs = {k: list(v) for k, v in store.aggregated_payloads.items()}
```
```suggestion
        new_block_sigs = copy.deepcopy(store.aggregated_payloads)
```

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -462,33 +553,47 @@ def on_block(
...
                 store = store.on_attestation(
-                    signed_attestation=SignedAttestation(
+                    attestation=Attestation(
                         validator_id=validator_id,
-                        message=aggregated_attestation.data,
-                        signature=signature,
+                        data=attestation_data,
                     ),
                     is_from_block=True,
                 )
```
tcoratger: Something like this should be more compact and better

```python
for att, sig in zip(aggregated_attestations, attestation_signatures, strict=True):
            validator_ids = att.aggregation_bits.to_validator_indices()
            data_root = att.data.data_root_bytes()
            
            # Pre-calculate the payload record once for this entire group
            record = (AggregationBits.from_validator_indices(validator_ids), sig)

            for vid in validator_ids:
                # Update Signature Map
                #
                # Store the payload so future block builders can reuse this aggregation
                new_block_sigs.setdefault((vid, data_root), []).append(record)

                # Update Fork Choice
                #
                # Register the vote immediately (historical/on-chain)
                store = store.on_attestation(
                    attestation=Attestation(validator_id=vid, data=att.data),
                    is_from_block=True,
                )
```

src/lean_spec/subspecs/xmss/aggregation.py
```diff
@@ -0,0 +1,106 @@
...
+from lean_spec.types import Uint64
+from lean_spec.types.byte_arrays import LeanAggregatedSignature
+
+
+class LeanMultisigError(RuntimeError):
+    """Base exception for lean-multisig aggregation helpers."""
+
+
+class LeanMultisigAggregationError(LeanMultisigError):
+    """Raised when lean-multisig fails to aggregate or verify signatures."""
```
tcoratger: We should avoid the `Lean` prefix for an object because basically everything is lean since that is the leanspec. So `MultisigError` and `MultisigAggregationError` should be enough.

src/lean_spec/subspecs/xmss/aggregation.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+# This function will change for recursive aggregation
+# which might additionally require hints.
+def aggregate_signatures(
+    public_keys: Sequence[PublicKey],
+    signatures: Sequence[XmssSignature],
+    message: bytes,
+    epoch: Uint64,
+) -> LeanAggregatedSignature:
```
tcoratger: Here in the future we will want to have an object corresponding to the list of signatures (a wrapper for `Sequence[XmssSignature]`) and have this function implemented inside to avoid having an isolated function that we need to import each time we need to use it.

But that is fine to let it as is for this PR.

src/lean_spec/subspecs/xmss/aggregation.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+    Aggregate XMSS signatures using lean-multisig.
```
tcoratger:
```suggestion
    Aggregate XMSS signatures.
```

src/lean_spec/subspecs/xmss/aggregation.py
```diff
@@ -0,0 +1,106 @@
...
+    except Exception as exc:
+        raise LeanMultisigAggregationError(f"lean-multisig aggregation failed: {exc}") from exc
+
+
+# This function will change for recursive aggregation verification
+# which might additionally require hints.
+def verify_aggregated_payload(
+    public_keys: Sequence[PublicKey],
+    payload: LeanAggregatedSignature,
+    message: bytes,
```
tcoratger: Here we should move this method inside `LeanAggregatedSignature` as `verify` simply.

src/lean_spec/types/byte_arrays.py
```diff
@@ -386,3 +386,9 @@ class ByteList2048(BaseByteList):
     """Variable-length byte list with a limit of 2048 bytes."""
 
     LIMIT = 2048
+
+
+class LeanAggregatedSignature(BaseByteList):
+    """Variable-length byte list with a limit of 1048576 bytes."""
```
tcoratger, line 392: Where does this `1048576` come from? Production config or test config?

tests/consensus/devnet/verify_signatures/test_invalid_signatures.py
```diff
@@ -13,85 +13,86 @@
 
 pytestmark = pytest.mark.valid_until("Devnet")
 
+# TODO: Add these tests back when we have prod aggregation & verificationAPI
```
tcoratger: I'm not really in favor of removing tests since this could potentially let bugs slip through that we wouldn't detect. 

Instead, wouldn't it be better to just comment out a small check or only the problematic step while still allowing the rest of the test to run?

tests/lean_spec/subspecs/containers/test_state_aggregation.py
```diff
@@ -0,0 +1,750 @@
...
+
+
+def make_state(num_validators: int) -> State:
+    """Create a genesis state with the requested number of validators."""
+    return State.generate_genesis(Uint64(0), validators=make_validators(num_validators))
+
+
+def make_checkpoint(root: Bytes32, slot: int) -> Checkpoint:
+    """Helper to build checkpoints with integer slots."""
+    return Checkpoint(root=root, slot=Slot(slot))
```
tcoratger: I think that function is useless since we can call directly `Checkpoint(root=root, slot=Slot(slot))`

tests/lean_spec/subspecs/containers/test_state_aggregation.py
```diff
@@ -0,0 +1,750 @@
...
+        slot=Slot(slot),
+        head=make_checkpoint(head_root, slot),
+        target=make_checkpoint(target_root, slot),
+        source=source,
+    )
+
+
+def make_attestation(validator_index: int, data: AttestationData) -> Attestation:
+    """Create an attestation for the provided validator."""
+    return Attestation(validator_id=Uint64(validator_index), data=data)
```
tcoratger: Same I think that is useless

tests/lean_spec/subspecs/containers/test_state_aggregation.py
```diff
@@ -0,0 +1,750 @@
...
+
+    assert post_state.latest_block_header.slot == Slot(1)
+    assert aggregated_atts == []
+    assert aggregated_sigs == []
+    assert list(block.body.attestations.data) == []
+
+
+# ============================================================================
+# Additional edge case tests for _aggregate_signatures_from_gossip
+# ============================================================================
```
tcoratger: No need for this we can directly put the other tests without this comment.

```suggestion
```

tests/lean_spec/subspecs/containers/test_state_aggregation.py
```diff
@@ -0,0 +1,750 @@
...
+        Slot(1),
+        {},  # empty dict
+    )
+
+    assert result is None
+
+
+# ============================================================================
+# Additional edge case tests for _aggregate_signatures_from_block_payload
+# ============================================================================
```
tcoratger: Same

```suggestion
```

tests/lean_spec/subspecs/containers/test_state_aggregation.py
```diff
@@ -0,0 +1,750 @@
...
+        data_root,
+        aggregated_payloads,
+    )
+
+    assert result == full_payload
+
+
+# ============================================================================
+# Additional edge case tests for split_aggregated_attestations
+# ============================================================================
```
tcoratger:
```suggestion
```

tests/lean_spec/subspecs/containers/test_state_aggregation.py
```diff
@@ -0,0 +1,750 @@
...
+    participant_sets = [
+        {int(v) for v in att.aggregation_bits.to_validator_indices()} for att in split_atts
+    ]
+    # The large set should be selected
+    assert {0, 2, 3, 4} in participant_sets or {0, 1, 2, 3, 4} in participant_sets
+
+
+# ============================================================================
+# Additional edge case tests for compute_aggregated_signatures
+# ============================================================================
```
tcoratger:
```suggestion
```

src/lean_spec/subspecs/containers/block/types.py
g11tech: makes sense

src/lean_spec/subspecs/containers/block/types.py
g11tech: yes looks simple to read

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -699,20 +767,233 @@ def build_block(
...
+                #   this validator+data.
+                has_gossip_sig = bool(
+                    gossip_signatures and gossip_signatures.get(attestation_key) is not None
+                )
+
+                has_block_payload = bool(
+                    aggregated_payloads and aggregated_payloads.get(attestation_key)
+                )
+
+                if has_gossip_sig or has_block_payload:
```
g11tech, line 763: this condition should always be true, we just need to track if we have gossip sig for it or not

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -699,20 +767,233 @@ def build_block(
...
+        return final_block, post_state, aggregated_attestations, aggregated_signatures
+
+    def compute_aggregated_signatures(
+        self,
+        attestations: list[Attestation],
+        gossip_signatures: dict[AttestationSignatureKey, "Signature"] | None = None,
+        aggregated_payloads: dict[AttestationSignatureKey, AggregatedSignaturePayloads]
+        | None = None,
+    ) -> tuple[list[AggregatedAttestation], list[LeanAggregatedSignature]]:
+        """
```
g11tech, line 803: i think this function is better written as to give an aggregate signature of whatever validator indices it can find it gossip for that data root and also returning pending validator ids which can be used in the next segments.

since the validators are voting every slot, most of the signatures will be covered in this, and only small remaining validator ids might need to be covered from already aggregated signatures.

this is the aggregation flow we need :

1. `completely_aggregated_attestations = AggregatedAttestation.aggregate_by_data(attestations)`
2.  for completely_aggregated_attestation in completely_aggregated_attestations:
     i) gossip_aggregated_signature, gossip_aggregated_signature_bit_list, remaining_validator_ids = self._aggregate_signatures_from_gossip
     most of the times remaining_validator_ids should be almost empty or null and clients can choose to not further find aggregated attestations for the  `remaining_validator_ids` since in the next round of voting they will get the gossip signatures if those validators are on the same fork of the chain.
     ii) If client choose to go further, then from the aggregated signatures, one by one pick where aggregated signature with most remaining validators ids, will no remaining validator ids set is zero
       i.e. 
       ```python
        aggregates_signatures_list = [ ]
        aggregated_signature_bitlist_list = []
        # add previously aggregated signature from gossip into this list we are maintaining
        aggregates_signatures_list.append(gossip_aggregated_signature)
        aggregated_signature_bitlist_list.append(gossip_aggregated_signature_bit_list)
        
        # pick and collect other aggregated signatures to cover all the validator ids
        while(remaining_validator_ids != []):
          # we should be able to find a previously aggregated signature because a vote is only imported if we have recieved
             from gossip or from a previously aggregated signature
          aggregated_signature, aggregated_signature_bitlist,remaining_validators_ids = pick_from_aggregated_signatures(remaining_validator_ids)
          
          aggregates_signatures_list.append(aggregated_signature)
          aggregated_signature_bitlist_list.append(aggregated_signature_bitlist)
          
       # now we can aggregate the aggregated signature list recurively into just one aggregated signature
       # but since for now we can't so we will just create multiple aggregated attestations with the same data, but using the
       # signature and bit list from here
 ```
        ```

src/lean_spec/types/byte_arrays.py
anshalshukla: Its independent, I have used a safe upper limit of 1 MiB to store the proof blob, we can change it in future based on the actual proof blob size. Afaik the generated proof is not of fixed length.

tests/consensus/devnet/verify_signatures/test_invalid_signatures.py
anshalshukla: I think we can add back both the tests as we are checking it against xmss sig and not aggregated, will test and add back

src/lean_spec/subspecs/xmss/aggregation.py
anshalshukla: I've moved this inside a separate container and bit of refactoring to make both aggregation and verification methods of `MultisigAggregatedSignature`. I don't think we should make Sequence[XmssSignature] a separate container as it is being used just once but can be handled in followup PR if you don't feel the same

tests/consensus/devnet/verify_signatures/test_invalid_signatures.py
anshalshukla: Okay, `test_invalid_signature` I've added back because here the proposer xmss signature is invalid which is caught because we don't have any aggregation there. In `test_mixed_valid_invalid_signatures` we have dummy agggregation in place so we are unable to verify it's correctness. I've added a comment describing the same there.

unnawut (approved):
I want to spend a bit more time to understand `compute_aggregated_signatures()` and `split_aggregated_attestations()` and the tests, but like Thomas said these can go separate conversations.

anshalshukla:
I've removed `compute_aggregated_signatures()` and simplified `split_aggregated_attestaations()`. I've also removed `validate_unique_participants()`.

In the updated logic there is a possibility that there are some attestations that are do not contribute with any unique validator, this is fine as in the following devnet, we will recursively aggregate the attestations with same `data_root` so having a little bit of inefficiency for clarity is okay I believe. In the current form, implementing recursive aggregation will be easier as well.

Other changes are nicely described [here](https://github.com/leanEthereum/leanSpec/pull/238#discussion_r2649152335)

g11tech (approved):

tcoratger (approved):

PR #242, opened 2025-12-24
xmss: mv `KeyPair` as a container

tcoratger:
## 🗒️ Description

As we already have `KeyPair` as a class in the testing framework, I think that is cleaner to migrate it as a proper XMSS container so that we can reuse it as a return type of the key generation process for example.

## ✅ Checklist

    ```console
    uvx tox
    ```


unnawut (approved):
makes sense!

PR #245, opened 2025-12-29
aggregation: cleanup and reorg

tcoratger:
## 🗒️ Description

@anshalshukla I've tried some kind of cleanup, let me know what you think about it?

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/tests/lean_spec/subspecs/containers/test_state_aggregation.py
+++ b/tests/lean_spec/subspecs/containers/test_state_aggregation.py
@@ -117,7 +117,7 @@ def test_gossip_aggregation_succeeds_with_all_signatures() -> None:
 
     result = state._aggregate_signatures_from_gossip(
         validator_ids,
-        data_root,
+        Bytes32(data_root),
         Slot(3),
         gossip_signatures,
     )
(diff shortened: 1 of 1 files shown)


src/lean_spec/subspecs/xmss/aggregation.py
```diff
@@ -1,123 +1,128 @@
...
+
+    Used to index signature caches by (validator, message) pairs.
+    """
+
+    validator_id: Uint64
+    """The validator who produced the signature."""
 
-if TYPE_CHECKING:
-    from lean_spec.subspecs.containers.attestation import AggregationBits
+    data_root: Bytes32
```
anshalshukla, line 30: Why not just make it bytes? Right now you have to wrap root field in Bytes32 in every function call.

anshalshukla (approved):

src/lean_spec/subspecs/xmss/aggregation.py
tcoratger: Thanks for the observation, to have a stronger typing, I changed instead the return type of `data_root_bytes()` function to have `Bytes32` directly and so we don't need wrappers anymore :)

PR #250, opened 2025-12-31
signature aggregation: unify into a single function

tcoratger:
## 🗒️ Description

@anshalshukla I feel that unifying everything under a single function, rather than distributing it across subfunctions, is clearer and more uniform. Let me know what you think about this. 

Since this logic isn't finalized yet, I'd like to avoid multiplying the utility functions to make the specification readable and as compact as possible.

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/src/lean_spec/subspecs/containers/state/state.py
+++ b/src/lean_spec/subspecs/containers/state/state.py
@@ -912,6 +912,9 @@ class State(Container):
                     break
 
                 # Step 3: Record the proof and remove covered validators.
+                #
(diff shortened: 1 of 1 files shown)


anshalshukla:
I think otherwise, imo it's easier to read a function handling a specific task then to read a single function doing multiple tasks plus it is easier to write tests as well.

unnawut:
I feel quite indifferent about this one so yes I guess we can close for now

Issue #243, opened 2025-12-27
poseidon2 not consistent between python and rust


GrapeBaBa:
```rust
#[test]
    fn test_poseidon2_consistency() {
        // Test that our Poseidon2 output matches Python's output
        // Using test vectors from leanSpec tests/lean_spec/subspecs/poseidon2/test_permutation.py
        use p3_koala_bear::{KoalaBear, default_koalabear_poseidon2_16};
        use p3_symmetric::Permutation;
        use p3_field::{PrimeCharacteristicRing, PrimeField32};
        
        let perm = default_koalabear_poseidon2_16();
        
        // Input from leanSpec test_permutation.py INPUT_16:
        let input_vals: [u32; 16] = [
            894848333, 1437655012, 1200606629, 1690012884,
            71131202, 1749206695, 1717947831, 120589055,
            19776022, 42382981, 1831865506, 724844064,
            171220207, 1299207443, 227047920, 1783754913,
        ];
        
        let mut state: [KoalaBear; 16] = core::array::from_fn(|i| KoalaBear::from_u64(input_vals[i] as u64));
        
        println!("Input: {:?}", state.map(|x| x.as_canonical_u32()));
        
        perm.permute_mut(&mut state);
        
        let output: Vec<u32> = state.iter().map(|x| x.as_canonical_u32()).collect();
        println!("Output: {:?}", output);
        
        // Expected from leanSpec test_permutation.py EXPECTED_16:
        let expected: [u32; 16] = [
            1934285469, 604889435, 133449501, 1026180808,
            1830659359, 176667110, 1391183747, 351743874,
            1238264085, 1292768839, 2023573270, 1201586780,
            1360691759, 1230682461, 748270449, 651545025,
        ];
        
        for (i, (got, exp)) in output.iter().zip(expected.iter()).enumerate() {
            if got != exp {
                println!("Mismatch at index {}: got {}, expected {}", i, got, exp);
            }
        }
        
        assert_eq!(output.as_slice(), &expected[..], "Poseidon2 output mismatch with leanSpec vectors!");
        println!("Poseidon2 consistency test passed!");
    }
```

Rust output: 
```
[190453639, 458899855, 383789123, 1958965770, 1470307143, 135446903, 1980271247, 26609194, 337889870, 543343594, 900082402, 1267415354, 1018710090, 902823573, 1161524658, 1483653556]
```

PR #251, opened 2026-01-02
Poseidon2: reorder constants as per Plonky3

tcoratger:
## 🗒️ Description

@GrapeBaBa I think we had an ordering problem when compared to Plonky3, please check this, this should solve your problem without any leansig modification.

## 🔗 Related Issues or PRs

Should close https://github.com/leanEthereum/leanSpec/issues/243

## ✅ Checklist

    ```console
    uvx tox
    ```


tcoratger:
> @tcoratger This requires key re-generation right? So I can do so for https://github.com/leanEthereum/leansig-test-keys for prod keys as well

Yes correct, please @GrapeBaBa confirm and then we can regenerate the keys

GrapeBaBa:
Works fine

PR #253, opened 2026-01-04
networking: implement enr

tcoratger:
## 🗒️ Description

This is a first version of the implementation based on https://eips.ethereum.org/EIPS/eip-778 and ream code. 

We can iterate on this later if we need to remove, add or modify things.

## 🔗 Related Issues or PRs

Followup on top of https://github.com/leanEthereum/leanSpec/pull/252

## ✅ Checklist

    ```console
    uvx tox
    ```


unnawut:
Only one small question, should we make a stronger distinction between the specs vs. supporting implementation? Maybe moving enr into support dir e.g. `src/lean_spec/subspecs/networking/support/enr` or even `src/support/networking/enr`?

E.g. `./reqresp` (that contains `BlocksByRootRequest` and `BlocksByRootResponse`) is part of lean spec that clients need to align on. But `./enr` (that contains `ENR`, `Eth2Data`, etc.) is not the spec that clients need to follow.

Another example is `subspecs/containers` vs. `subspecs/ssz`, but it's quite clear that containers are the specs, and ssz is the supporting impl.

It'll also help with spec diffing in the future. But we can also worry about all this later

PR #258, opened 2026-01-07
fix: fix process attestation source/target check bug

GrapeBaBa:
## 🗒️ Description

This pull request makes a small fix to the logic in the `process_attestations` function in `src/lean_spec/subspecs/containers/state/state.py`. The change corrects the conditional statement to ensure that votes about unknown or conflicting forks are properly filtered.

* Updated the conditional logic in `process_attestations` to use `or` instead of `and`, preventing votes about unknown or conflicting forks more accurately

## ✅ Checklist

    ```console
    uvx tox
    ```


tcoratger (approved):
Thanks a lot!

PR #273, opened 2026-01-12
fix: fix justifications unpacking/repacking logic to handle finalized slot boundaries

GrapeBaBa:
## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/src/lean_spec/subspecs/containers/state/state.py
+++ b/src/lean_spec/subspecs/containers/state/state.py
@@ -298,9 +298,14 @@ class State(Container):
         # latest finalized checkpoint.
         #
         # Here, we extend the storage capacity to ensure the range from the
-        # finalized boundary up to the current block's slot is fully tracked
-        # and addressable.
-        new_justified_slots_data = self.justified_slots.extend_to_slot(self.latest_finalized.slot, block.slot)
+        # finalized boundary up to the last materialized slot is fully tracked
+        # and addressable. The current block's slot is not materialized until
+        # its header is fully processed, so we stop at slot (block.slot - 1).
+        last_materialized_slot = block.slot - Slot(1)
+        new_justified_slots_data = self.justified_slots.extend_to_slot(
+            self.latest_finalized.slot,
+            last_materialized_slot,
+        )
 
         # Construct the new latest block header.
         #
@@ -417,7 +422,12 @@ class State(Container):
         justified_slots = self.justified_slots
 
         # Map roots to their slots for pruning when finalization advances.
-        root_to_slot = {root: Slot(i) for i, root in enumerate(self.historical_block_hashes)}
+        # Only track roots after the finalized boundary; earlier roots are pruned.
+        start_slot = int(finalized_slot) + 1
+        root_to_slot = {
+            self.historical_block_hashes[i]: Slot(i)
+            for i in range(start_slot, len(self.historical_block_hashes))
+        }
 
         # Process each attestation independently
         #
@@ -510,7 +520,11 @@ class State(Container):
                 #
                 # The chain now considers this block part of its safe head.
                 latest_justified = target
-                justified_slots.set_justified(finalized_slot, target.slot, True)
+                justified_slots = justified_slots.with_justified(
+                    finalized_slot,
+                    target.slot,
+                    True,
+                )
 
                 # There is no longer any need to track individual votes for this block.
                 del justifications[target.root]
--- a/src/lean_spec/subspecs/containers/state/types.py
+++ b/src/lean_spec/subspecs/containers/state/types.py
@@ -54,7 +54,7 @@ class JustifiedSlots(BaseBitlist):
 
         # Check the tracked bitfield for the slot's status.
         #
-        # We assume the slot is within the tracked range. 
+        # We assume the slot is within the tracked range.
         #
         # If the caller asks for a slot too far in the future, it indicates a logic error.
         try:
@@ -65,7 +65,12 @@ class JustifiedSlots(BaseBitlist):
                 f"(finalized_boundary={finalized_slot}, tracked_length={len(self)})"
             ) from e
 
-    def with_justified(self, finalized_slot: Slot, target_slot: Slot, value: bool | Boolean) -> None:
+    def with_justified(
+        self,
+        finalized_slot: Slot,
+        target_slot: Slot,
+        value: bool | Boolean,
+    ) -> "JustifiedSlots":
         """
         Return a new bitfield with the justification status updated.
 
@@ -107,7 +112,7 @@ class JustifiedSlots(BaseBitlist):
         # 3. Use model_copy to return a new instance with the updated data.
         new_data = list(self.data)
         new_data[relative_index] = value
-        
+
         return self.model_copy(update={"data": new_data})
 
     def extend_to_slot(self, finalized_slot: Slot, target_slot: Slot) -> JustifiedSlots:


src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -70,6 +70,58 @@ class State(Container):
...
+    def _justified_slots_index(finalized_slot: Slot, slot: Slot) -> int | None:
+        """Return index into `justified_slots` for `slot`, or None if implicitly justified.
+
+        The spec stores `justified_slots` starting at `finalized_slot + 1`.
+        Any slot <= finalized_slot is treated as justified (finalized history).
+        """
+        if slot <= finalized_slot:
+            return None
+        base = finalized_slot + Slot(1)
+        return int(slot - base)
```
tcoratger: I recommend implementing this directly into `Slot` to avoid overloading the state as something like:

```python
    def justified_index_after(self, finalized_slot: Slot) -> int | None:
        """
        Return the bitfield index of this slot relative to a finalized slot.

        This index is used to look up justification status in the `justified_slots`
        bitfield. The bitfield starts tracking at `finalized_slot + 1`.

        Logic:
          - If self <= finalized_slot: It is implicitly justified (part of history).
            Returns None.
          - If self > finalized_slot: It maps to index (self - finalized_slot - 1).
            Example: finalized=100.
                     slot=101 -> index 0
                     slot=102 -> index 1

        Args:
            finalized_slot: The anchor slot representing the finalized history.

        Returns:
            The 0-based index for the bitfield, or None if implicitly justified.
        """
        if self <= finalized_slot:
            return None
        
        # Calculate distance and adjust for 0-based indexing starting at finalized+1
        return int(self - finalized_slot) - 1
```

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -70,6 +70,58 @@ class State(Container):
...
+        idx = cls._justified_slots_index(finalized_slot, slot)
+        if idx is None:
+            return True
+        if idx >= len(justified_slots):
+            raise IndexError(
+                "Justified slot index out of bounds "
+                f"(idx={idx}, len={len(justified_slots)}, "
+                f"slot={slot}, finalized_slot={finalized_slot})"
+            )
+        return bool(justified_slots[idx])
```
tcoratger: Again not to overload the state, I think that it makes sense to move this to `JustifiedSlots` implementation as:

```python
def is_slot_justified(self, finalized_slot: Slot, target_slot: Slot) -> bool:
        """
        Check if a target slot is justified.
        
        This handles the logic of:
        1. Implicit justification (slots <= finalized).
        2. Relative indexing into the bitfield.
        3. Bounds checking.
        """
        # Use the new method on Slot to get the relative index
        idx = target_slot.justified_index_after(finalized_slot)

        # Case 1: The slot is part of finalized history (implicitly justified)
        if idx is None:
            return True

        # Case 2: The slot is in the future relative to our tracked bits
        if idx >= len(self):
            raise IndexError(
                f"Slot {target_slot} is out of bounds for justification tracking "
                f"(max trackable slot: {finalized_slot + Slot(len(self))})"
            )

        # Case 3: Check the specific bit in our collection
        return bool(self[idx])
```

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -70,6 +70,58 @@ class State(Container):
...
+        if idx is None:
+            # Slots at or before finalization are implicitly justified.
+            return
+        if idx >= len(justified_slots):
+            raise IndexError(
+                "Justified slot index out of bounds "
+                f"(idx={idx}, len={len(justified_slots)}, "
+                f"slot={slot}, finalized_slot={finalized_slot})"
+            )
+        justified_slots[idx] = Boolean(value)
```
tcoratger: I think that you can move this as well to `JustifiedSlots` as

```python
def set_justified(self, finalized_slot: Slot, target_slot: Slot, value: bool | Boolean) -> None:
        """
        Set the justified status for a target slot.
        
        This handles the relative indexing logic internally.
        
        Args:
            finalized_slot: The anchor slot representing finalized history.
            target_slot: The slot to update.
            value: The boolean status to set.
        """
        # 1. Calculate the relative index
        idx = target_slot.justified_index_after(finalized_slot)

        # 2. Handle implicit justification (No-op)
        # If the slot is already finalized, we typically cannot "un-justify" it,
        # and it is implicitly justified anyway. We effectively ignore writes here.
        if idx is None:
            return

        # 3. Check bounds
        if idx >= len(self):
            raise IndexError(
                f"Cannot set justification for slot {target_slot}: index {idx} "
                f"out of bounds (len={len(self)})"
            )

        # 4. Update the bit
        # We explicitly cast to Boolean to ensure SSZ type safety if required
        self[idx] = Boolean(value)
```

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -292,26 +344,25 @@ def process_block_header(self, block: Block) -> "State":
...
+
+        new_bits = list(self.justified_slots.data)
+        for slot_i in range(int(next_slot), int(block.slot)):
+            slot = Slot(slot_i)
+            if slot == parent_header.slot:
+                new_bits.append(Boolean(is_genesis_parent))
+            else:
+                new_bits.append(Boolean(False))
+
+        new_justified_slots_data = JustifiedSlots(data=new_bits)
```
tcoratger: Thanks a lot for finding the bug, maybe for clarity something like this would be better no?

```suggestion
        # 1. Calculate the slot where our current list ends (the "frontier").
        #    Formula: finalized + 1 (start offset) + length
        frontier_slot = finalized_slot + Slot(1) + Slot(len(self))

        # 2. Calculate how many empty slots exist between the frontier and the target.
        #    If target is 100 and frontier is 98, we need to fill slots 98 and 99.
        #    Gap = 100 - 98 = 2.
        gap_size = int(target_slot - frontier_slot)

        # 3. If no gap, return self unchanged.
        if gap_size <= 0:
            return self

        # 4. Fill the gap with False (unjustified) entries.
        return JustifiedSlots(data=self.data + [Boolean(False)] * gap_size)
```

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -537,7 +592,24 @@ def process_attestations(
...
+                    #
+                    # We also prune any pending `justifications` whose slots are now finalized.
+                    delta = int(finalized_slot - old_finalized_slot)
+                    if delta > 0:
+                        justified_slots = JustifiedSlots(data=justified_slots.data[delta:])
+                        justifications = {
+                            root: votes
+                            for root, votes in justifications.items()
+                            if root_to_slot.get(root, Slot(0)) > finalized_slot
+                        }
```
tcoratger, line 564: Maybe we could add to `JustifiedSlots` this

```python
def shift_window(self, delta: int) -> "JustifiedSlots":
        """
        Advance the tracking window by `delta` slots.
        
        Removes the first `delta` entries because they are now part of the 
        finalized history (implicitly justified).
        """
        if delta <= 0:
            return self
        
        # Return a new instance with the sliced data
        return JustifiedSlots(data=self.data[delta:])
```

and then just have this no?

```python
delta = int(finalized_slot - old_finalized_slot)
                    
                    if delta > 0:
                        # Tell the bitfield to shift its window
                        justified_slots = justified_slots.shift_window(delta)

                        # Prune pending votes for blocks that are now finalized.
                        #    We keep only the roots that are still in the future relative to finalization.
                        justifications = {
                            root: votes
                            for root, votes in justifications.items()
                            if root_to_slot.get(root, Slot(0)) > finalized_slot
                        }
```

GrapeBaBa:
> Just left couple of small comments.
> 
> Thanks for finding the bug. Any reason why this draft PR is not converted to a real PR? I think this makes sense to merge is as soon as possible no?

Thanks @tcoratger , I was going to let @g11tech pre review it first. I will fix the comments and make it ready to review soon.

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -292,26 +292,14 @@ def process_block_header(self, block: Block) -> "State":
...
-        # 3. The skipped slots status (multiple entries):
-        #    We append False for every empty slot between the parent and the
-        #    current block. Since no blocks exist there, they are permanently
-        #    unjustified.
-        new_justified_slots_data = (
-            self.justified_slots + [Boolean(is_genesis_parent)] + [Boolean(False)] * num_empty_slots
-        )
+        # Here we append flags for every newly materialized slot in:
+        #   [parent_header.slot, block.slot)
+        # but only for slots >= (latest_finalized.slot + 1).
```
tcoratger: I think copilot is right here, I propose something like in plain english to avoid depending on variable names that can change in the future:

```python
# Update the list of justified slot flags.
#
# IMPORTANT: This list is stored relative to the finalized boundary.
#
# The first entry corresponds to the slot immediately following the
# latest finalized checkpoint.
#
# Here, we extend the storage capacity to ensure the range from the
# finalized boundary up to the current block's slot is fully tracked
# and addressable.
```

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -292,26 +292,14 @@ def process_block_header(self, block: Block) -> "State":
...
+        finalized_slot = self.latest_finalized.slot
+        new_justified_slots_data = self.justified_slots.extend_to_slot(finalized_slot, block.slot)
```
```suggestion
        new_justified_slots_data = self.justified_slots.extend_to_slot(self.latest_finalized.slot, block.slot)
```

src/lean_spec/subspecs/containers/state/types.py
```diff
@@ -26,6 +27,66 @@ class JustifiedSlots(BaseBitlist):
...
+        if idx is None:
+            return True
+
+        if idx >= len(self):
+            raise IndexError(
+                "Justified slot index out of bounds "
+                f"(idx={idx}, len={len(self)}, slot={target_slot}, finalized_slot={finalized_slot})"
+            )
+
+        return bool(self[idx])
```
tcoratger:
```suggestion
        """
        Determine if a specific slot is considered justified.

        The check follows these rules:
        - Slots at or before the finalized boundary are implicitly justified.
        - Future slots are checked against the tracked bitfield.

        Args:
            finalized_slot: The anchor point for the tracking window.
            target_slot: The slot to query.

        Returns:
            True if the slot is justified or finalized, False otherwise.

        Raises:
            IndexError: If the target slot is active but outside the tracked range.
        """
        # First, determine the position of the target relative to the anchor.
        #
        # If the result is None, the slot is behind the finalized boundary.
        # By definition, finalized slots are justified.
        if (relative_index := target_slot.justified_index_after(finalized_slot)) is None:
            return True

        # Check the tracked bitfield for the slot's status.
        #
        # We assume the slot is within the tracked range. 
        #
        # If the caller asks for a slot too far in the future, it indicates a logic error.
        try:
            return bool(self[relative_index])
        except IndexError as e:
            raise IndexError(
                f"Slot {target_slot} is outside the tracked range "
                f"(finalized_boundary={finalized_slot}, tracked_length={len(self)})"
            ) from e
```

src/lean_spec/subspecs/containers/state/types.py
```diff
@@ -26,6 +27,66 @@ class JustifiedSlots(BaseBitlist):
...
+    def shift_window(self, delta: int) -> JustifiedSlots:
+        """
+        Advance the tracking window by dropping slots that became finalized.
+
+        A non-positive delta keeps the tracking window unchanged.
+        """
+        if delta <= 0:
+            return self
+
+        return JustifiedSlots(data=self.data[delta:])
```
tcoratger:
```suggestion
        # If the boundary hasn't moved forward, the window stays the same.
        if delta <= 0:
            return self

        # Return a new instance containing only the relevant subset of data.
        return self.model_copy(update={"data": self.data[delta:]})
```

src/lean_spec/subspecs/containers/state/types.py
```diff
@@ -26,6 +27,66 @@ class JustifiedSlots(BaseBitlist):
...
+        if idx is None:
+            return
+
+        if idx >= len(self):
+            raise IndexError(
+                "Justified slot index out of bounds "
+                f"(idx={idx}, len={len(self)}, slot={target_slot}, finalized_slot={finalized_slot})"
+            )
+
+        self[idx] = value
```
tcoratger: Here I think that copilot is right, we want to preserve immutability to be friendly with formal verification for example, so I propose something like:

```suggestion
        """
        Return a new bitfield with the justification status updated.

        This method follows the immutable pattern:
        - Returns 'self' if the slot is finalized (immutable).
        - Returns a clone with the specific bit updated for active slots.

        Args:
            finalized_slot: The anchor point for the tracking window.
            target_slot: The slot to update.
            value: The new justification status.

        Returns:
            A new, updated JustifiedSlots instance.

        Raises:
            IndexError: If the target slot is active but outside the tracked range.
        """
        # Determine the position of the target relative to the anchor.
        #
        # If the slot is behind the finalized boundary, we return 'self' unchanged.
        # We cannot modify the status of finalized history, and treating it as a
        # no-op preserves the immutability of the conceptual chain history.
        if (relative_index := target_slot.justified_index_after(finalized_slot)) is None:
            return self

        # Ensure we are not trying to write to a future slot that does not exist
        # in our tracking list yet. The state must be explicitly extended first.
        if relative_index >= len(self):
            raise IndexError(
                f"Slot {target_slot} is outside the tracked range "
                f"(finalized_boundary={finalized_slot}, tracked_length={len(self)})"
            )

        # Clone and update in one smooth operation.
        #
        # 1. Create a shallow copy of the data list to avoid mutating the original.
        # 2. Update the specific bit in the copy.
        # 3. Use model_copy to return a new instance with the updated data.
        new_data = list(self.data)
        new_data[relative_index] = value
        
        return self.model_copy(update={"data": new_data})
```

src/lean_spec/subspecs/containers/state/types.py
```diff
@@ -26,6 +27,66 @@ class JustifiedSlots(BaseBitlist):
...
+
+        if idx >= len(self):
+            raise IndexError(
+                "Justified slot index out of bounds "
+                f"(idx={idx}, len={len(self)}, slot={target_slot}, finalized_slot={finalized_slot})"
+            )
+
+        return bool(self[idx])
+
+    def set_justified(self, finalized_slot: Slot, target_slot: Slot, value: bool | Boolean) -> None:
```
tcoratger: I would probably rename this function to `with_justified` since we use immutability.

src/lean_spec/subspecs/containers/state/types.py
```diff
@@ -26,6 +27,66 @@ class JustifiedSlots(BaseBitlist):
...
+        Extend tracking so that slots below the target are addressable.
+
+        Newly created entries are initialized to False.
+        """
+        frontier_slot = finalized_slot + Slot(1) + Slot(len(self))
+        gap_size = int(target_slot - frontier_slot)
+        if gap_size <= 0:
+            return self
+
+        return JustifiedSlots(data=list(self.data) + [Boolean(False)] * gap_size)
```
tcoratger: I think we should use model copy again here with something like this:

```suggestion
        """
        Extend the tracking capacity to cover a new target slot.

        This prepares the state to process a new block by ensuring the
        bitfield is long enough to store its justification status.
        Gaps are filled with False (unjustified).

        Args:
            finalized_slot: The anchor point for the tracking window.
            target_slot: The slot that must be addressable.

        Returns:
            A new instance with sufficient capacity.
        """
        # Calculate the index required to store the status of the target.
        #
        # If the target is already finalized, no extension is needed because
        # we don't track finalized data.
        if (relative_index := target_slot.justified_index_after(finalized_slot)) is None:
            return self

        # Calculate how many new entries we need to append.
        #
        # Since indices are zero-based, the required capacity is index + 1.
        # If we already have enough capacity, the gap will be zero or negative.
        required_capacity = relative_index + 1
        if (gap_size := required_capacity - len(self)) <= 0:
            return self

        # Return a new instance with the extended data list.
        #
        # We extend the existing data with False values to bridge the gap.
        return self.model_copy(
            update={"data": list(self.data) + [Boolean(False)] * gap_size}
        )
```

tcoratger (commented):
Overall this looks good, just left couple of suggestions for your consideration but it's not a blocker for me for merging except the fact that we should use immutability and fix the comments of copilot.

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -424,8 +412,12 @@ def process_attestations(
         # Track state changes to be applied at the end
         latest_justified = self.latest_justified
         latest_finalized = self.latest_finalized
+        finalized_slot = latest_finalized.slot
         justified_slots = self.justified_slots
 
+        # Map roots to their slots for pruning when finalization advances.
+        root_to_slot = {root: Slot(i) for i, root in enumerate(self.historical_block_hashes)}
```
g11tech: we don't need to create this map from 0 to the end, only from the `finalized_slot+1` to `self.historical_block_hashes.len` because those roots are anyway not in justifications maps as we prune them out

g11tech: this will also keep this map compressed otherwise the benefit of this PR goes away

tcoratger (approved):

PR #279, opened 2026-01-13
checkpoint sync API specification

ch4r10t33r:
## 🗒️ Description

- Add HTTP API server for checkpoint sync, matching zeam's implementation
- New /lean/states/finalized endpoint serves finalized state as SSZ binary for fast node bootstrapping
- New /health endpoint for service health checks
- Add client utilities for fetching and verifying finalized state from remote nodes
- 
## ✅ Checklist

    ```console
    uvx tox
    ```


src/lean_spec/subspecs/api/client.py
```diff
@@ -0,0 +1,135 @@
+"""
+Checkpoint sync client for downloading finalized state from another node.
+
+This client is used for fast synchronization - instead of syncing from genesis,
+a node can download the finalized state from a trusted peer and start from there.
+
+This matches the checkpoint sync client implemented in zeam.
```
tcoratger: As this is the reference spec, I guess this is better to avoid mentioning any other client.

```suggestion
```

g11tech: +1

src/lean_spec/subspecs/api/client.py
```diff
@@ -0,0 +1,135 @@
...
+        # The latest_block_header.state_root should be zero for the current state
+        # (it gets filled in after the state transition)
+        # We verify the structure is valid by checking other fields
+
+        # Basic sanity checks
+        if int(state.slot) < 0:
+            logger.error("Invalid state: negative slot")
+            return False
+
+        if len(state.validators) == 0:
```
g11tech: this could be matched with config

src/lean_spec/subspecs/api/client.py
```diff
@@ -0,0 +1,135 @@
...
+        url: Base URL of the node API (e.g., "http://localhost:5052")
+        state_class: The State class to deserialize into
+
+    Returns:
+        The finalized State object
+
+    Raises:
+        CheckpointSyncError: If the request fails or state is invalid
+    """
+    import asyncio
```
tcoratger: The import should be at the top of the file no?

src/lean_spec/subspecs/api/client.py
```diff
@@ -0,0 +1,135 @@
...
+
+    Performs basic validation checks on the downloaded state.
+
+    Args:
+        state: The state to verify
+
+    Returns:
+        True if valid, False otherwise
+    """
+    from lean_spec.subspecs.ssz.hash import hash_tree_root
```
tcoratger: Import on the top of the file?

src/lean_spec/subspecs/api/client.py
tcoratger, line 1: Can't we use the httpx library here to simplify the file a lot and avoid the manual `if "200" not in status_line` for example. I asked ChatGpt (I haven't test the code) and it gives something like this, could it work?

```python
"""
Checkpoint sync client for downloading finalized state from another node.

This client is used for fast synchronization - instead of syncing from genesis,
a node can download the finalized state from a trusted peer and start from there.
"""

from __future__ import annotations

import logging
from typing import TYPE_CHECKING, Any, TypeVar, Type

# We use httpx for robust async HTTP handling (supports HTTP/1.1, HTTP/2, TLS)
import httpx

if TYPE_CHECKING:
    from lean_spec.subspecs.containers import State

# Define a TypeVar to allow the state_class to return the specific type of State
S = TypeVar("S", bound="State")

logger = logging.getLogger(__name__)

# Constants
DEFAULT_TIMEOUT = 60.0  # Seconds
FINALIZED_STATE_ENDPOINT = "/lean/states/finalized"


class CheckpointSyncError(Exception):
    """Base error for checkpoint sync failures."""
    pass


class NetworkError(CheckpointSyncError):
    """Raised when the network request fails."""
    pass


class ValidationError(CheckpointSyncError):
    """Raised when the state is downloaded but invalid."""
    pass


async def fetch_finalized_state(url: str, state_class: Type[S]) -> S:
    """
    Fetch finalized state from a node via checkpoint sync.

    Downloads the finalized state as SSZ binary and deserializes it.

    Args:
        url: Base URL of the node API (e.g., "http://localhost:5052")
        state_class: The State class to deserialize into

    Returns:
        The finalized State object

    Raises:
        CheckpointSyncError: If the request fails or state is invalid
    """
    # Clean up URL to ensure we hit the correct endpoint
    base_url = url.rstrip("/")
    full_url = f"{base_url}{FINALIZED_STATE_ENDPOINT}"

    logger.info(f"Fetching finalized state from {full_url}")

    headers = {
        "Accept": "application/octet-stream",  # Explicitly request binary SSZ
        "User-Agent": "lean-client/checkpoint-sync",
    }

    try:
        async with httpx.AsyncClient(timeout=DEFAULT_TIMEOUT) as client:
            response = await client.get(full_url, headers=headers)
            
            # Raise error for 4xx or 5xx status codes
            response.raise_for_status()
            
            # Read binary data
            ssz_data = response.content
            logger.info(f"Downloaded {len(ssz_data)} bytes of SSZ state data")

            # Deserialize SSZ
            # Assuming decode_bytes is a synchronous CPU-bound method. 
            # If it's heavy, it might block the event loop, but usually acceptable for one-off sync.
            state = state_class.decode_bytes(ssz_data)

            logger.info(f"Deserialized state at slot {state.slot}")
            return state

    except httpx.RequestError as exc:
        raise NetworkError(f"Network error while connecting to {exc.request.url}: {exc}") from exc
    except httpx.HTTPStatusError as exc:
        raise NetworkError(f"HTTP error {exc.response.status_code}: {exc.response.text[:200]}") from exc
    except Exception as e:
        raise CheckpointSyncError(f"Failed to deserialize or process state: {e}") from e


async def verify_checkpoint_state(state: "State") -> bool:
    """
    Verify that a checkpoint state is valid.

    Performs basic sanity checks on the downloaded state.

    Args:
        state: The state to verify

    Returns:
        True if valid, False otherwise
    """
    from lean_spec.subspecs.ssz.hash import hash_tree_root

    try:
        # 1. Structure Checks
        if state.slot < 0:
            raise ValidationError(f"Invalid state: negative slot {state.slot}")

        if not state.validators:
            raise ValidationError("Invalid state: validator set is empty")

        # 2. Cryptographic Check
        # Compute the tree root to ensure the object is structurally sound and hashable
        computed_root = hash_tree_root(state)
        root_hex = computed_root.hex()

        logger.info(
            f"Checkpoint state verified:\n"
            f"  Slot: {state.slot}\n"
            f"  Validators: {len(state.validators)}\n"
            f"  State Root: 0x{root_hex[:12]}..."
        )
        return True

    except ValidationError as e:
        logger.error(f"State validation failed: {e}")
        return False
    except Exception as e:
        logger.error(f"Unexpected error during verification: {e}")
        return False
```

src/lean_spec/subspecs/api/server.py
tcoratger, line 1: Same here, I think that we reimplement the http protocol a bit too deep and we can simply reuse for this aiohttp as an asynchronous HTTP client/server. Again LLM gave me a simplified version like this that should work with less code:

```python
"""
API server for checkpoint sync and node status endpoints.

Provides HTTP endpoints for:
- /lean/states/finalized - Serve finalized checkpoint state as SSZ
- /health - Health check endpoint
"""

from __future__ import annotations

import asyncio
import logging
from collections.abc import Callable
from dataclasses import dataclass, field
from typing import TYPE_CHECKING, Optional

# Use aiohttp for a robust, production-grade async server
from aiohttp import web

if TYPE_CHECKING:
    from lean_spec.subspecs.forkchoice import Store

logger = logging.getLogger(__name__)


@dataclass(frozen=True, slots=True)
class ApiServerConfig:
    """Configuration for the API server."""
    host: str = "0.0.0.0"
    port: int = 5052
    enabled: bool = True


class ApiServer:
    """
    HTTP API server for checkpoint sync and node status.
    
    Uses aiohttp to handle HTTP protocol details efficiently.
    """

    def __init__(
        self, 
        config: ApiServerConfig, 
        store_getter: Callable[[], Optional["Store"]] = lambda: None
    ):
        self.config = config
        self._store_getter = store_getter
        self._runner: Optional[web.AppRunner] = None
        self._site: Optional[web.TCPSite] = None

    @property
    def store(self) -> Optional["Store"]:
        return self._store_getter()

    async def start(self) -> None:
        """Start the API server in the background."""
        if not self.config.enabled:
            logger.info("API server is disabled")
            return

        # Setup the application and routes
        app = web.Application()
        app.add_routes([
            web.get('/health', self._handle_health),
            web.get('/lean/states/finalized', self._handle_finalized_state)
        ])

        # Setup the runner
        self._runner = web.AppRunner(app)
        await self._runner.setup()

        # Start the site
        self._site = web.TCPSite(self._runner, self.config.host, self.config.port)
        await self._site.start()
        
        logger.info(f"API server listening on {self.config.host}:{self.config.port}")

    async def stop(self) -> None:
        """Gracefully stop the server."""
        if self._runner:
            await self._runner.cleanup()
            logger.info("API server stopped")

    async def _handle_health(self, request: web.Request) -> web.Response:
        """
        Handle health check endpoint.
        """
        return web.json_response({"status": "healthy", "service": "lean-spec-api"})

    async def _handle_finalized_state(self, request: web.Request) -> web.Response:
        """
        Handle finalized checkpoint state endpoint.
        
        This offloads the heavy SSZ encoding to a separate thread to ensure
        the main event loop (handling network traffic) remains responsive.
        """
        store = self.store
        if store is None:
            raise web.HTTPServiceUnavailable(reason="Store not initialized")

        finalized = store.latest_finalized
        
        if finalized.root not in store.states:
            raise web.HTTPNotFound(reason="Finalized state not available")

        state = store.states[finalized.root]

        # ---------------------------------------------------------
        # CRITICAL OPTIMIZATION:
        # Run CPU-intensive SSZ encoding in a separate thread.
        # This prevents the server from stuttering/freezing while encoding large states.
        # ---------------------------------------------------------
        try:
            ssz_bytes = await asyncio.to_thread(state.encode_bytes)
        except Exception as e:
            logger.error(f"Failed to encode state: {e}")
            raise web.HTTPInternalServerError(reason="Encoding failed")

        return web.Response(
            body=ssz_bytes,
            content_type="application/octet-stream"
        )
```

README.md
```diff
@@ -23,6 +23,22 @@ This project requires Python 3.12 or later and should be installed via `uv`:
...
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
+
+# Install and set nightly as default
+rustup install nightly
+rustup default nightly
+
+# Verify installation
+rustc --version  # Should show nightly version
+```
+
```
tcoratger: Maybe we should add a line to specify that this is a temporary solution and in the future this will be removed to avoid using bindings and spec out things in Python directly.

src/lean_spec/subspecs/api/client.py
ch4r10t33r: Good suggestion. I don't know what I was thinking by creating a hacky implementation. I have now updated the code to use httpx for the client and aiohttp for the server. This should be cleaner now.

tcoratger (approved):
Looks good to me, thanks a lot.

PR #281, opened 2026-01-14
feat: numpy poseidon permute for test framework keygen

unnawut:
## 🗒️ Description

This PR is an attempt to optimize the keygen logic that took ~2.5 days to generate 12 prod scheme keys on a MacBookPro M1 by using numpy arrays instead of `Fp` objects for Poseidon2 permute.

I retained the same function names and arguments. The numpy implementation seems quite apple to apple with the naive permute so I think we can have it side by side with property-based testing 100 examples each width (is this enough?) to verify the equivalence.

Quick benchmark with low LOG_LIFETIME shows 6-7x faster keygen and it's linear with the increased LOG_LIFETIME. So 2.5 days could go down to 8-10 hours. Maybe with 1 or 2 other optimizations we can get it down to a few hours or less.

For specs consistency, the specs will always use the native permute. The numpy permute replaces the naive one only during keygen and this is done within the consensus_testing package only.

## ✅ Checklist

    ```console
    uvx tox
    ```


tcoratger (commented):
Okay, after seeing this PR, I've changed my mind about what I told you in DMs.

The numpy permutation here is simple enough to be used as the official specification. So here's my proposal:

- You replace the content of our current `poseidon2/permutation.py` file with the content of your `permutation_numpy.py` file, trying to keep the same comments as in the original `poseidon2/permutation.py` file (I tried to write docstrings and line-by-line documentation inside the functions so that reading the code is almost as smooth as reading a book hopefully).

- You can then remove your `test_permutation_numpy.py` because our current test suite already covers this perfectly with reference results taken from Plonky3 audited implementation.

- You don't need the switch anymore for the key generation thing because from now on we'll always use the numpy version everywhere for everything including also our test vectors where I think that we will gain as well.

To be honest, when I read your code, it's almost identical to the original specification, with just a couple of extra characters and things due to numpy, but perfectly manageable as a specification.

Let me know what you think and if you're okay with doing it, or I can merge your pull request and make the adjustments myself if you don't have the time? Whichever you prefer :)

tcoratger (approved):
Thanks, this looks good. Indeed, with your new suggestion using numpy, the s box degree is no longer used as a constant because you represent it manually. 

This is perfectly fine; just in case we change the field, we will need to modify the formula because not all fields have the same s box degree.

PR #282, opened 2026-01-14
Committee aggregation

kamilsa:
## 🗒️ Description
Introduces aggregator role and subnet aggregation.

* Every validator is assigned to one of the attestation subnets
* New role: aggregator. Any validator could become an aggregator.
* Aggregators collect signatures that correspond to validators from their subnet
~* If aggregators collected 90% of signatures from their subnet by the beginning of slot 2, they produce aggregated attestation and propagate it into aggregation topic~
* Validators subscribe to aggregation topic, so that next block proposer may include committee aggregations into the block (without recursive aggregation for now)
* There are **five** 3SF intervals now (each 800ms):
    * Interval 0: block proposal
    * Interval 1: validators vote and gossip their vote to their attestation subnet
    * Interval 2: aggregators use collected votes, aggregate them into aggregated attestation, and broadcast to aggregation topic
    * Interval 3: validators update their safe target based on received aggregated attestations for the current slot
    * Interval 4: validators accept aggregated attestations and move them to known aggregated attestations, so that they are used in fork choice by the next proposer
* `latest_new_attestations` and `latest_known_attestations` were replaced with `latest_new_aggregated_payloads` and `latest_known_aggregated_payloads`
* Some tests became irrelevant and got removed

## 🔗 Related Issues or PRs
https://github.com/leanEthereum/pm/pull/56
https://github.com/leanEthereum/pm/pull/58

## ✅ Checklist

    ```console
    uvx tox
    ```
- [ ] Considered adding appropriate tests for the changes

### DIFF
--- a/src/lean_spec/subspecs/forkchoice/store.py
+++ b/src/lean_spec/subspecs/forkchoice/store.py
@@ -145,6 +145,8 @@ class Store(Container):
 
     This allows reconstructing attestations from aggregated payloads.
     Keyed by data_root_bytes() of AttestationData.
+
+    # TODO(kamilsa): #361 Consider pruning old entries based on justification/finalization.
     """
 
     latest_new_aggregated_payloads: dict[SignatureKey, list[AggregatedSignatureProof]] = {}


docs/client/validator.md
```diff
@@ -16,6 +17,32 @@ diversity helps test interoperability.
...
+attestations propagation, in addition to publishing into per committee subnets.
+This is due to 3SF-mini consensus design, that requires 2/3 + 1 of all 
+attestations to be observed by any validator to compute safe target correctly.
+
+Note that non-aggregating validators do not need to subscribe to committee
+attestation subnets. They only need to subscribe to the global attestation 
+subnet.
+
+Every validator is assigned to a single committee. Number of committees is 
+defined in config.yaml. Each committee maps to a subnet ID. Validators
```
jihoonsong:
```suggestion
defined in config.yaml. Each committee maps to a subnet ID. Validator's
```

docs/client/validator.md
```diff
@@ -16,6 +17,32 @@ diversity helps test interoperability.
...
+subnet ID is derived using their validator index modulo number of committees.
+This is to simplify debugging and testing. In the future, validators subnet id
```
jihoonsong:
```suggestion
This is to simplify debugging and testing. In the future, validator's subnet ID
```

docs/client/validator.md
```diff
@@ -98,8 +125,7 @@ blocks and attestations.
 Attestation aggregation combines multiple attestations into one. This saves bandwidth and
 block space.
 
-Devnet 0 has no aggregation. Each attestation is separate. Future devnets will add
-aggregation.
+Devnet-2 introduced signatures aggregation. Aggregations are produced by block proposers.
 
 When aggregation is added, aggregators will collect attestations and combine them.
 Aggregated attestations will be broadcast separately.
```
jihoonsong:
```suggestion
Devnet-2 introduces signatures aggregation. Aggregators will collect attestations and combine them. Aggregated attestations will be broadcast separately.
```
Perhaps we could do something like this.

docs/client/validator.md
```diff
@@ -16,6 +17,32 @@ diversity helps test interoperability.
 In production, validator assignment will work differently. The current approach
 is temporary for devnet testing.
 
+## Committees and Subnets
```
jihoonsong: Perhaps it is better to narrow down which committees and subnets these are, if we were to have more committee and subnet in the future.

docs/client/validator.md
```diff
@@ -16,6 +17,32 @@ diversity helps test interoperability.
...
+
+Committee is a group of validators assigned to aggregate attestations. 
+Beacon chain uses subnets as network channels for specific committees.
+
+In the devnet-3 design, however, there is one global subnet for signed 
+attestations propagation, in addition to publishing into per committee subnets.
+This is due to 3SF-mini consensus design, that requires 2/3 + 1 of all
```
jihoonsong:
```suggestion
This is due to 3SF-mini consensus design, that requires 2/3+ of all 
```

docs/client/validator.md
```diff
@@ -16,6 +17,32 @@ diversity helps test interoperability.
...
+defined in config.yaml. Each committee maps to a subnet ID. Validators
+subnet ID is derived using their validator index modulo number of committees.
+This is to simplify debugging and testing. In the future, validators subnet id
+will be assigned randomly per epoch.
+
+## Aggregator assignment
+
+Some validators are self-assigned as aggregators. Aggregators collect and combine
+attestations from other validators in their committee. To become an aggregator,
+a validator sets `is_validator` flag to true as ENR record field.
```
jihoonsong:
```suggestion
a validator sets `is_aggregator` flag to true as ENR record field.
```

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -90,6 +91,7 @@ def generate_genesis(cls, genesis_time: Uint64, validators: Validators) -> "Stat
         # Configure the genesis state.
         genesis_config = Config(
             genesis_time=genesis_time,
+            attestation_subnet_count=AGGREGATION_COMMITTEE_COUNT,
```
jihoonsong: I admit that I was the one who advocated for attestation committee, but based on the fact that validators only push their attestations to aggregators in their subnet without subscribing to it, I now think aggregation committee gives us slightly better description.

I don't mind whichever we choose—either attestation committee or aggregation committee—but I do think we need to stick to one thing consistently in the Lean spec and `pq-devnet-3.md` in the pm repo.

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -454,9 +471,92 @@ def on_attestation(
             }
         )
 
+    def on_gossip_committee_aggregation(self, signed_attestation: SignedAggregatedAttestation) -> "Store":
```
jihoonsong:
```suggestion
    def on_gossip_aggregated_attestation(self, signed_attestation: SignedAggregatedAttestation) -> "Store":
```
This seems more aligned with other gossip handlers?

src/lean_spec/subspecs/containers/state/state.py
kamilsa: I don't have a strong preference too, however I think the rationale for topic names in beacon chain spec is based on the type of messages that are being propagated to this topic. For consistency we should probably stick to the same logic and keep using attestation subnets and attestation committees

src/lean_spec/subspecs/networking/gossipsub/topic.py
```diff
@@ -103,6 +116,12 @@ class TopicKind(Enum):
     ATTESTATION = ATTESTATION_TOPIC_NAME
     """Signed attestation messages."""
 
+    ATTESTATION_SUBNET = ATTESTATION_SUBNET_TOPIC_NAME
+    """Attestation subnet messages."""
+
+    COMMITTEE_AGGREGATION = COMMITTEE_AGGREGATION_TOPIC_NAME
```
jihoonsong:
```suggestion
    AGGREGATED_ATTESTATION = AGGREGATED_ATTESTATION_TOPIC_NAME
```

docs/client/validator.md
```diff
@@ -2,8 +2,9 @@
 
 ## Overview
 
-Validators participate in consensus by proposing blocks and producing attestations. This
-document describes what honest validators do.
+Validators participate in consensus by proposing blocks and producing attestations. 
+Optionally validators can opt-in to behave as aggregators in a single or multiple
```
jihoonsong: A validator is assigned to a certain subnet but can they perform an aggregator role in multiple subnets?

docs/client/validator.md
```diff
@@ -16,6 +17,33 @@ diversity helps test interoperability.
 In production, validator assignment will work differently. The current approach
 is temporary for devnet testing.
 
+## Attestation Committees and Subnets
+
+Attestation committee is a group of validators contributing to the common 
+aggregated attestations. Beacon chain uses subnets as network channels for 
+specific committees.
```
jihoonsong:
```suggestion
aggregated attestations. Subnets are network channels dedicated to specific committees.
```
It seems a bit unnecessary to mention beacon chain.

docs/client/validator.md
```diff
@@ -98,11 +126,7 @@ blocks and attestations.
 Attestation aggregation combines multiple attestations into one. This saves bandwidth and
 block space.
 
-Devnet 0 has no aggregation. Each attestation is separate. Future devnets will add
-aggregation.
-
-When aggregation is added, aggregators will collect attestations and combine them.
-Aggregated attestations will be broadcast separately.
+Devnet-2 introduces signatures aggregation. Aggregators will collect attestations and combine them. Aggregated attestations will be broadcast separately.
```
jihoonsong:
```suggestion
Devnet-3 introduces signatures aggregation. Aggregators will collect attestations and combine them. Aggregated attestations will be broadcast separately.
```

src/lean_spec/subspecs/containers/state/state.py
jihoonsong: Sounds great! In the same vein, what do you think about this [one](https://github.com/leanEthereum/leanSpec/pull/282#discussion_r2694018679)?

docs/client/validator.md
kamilsa: Good question. I think we may add this possibility in future. But lets start with one subnet per aggregator.

I think we may consider similar logic to how validators self-assign themselves for sampling of single or multiple columns data in PeerDAS

src/lean_spec/subspecs/containers/state/state.py
kamilsa: looks good to me, applied your suggestion

docs/client/validator.md
jihoonsong: Sounds great! We can update this description to a single committee and modify it later once we introduce multiple subnets aggregator.

g11tech: > Good question. I think we may add this possibility in future. But lets start with one subnet per aggregator.
> 
> I think we may consider similar logic to how validators self-assign themselves for sampling of single or multiple columns data in PeerDAS

I think we need to define this properly, just to clarify, nodes self-assign themselves (validators attached or not), a distinction required here because a signedaggregatedattestation can't be produced here without a validator

docs/client/validator.md
```diff
@@ -16,6 +17,32 @@ diversity helps test interoperability.
...
 is temporary for devnet testing.
 
+## Attestation Committees and Subnets
+
+Attestation committee is a group of validators contributing to the common 
+aggregated attestations. Subnets are network channels dedicated to specific committees.
+
+In the devnet-3 design, however, there is one global subnet for signed 
+attestations propagation, in addition to publishing into per committee subnets.
+This is due to 3SF-mini consensus design, that requires 2/3+ of all
```
g11tech: this global bit is not required, once the aggregtors publish signed attestations in the 2nd interval, they can be imported by all validators in the 3rd interval

src/lean_spec/subspecs/chain/config.py
```diff
@@ -37,6 +37,12 @@
 VALIDATOR_REGISTRY_LIMIT: Final = Uint64(2**12)
 """The maximum number of validators that can be in the registry."""
 
+ATTESTATION_COMMITTEE_COUNT: Final = Uint64(1)
+"""The number of attestation committees per slot."""
+
+COMMITTEE_SIGNATURE_THRESHOLD_RATIO: Final = 0.9
```
g11tech: seems too high, anyway the aggregation should be deadline based and not threshold based, this is something a client implementations can deal with themselves when to do optimistic aggregation

src/lean_spec/subspecs/containers/attestation/attestation.py
```diff
@@ -107,3 +108,10 @@ def aggregate_by_data(
             )
             for data, validator_ids in data_to_validator_ids.items()
         ]
+
+class SignedAggregatedAttestation(Container):
```
g11tech, line 114: @anshalshukla / @GrapeBaBa  do we already have this type?

g11tech: also better to use message, signature terminlogy

g11tech: I think we also need aggregated bit vector here as well,

src/lean_spec/subspecs/chain/config.py
kamilsa: The rationale here is not to waste bandwidth on proofs that aggregate too small amount of signatures.

But for simplicity we may remove it in the first iteration (devnet3) and just aggregate in third interval whatever the aggregator has collected

docs/client/validator.md
kamilsa: Unfortunately, without global topic for attestations, we might not be able to receive proofs in time to update safe target during interval 2:

* Interval 0: block propagation
* Interval 1: votes propagation
* Interval 2: signatures aggregation (up to one second for 1000 validators in subnet with 1000sigs/second expected sigs aggregation rate) + proof distribution => No time for updating safe target => in next slot validator votes for old target

src/lean_spec/subspecs/containers/attestation/attestation.py
kamilsa: > I think we also need aggregated bit vector here as well,

AggregatedSignatureProof contains `AggregationBits`

GrapeBaBa: > @anshalshukla / @GrapeBaBa do we already have this type?

no

docs/client/networking.md
```diff
@@ -62,15 +63,35 @@ Messages are organized by topic. Topic names follow a pattern that includes:
 
 This structure lets clients subscribe to relevant messages and ignore others.
 
+The payload carried in the gossipsub message is the SSZ-encoded, 
+Snappy-compressed message, which type is identified by the topic:
+
+| Topic Name                                                 | Message Type                | Encoding     |
+|------------------------------------------------------------|-----------------------------|--------------|
+| /lean/consensus/devnet3/blocks/ssz_snappy                  | SignedBlockWithAttestation  | SSZ + Snappy |
```
GrapeBaBa: original prefix is `/leanconsensus`, this is an expected change `/lean/consensus`?

src/lean_spec/subspecs/chain/config.py
kamilsa: Removed threshold at all after last week's call

docs/client/networking.md
kamilsa: no.
Thanks for noticing. Fixed

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -884,14 +858,88 @@ def compute_aggregated_signatures(
...
-                        AggregatedAttestation(aggregation_bits=participants, data=data),
-                        proof,
-                    )
-                )
+                attestation = AggregatedAttestation(aggregation_bits=participants, data=data)
+                results.append((attestation, proof))
+
+        return results
+
+    def compute_aggregated_signatures(
```
unnawut: We should remove this `compute_aggregated_signatures()` function since it's only serving the spec's unit tests. I can have a go at this after the PR is merged, but just flagging that clients don't need to implement this function.

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -313,16 +332,27 @@ def on_gossip_attestation(
...
+            else:
+                sig_key = SignatureKey(validator_id, attestation_data.data_root_bytes())
+                new_commitee_sigs[sig_key] = signature
 
         # Process the attestation data
         store = self.on_attestation(attestation=attestation, is_from_block=False)
 
-        # Return store with updated signature map
-        return store.model_copy(update={"gossip_signatures": new_gossip_sigs})
+        # Return store with updated signature maps
```
unnawut: super nit: it's still a single map

```suggestion
        # Return store with updated signature map
```

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -454,9 +484,90 @@ def on_attestation(
...
+                is_from_block=False,
+            )
+
+        # Return store with updated aggregated payloads
+        return store.model_copy(update={"aggregated_payloads": new_aggregated_payloads})
+
     def on_block(
         self,
         signed_block_with_attestation: SignedBlockWithAttestation,
+        current_validator: ValidatorIndex | None = None,
```
unnawut: Should this use the internal `self.validator_id` similar to [`on_gossip_attestation()` above](https://github.com/leanEthereum/leanSpec/pull/282/changes#diff-67c42b82824779fcb11cc28349098aa13e096ed5a4b759919fce36a185ffcbe4R339) instead of allowing a custom arg?

kamilsa: ✅  Introduced `validator_id` in Store

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -849,13 +830,11 @@ def compute_aggregated_signatures(
             # Track validators we couldn't find signatures for.
             #
             # These will need to be covered by Phase 2 (existing proofs).
```
unnawut:
```suggestion
```

since the line below is removed

unnawut (approved):
The diffs are very large and spread out now. I suggest we rebase and merge this asap and do incremental reviews after.

PR #290, opened 2026-01-16
framework: add consensus testing claude skill

### LINKED ISSUE #130: Revisit Fork Choice tests
Currently some of the forkchoice rules are written in a fashion that assumes that we get a new validator for each slot but in reality the default config with 4 validator is being used which is rotated in round robin fashion.
Todos:
- Revisit all the tests to see if it is works as intended even with the default 4 validator set. 
- Eg: `test_three_block_deep_reorg` is incorrect, it doesn't follow the scenario described in the doc string. Total number of validators are 4 so fork A cannot have 3 blocks, and fork B with 4 blocks. some validator votes are being overwritten. Can be fixed by adding multiple validators at genesis, like done in `test_reorg_with_slot_gaps`
- I've just mentioned one case above but there might be more so it's good if we can revisit all of them to ensure everything is as intended.

unnawut:
## 🗒️ Description

Add a claude skill that extends the existing code-tester agent so that it's knowledgeable about how to test consensus related features. With this skill, the agent knows how many validators to setup, the common scenarios for justification/non-justification, and fork choice reorgs.

I didn't include this as part of the code-tester agent and created a new skill instead because the code-tester agent definition is getting quite long, and I assume as the spec gets more feature-rich it's worth segregating the context for other things and avoid the AI getting confused.

To validate this skill, it's used to fix #130 which is also included in this PR.

## 🔗 Related Issues or PRs
#130

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/.claude/agents/code-tester.md
+++ b/.claude/agents/code-tester.md
@@ -11,6 +11,18 @@ You are SpecForge, an elite Test Engineer specializing in the Lean Ethereum Cons
 
 Generate rigorous, comprehensive unit tests and spec test fillers for the leanSpec repository. Your tests verify spec compliance and ensure cross-client interoperability across all modules.
 
+## Auto-Invoke Skills
+
+### Consensus Testing
+
+When writing tests for consensus-related code, invoke the `/consensus-testing` skill first to load specialized multi-validator testing patterns.
+
+**Triggers to invoke the skill:**
+- Test file is in `tests/consensus/`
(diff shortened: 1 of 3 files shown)


tcoratger (approved):
Cool, this makes a lot of sense, I discovered the concept of skills recently and this makes a lot of sense to have them here. I hope that in the future we will be able to make the overall spec more agentic friendly.

PR #294, opened 2026-01-16
fix: fix the duplicate root to slot

GrapeBaBa:
## 🗒️ Description

Fix finalization pruning to handle duplicate roots by tracking all slots per root.

## ✅ Checklist

    ```console
    uvx tox
    ```


tests/lean_spec/subspecs/containers/test_state_justified_slots.py
```diff
@@ -97,3 +108,48 @@ def test_is_slot_justified_raises_on_out_of_bounds() -> None:
         State.generate_genesis(Uint64(0), make_validators(1)).justified_slots.is_slot_justified(
             Slot(0), Slot(1)
         )
+
+
+def test_pruning_with_duplicate_roots_keeps_pending_justification() -> None:
```
tcoratger, line 113: I've checked to run this test locally on the main branch without your fix and it works so I think it doesn't test the right thing regarding your fix in this PR.

PR #296, opened 2026-01-16
state: fix bug about duplicate roots in root to slots map

tcoratger:
## 🗒️ Description

@GrapeBaBa I think this one could replace https://github.com/leanEthereum/leanSpec/pull/294 because here I think that I isolated the bug properly with the test (checked with the code on main and it fails).

What do you think?

## ✅ Checklist

    ```console
    uvx tox
    ```


GrapeBaBa:
Thanks, closed #294

PR #298, opened 2026-01-19
use `// INTERVALS_PER_SLOT` for `store.time` slot calculations

unnawut:
## 🗒️ Description

- Fixes following #222 (thanks @crStiv) in a few more places where the code still assumes `store.time` is stored as seconds
  - For reference, `store.time` is "intervals since genesis" as documented on the [class definition](https://github.com/leanEthereum/leanSpec/blob/main/src/lean_spec/subspecs/forkchoice/store.py#L69-L70) and originally in the [markdown specs](https://github.com/leanEthereum/leanSpec/blob/4b750f2748a3718fe3e1e9cdb3c65e3a7ddabff5/docs/client/forkchoice.md#get_forkchoice_store) 
- Adding `TestGetForkchoiceStore` as test cases for #222
- Removing `TestSlotTimeCalculations` that's not testing actual code

## 🔗 Related Issues or PRs
Relates to #222

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/tests/lean_spec/subspecs/forkchoice/test_time_management.py
+++ b/tests/lean_spec/subspecs/forkchoice/test_time_management.py
@@ -1,7 +1,10 @@
 """Tests for time advancement, intervals, and slot management."""
 
 import pytest
+from hypothesis import given, settings
+from hypothesis import strategies as st
 
+from lean_spec.subspecs.chain.config import INTERVALS_PER_SLOT
 from lean_spec.subspecs.containers import (
     Block,
     BlockBody,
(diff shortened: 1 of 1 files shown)


tests/lean_spec/subspecs/forkchoice/test_time_management.py
```diff
@@ -61,6 +61,46 @@ def sample_store(sample_config: Config) -> Store:
...
+        "anchor_slot,expected_time",
+        [
+            # store.time = anchor_slot * INTERVALS_PER_SLOT
+            # With INTERVALS_PER_SLOT=4:
+            (0, 0),
+            (1, 4),
+            (2, 8),
+            (5, 20),
+            (10, 40),
+            (25, 100),
```
tcoratger: Here in the computations instead of having the hardcoded numbers, can't we use `INTERVALS_PER_SLOT` so that this will resist changes in the constant?

unnawut: Sounds good. Also made it property-based now so it doesn't stick to just the same slots: https://github.com/leanEthereum/leanSpec/pull/298/changes/bad1d29b64b0adea7a1193a3cf9c0910b504fbb4

tcoratger (approved):

PR #303, opened 2026-01-20
Containerize to support lean-quickstart integration

ch4r10t33r:
## 🗒️ Description

This PR adds complete Docker support for running lean_spec as a containerized consensus node with multi-stage builds, environment-based configuration, and comprehensive documentation.

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -2,4 +2,4 @@
 | :------: | :--------------------------------------: | :------------------------------------------------------------------------------------: |
 | Devnet 0 | 4b750f2748a3718fe3e1e9cdb3c65e3a7ddabff5 | https://github.com/leanEthereum/leanSpec/tree/4b750f2748a3718fe3e1e9cdb3c65e3a7ddabff5 |
 | Devnet 1 | 050fa4a18881d54d7dc07601fe59e34eb20b9630 | https://github.com/leanEthereum/leanSpec/tree/050fa4a18881d54d7dc07601fe59e34eb20b9630 |
-| Devnet 2 | fbbacbea4545be870e25e3c00a90fc69e019c5bb | https://github.com/leanEthereum/leanSpec/tree/fbbacbea4545be870e25e3c00a90fc69e019c5bb |
+| Devnet 2 | c187aab89e0ecc6ce9c1fd9304fd708312ea7106 | https://github.com/leanEthereum/leanSpec/tree/c187aab89e0ecc6ce9c1fd9304fd708312ea7106 |


src/lean_spec/subspecs/chain/service.py
```diff
@@ -133,6 +136,27 @@ async def run(self) -> None:
             # the updated time.
             self.sync_service.store = new_store
 
+            # Log slot progression after tick
+            seconds_since_genesis = int(current_time - self.clock.genesis_time)
+            # 4 intervals per slot
+            current_slot = seconds_since_genesis // (int(SECONDS_PER_INTERVAL) * 4)
```
tcoratger: We have a `INTERVALS_PER_SLOT` constant to avoid the hardcoded 4 here.

README.md
```diff
@@ -203,18 +209,147 @@ def test_withdrawal_amount_above_uint64_max():
...
+docker run --rm -it lean-spec:dev
+
+# Mount local directory for development
+docker run --rm -it -v $(pwd):/app lean-spec:dev
+```
+
+## Documentation
+
+- **[DOCKER_QUICKSTART.md](DOCKER_QUICKSTART.md)** - Complete Docker deployment guide with examples
+- **[CONTRIBUTING.md](CONTRIBUTING.md)** - Contribution guidelines
```
tcoratger: We already have the `Contributing` section for this no?

```suggestion
```

src/lean_spec/subspecs/chain/service.py
```diff
@@ -133,6 +136,27 @@ async def run(self) -> None:
...
+            current_interval = int((current_time - self.clock.genesis_time) // SECONDS_PER_INTERVAL)
```
tcoratger:
```suggestion
            current_interval = seconds_since_genesis // SECONDS_PER_INTERVAL)
```

src/lean_spec/subspecs/chain/service.py
```diff
@@ -133,6 +136,27 @@ async def run(self) -> None:
...
+
+            # Get head and finalized info (defensive for testing/mocking)
+            head_str = new_store.head.hex()[:16] if hasattr(new_store, "head") else "unknown"
```
tcoratger: Why do we do `[:16]` here?

Dockerfile
```diff
@@ -0,0 +1,203 @@
+# Lean Ethereum Specifications - Docker Image
```
tcoratger:
```suggestion
# Lean Consensus Specification - Docker Image
```

README.md
```diff
@@ -2,8 +2,14 @@
 
 The Lean Ethereum protocol specifications and cryptographic subspecifications.
 
+> **🐳 Running with Docker?** Skip the setup and jump to [DOCKER_QUICKSTART.md](DOCKER_QUICKSTART.md) for containerized deployment examples.
+
 ## Quick Start
 
+**New to lean_spec?** Choose your path:
```
tcoratger:
```suggestion
**New to leanSpec?** Choose your path:
```

tcoratger (commented):
Just left couple of comments but I'm far from a Docker specialist so maybe @unnawut, @g11tech or @KatyaRyazantseva can bring another perspective here?

src/lean_spec/subspecs/chain/service.py
unnawut: Or generally for these few lines maybe we can use [`SlotClock.current_slot()`](https://github.com/leanEthereum/leanSpec/blob/main/src/lean_spec/subspecs/chain/clock.py#L46-L53), etc?

tests/lean_spec/subspecs/chain/test_service.py
unnawut, line 1: This might be out of scope of this PR

DOCKER_QUICKSTART.md
```diff
@@ -0,0 +1,164 @@
...
+### 3. Validator Node
+
+Run as a validator with keys:
+
+```bash
+docker run --rm \
+  -v /path/to/genesis:/app/data:ro \
+  -e GENESIS_FILE=/app/data/config.yaml \
+  -e VALIDATOR_KEYS_PATH=/app/data \
+  -e NODE_ID=zeam_0 \
```
unnawut:
```suggestion
  -e NODE_ID=lean_spec_0 \
```

applies to a few more places

DOCKER_QUICKSTART.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Troubleshooting
+
+### Error: "GENESIS_FILE environment variable is required"
+
+Make sure you're setting the `-e GENESIS_FILE=...` environment variable.
+
+### Error: "License file does not exist"
+
+You may need to rebuild the image. The Dockerfile now includes LICENSE and README.md.
```
unnawut, line 191:
```suggestion
```

Not sure if this can still happen. We have both LICENSE and README.md

Dockerfile
```diff
@@ -0,0 +1,203 @@
...
+
+# Node ID
+if [ -n "$NODE_ID" ]; then
+    CMD="$CMD --node-id $NODE_ID"
+fi
+
+# Verbose logging
+if [ "$VERBOSE" = "true" ] || [ "$VERBOSE" = "1" ]; then
+    CMD="$CMD -v"
+fi
```
unnawut: I wonder if we need to maintain this mapping here, or can we pass the CLI arguments directly from docker run command into the client CLI if we set CMD to the node binary instead of /bin/bash?

But not a blocker for merging

README.md
```diff
@@ -203,18 +209,147 @@ def test_withdrawal_amount_above_uint64_max():
...
+### Development with Docker
+
+```bash
+# Interactive development shell
+docker run --rm -it lean-spec:dev
+
+# Mount local directory for development
+docker run --rm -it -v $(pwd):/app lean-spec:dev
+```
+
```
unnawut, line 287: Personally I think this is takes too much estate from main README.md and overlapping content with DOCKER_QUICKSTART.md. Docker build and docker run section seems enough to get going. Can we maybe move this chunk to DOCKER_QUICKSTART.md?

Dockerfile
```diff
@@ -0,0 +1,203 @@
...
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
+
+# Create non-root user for security
+RUN useradd --create-home --shell /bin/bash leanspec
+USER leanspec
+
+WORKDIR /app
+
+# Copy the virtual environment and source from builder
+COPY --from=builder --chown=leanspec:leanspec /app/.venv /app/.venv
```
unnawut: I think venv is specific to a machine so not sure if it should be copied

```suggestion
# Copy the source from builder
```

src/lean_spec/subspecs/chain/service.py
ch4r10t33r: To make it easier to scan, otherwise the log is too long. However, I do not have a strong opinion on this. I can remove it.

Dockerfile
ch4r10t33r: This is fair. I'll change this as well.

Dockerfile
ch4r10t33r: Actually, copying the venv is the correct and recommended approach for multi-stage Docker builds when both stages use the same base image. I've updated the comments to reflect this.

unnawut: ah ok i overlooked the --from=builder part

tcoratger (approved):

PR #305, opened 2026-01-21
refactor: don't use zero hash in justification roots

GrapeBaBa:
## 🗒️ Description

Update leanSpec justifications logic to behavior: zero‑hash votes are ignored, zero‑hash roots are invalid, and pruning uses the latest slot per root with an error on missing mapping.

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/src/lean_spec/subspecs/containers/state/state.py
+++ b/src/lean_spec/subspecs/containers/state/state.py
@@ -404,9 +404,9 @@ class State(Container):
         #       (block root) → [vote flags for validators 0..N-1]
         #
         # which makes the rest of the logic easier to express and understand.
-        assert not any(
-            root == ZERO_HASH for root in self.justifications_roots
-        ), "zero hash is not allowed in justifications roots"
+        assert not any(root == ZERO_HASH for root in self.justifications_roots), (
+            "zero hash is not allowed in justifications roots"
+        )
         justifications = (
             {
                 root: self.justifications_validators[
@@ -569,9 +569,9 @@ class State(Container):
                     delta = int(finalized_slot - old_finalized_slot)
                     if delta > 0:
                         justified_slots = justified_slots.shift_window(delta)
-                        assert all(
-                            root in root_to_slot for root in justifications
-                        ), "Justification root missing from root_to_slot"
+                        assert all(root in root_to_slot for root in justifications), (
+                            "Justification root missing from root_to_slot"
+                        )
                         justifications = {
                             root: votes
                             for root, votes in justifications.items()


src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -404,6 +404,8 @@ def process_attestations(
         #       (block root) → [vote flags for validators 0..N-1]
         #
         # which makes the rest of the logic easier to express and understand.
+        if any(root == ZERO_HASH for root in self.justifications_roots):
+            raise AssertionError("ZERO_HASH is not allowed in justifications_roots")
```
unnawut: If we are already preventing zero hash from entering `self.justifications_roots` in the first place, maybe this assertion is not needed?

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -558,16 +563,17 @@ def process_attestations(
...
-                    # We also prune any pending justifications whose slots are now finalized.
-                    # A root may appear at multiple slots; keep the justification if ANY
-                    # slot for that root is still unfinalized (conservative approach).
+                    # We also prune any pending justifications whose latest slot
+                    # is now finalized (latest <= finalized_slot).
                     delta = int(finalized_slot - old_finalized_slot)
                     if delta > 0:
                         justified_slots = justified_slots.shift_window(delta)
+                        if any(root not in root_to_slot for root in justifications):
+                            raise AssertionError("Justification root missing from root_to_slot")
```
unnawut: I might be wrong on this. Isn't `root in justifications`, which in order to be in justifications, was already [checked that it must exist](https://github.com/leanEthereum/leanSpec/pull/305/changes#diff-c33ac3d9a3d230286b9b58b4608c6452a177cc99780ab25d661aef495114b147R467-R478) in `self.historical_block_hashes`?

And `root_to_slot` is [constructed](https://github.com/leanEthereum/leanSpec/pull/305/changes#diff-c33ac3d9a3d230286b9b58b4608c6452a177cc99780ab25d661aef495114b147R432-R436) from `self.historical_block_hashes` so this assertion is unreachable?

unnawut: Actually for spec readability, it could make sense to have this but maybe assert instead of raise?

src/lean_spec/subspecs/containers/state/state.py
unnawut: Same comment as below for spec readability maybe we keep this but as an `assert`?

unnawut:
Also should we have this in devnet2 or 3?

Since devnet2 spec was already cut off in https://github.com/leanEthereum/pm/pull/60 we'll need to let all teams know this is coming in if we need it in devnet2

src/lean_spec/subspecs/containers/state/state.py
g11tech: someone can send an invalid block no?

g11tech: ahh yes a simple assert might also be just fine

src/lean_spec/subspecs/containers/state/state.py
g11tech: again someone can give/contruct an invalid block, agree simple asset is just fine, clients can add more descriptive errors if they like

src/lean_spec/subspecs/containers/state/state.py
unnawut: > someone can send an invalid block no?

I think this case is covered here: https://github.com/GrapeBaBa/leanSpec/blob/2bc13509099e459d342018f54813f798eb168f24/src/lean_spec/subspecs/containers/state/state.py#L474-L478 so attestations to invalid blocks wouldn't enter `self.justifications_roots` in the first place

but I guess it's useful for spec clarity so converting to assert would be great

unnawut (approved):
❤️

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -404,6 +404,9 @@ def process_attestations(
         #       (block root) → [vote flags for validators 0..N-1]
         #
         # which makes the rest of the logic easier to express and understand.
+        assert not any(
+            root == ZERO_HASH for root in self.justifications_roots
+        ), "ZERO_HASH is not allowed in justifications_roots"
```
```suggestion
        ), "zero hash is not allowed in justifications roots"
```

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -421,18 +424,17 @@ def process_attestations(
         finalized_slot = latest_finalized.slot
         justified_slots = self.justified_slots
 
-        # Map roots to all slots where they appear.
+        # Map roots to their latest slot for pruning.
         #
-        # Missed slots produce duplicate zero hashes in history.
-        # During pruning, we must check all slot occurrences for each root.
-        # A single-slot mapping would fail when iterating over slots.
+        # Votes for ZERO_HASH are ignored, so we only need the most recent slot
```
```suggestion
        # Votes for zero hash are ignored, so we only need the most recent slot
```

PR #309, opened 2026-01-21
feat: add `/lean/states/justified` API endpoint

ch4r10t33r:
## 🗒️ Description
Add new HTTP endpoint to expose the latest justified checkpoint:
- Returns JSON with slot number and root hash
- Follows same pattern as /lean/states/finalized endpoint
- Returns 503 when store not initialized

This enables monitoring of consensus justification progress and provides visibility into fork choice state for operators.

Response format:
{
  "slot": <slot_number>,
  "root": "<hex_root_hash>"
}

Tests:
- Verify 503 when store not available
- Verify JSON response format and content
- All 13 API tests pass

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/tests/lean_spec/subspecs/api/test_server.py
+++ b/tests/lean_spec/subspecs/api/test_server.py
@@ -182,7 +182,7 @@ class TestJustifiedEndpoint:
                     assert response.status_code == 200
                     assert "application/json" in response.headers["content-type"]
 
-                    # Verify JSON structure
+                    # Verify JSON structure and types
                     data = response.json()
                     assert "slot" in data
                     assert "root" in data
@@ -190,6 +190,10 @@ class TestJustifiedEndpoint:
                     assert isinstance(data["root"], str)
                     # Root should be a hex string
                     assert len(data["root"]) == 64  # 32 bytes * 2 hex chars
+
+                    # Verify actual values match the store's latest justified checkpoint
+                    assert data["slot"] == int(base_store.latest_justified.slot)
+                    assert data["root"] == base_store.latest_justified.root.hex()
 
             finally:
                 server.stop()


src/lean_spec/subspecs/api/server.py
```diff
@@ -3,6 +3,7 @@
 
 Provides HTTP endpoints for:
 - /lean/states/finalized - Serve finalized checkpoint state as SSZ
+- /lean/state/justified - Return latest justified checkpoint information
```
syjn99:
```suggestion
- /lean/states/justified/root - Return latest justified checkpoint root
```

Seems like this can be more RESTful, also matching with [Beacon-APIs](https://ethereum.github.io/beacon-APIs/#/Beacon/getStateRoot). Also I think versioning APIs is always good, will open a PR on this to get some feedbacks.

tcoratger: Agreed, I think this requires a modification at different places through the PR as well.

tests/lean_spec/subspecs/api/test_server.py
```diff
@@ -142,6 +142,62 @@ async def run_test() -> None:
...
+                    assert "application/json" in response.headers["content-type"]
+
+                    # Verify JSON structure
+                    data = response.json()
+                    assert "slot" in data
+                    assert "root" in data
+                    assert isinstance(data["slot"], int)
+                    assert isinstance(data["root"], str)
+                    # Root should be a hex string
+                    assert len(data["root"]) == 64  # 32 bytes * 2 hex chars
```
tcoratger: In addition to checking the types and length can we check that the dates are the right ones too? Meaning check the root and slot directly against the expectation.

src/lean_spec/subspecs/api/server.py
ch4r10t33r: This was a typo from my side. I'll rename it to /lean/states/justified. I think root is redundant here.

ch4r10t33r: @syjn99 Pls go ahead with the versioning in a separate PR.

PR #368, opened 2026-02-06
Add is-aggregator configuration flag

ch4r10t33r:
## Summary

Introduces a new `--is-aggregator` command-line parameter that enables aggregator mode for consensus nodes. When enabled, the node performs attestation aggregation operations and advertises this capability in its ENR for peer discovery.

## Changes

- **NodeConfig**: Add `is_aggregator` field with default value `False`
- **ENR Keys**: Add `IS_AGGREGATOR` constant for ENR key-value pairs
- **ENR Class**: Add `is_aggregator` property to read aggregator flag from ENR records
- **NetworkService**: Add `is_aggregator` field for ENR advertisement
- **CLI**: Add `--is-aggregator` argument with help text
- **Initialization**: Thread flag through genesis and checkpoint sync paths
- **Logging**: Add informational log when aggregator mode is enabled

## Usage

```bash
# Run node in standard mode (default)
python -m lean_spec --genesis config.yaml --validator-keys ./keys

# Run node in aggregator mode
python -m lean_spec --genesis config.yaml --validator-keys ./keys --is-aggregator
```

### Configuration Flow
The flag flows through the entire initialization chain:
```
CLI → run_node() → _init_from_genesis() → NodeConfig → NetworkService
```

### ENR Advertisement
- When `is_aggregator=True`, the node's ENR includes the `is_aggregator` key
- The key uses 1 byte encoding: `0x00` = false, `0x01` = true
- Other nodes can query this via the `ENR.is_aggregator` property

### Backward Compatibility
- Default value is `False`, maintaining existing behavior
- No breaking changes to existing APIs
- Optional feature that can be enabled per-node

## Test plan

- [ ] Verify CLI argument parsing works correctly
- [ ] Test node initialization with `--is-aggregator` flag
- [ ] Confirm flag propagates through NodeConfig to NetworkService
- [ ] Verify ENR property reads aggregator flag correctly
- [ ] Test both genesis and checkpoint sync initialization paths
- [ ] Confirm logging output when aggregator mode is enabled
- [ ] Verify backward compatibility (default behavior unchanged)

## Next Steps

Future work to complete aggregator functionality:
1. Update ENR creation/signing to include `is_aggregator` key in the record
2. Implement attestation aggregation logic in validator service
3. Add runtime checks to enable/disable aggregation based on flag
4. Add metrics for aggregation operations

### DIFF
--- a/src/lean_spec/subspecs/node/helpers.py
+++ b/src/lean_spec/subspecs/node/helpers.py
@@ -3,18 +3,24 @@
 from lean_spec.subspecs.containers.validator import ValidatorIndex
 
 
-def is_aggregator(validator_id: ValidatorIndex | None) -> bool:
+def is_aggregator(
+    validator_id: ValidatorIndex | None,
+    node_is_aggregator: bool = False,
+) -> bool:
     """
     Determine if a validator is an aggregator.
 
+    A validator acts as an aggregator when:
+    1. The validator is active (validator_id is not None)
+    2. The node operator has enabled aggregator mode
+
     Args:
         validator_id: The index of the validator.
+        node_is_aggregator: Whether the node is configured as an aggregator.
 
     Returns:
-        True if the validator is an aggregator, False otherwise.
+        True if the validator should perform aggregation, False otherwise.
     """
     if validator_id is None:
         return False
-    return (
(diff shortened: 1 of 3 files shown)


tcoratger (approved):
Thanks!

PR #380, opened 2026-02-10
tests: add test vectors for SSZ containers

unnawut:
## 🗒️ Description

Add test vectors for SSZ containers

## ✅ Checklist

    ```console
    uvx tox
    ```


src/lean_spec/subspecs/ssz/hash.py
unnawut, line 1: Need Fp implementation here so that Fp in the Signature can be serialized and merkleized

PR #385, opened 2026-02-11
Aggregation broadcast & unskip interop tests

kamilsa:
## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/tests/interop/helpers/node_runner.py
+++ b/tests/interop/helpers/node_runner.py
@@ -202,6 +202,8 @@ class NodeCluster:
     def __post_init__(self) -> None:
         """Initialize validators and keys."""
         self._generate_validators()
+        # Default genesis time for single-node starts.
+        # start_all() overrides this to align with service start.
         self._genesis_time = int(time.time())
 
     def _generate_validators(self) -> None:
@@ -417,6 +419,13 @@ class NodeCluster:
         if validators_per_node is None:
             validators_per_node = self._distribute_validators(num_nodes)
 
+        # Set genesis time to coincide with service start.
+        #
+        # Phases 1-3 (node creation, connection, mesh stabilization) take ~10s.
+        # Setting genesis in the future prevents wasting slots during setup.
+        # The first block will be produced at slot 1, shortly after services start.
+        self._genesis_time = int(time.time()) + 10
+
         # Phase 1: Create nodes with networking ready but services not running.
         #
         # This allows the gossipsub mesh to form before validators start
@@ -464,10 +473,8 @@ class NodeCluster:
         # 2. Subscription RPCs to be exchanged
         # 3. GRAFT messages to be sent and processed
         #
-        # A longer delay ensures proper mesh formation before block production.
-        # CI machines need more time due to lower CPU/scheduling priority.
-        # Increased to 15s to handle slow CI and strict attestation target walkback.
-        await asyncio.sleep(15.0)
+        # 5s allows ~7 heartbeats which is sufficient for mesh formation.
+        await asyncio.sleep(5.0)
 
         # Phase 4: Start node services (validators, chain service, etc).
         #
--- a/tests/interop/test_multi_node.py
+++ b/tests/interop/test_multi_node.py
@@ -45,7 +45,7 @@ logger = logging.getLogger(__name__)
 pytestmark = pytest.mark.interop
 
 
-@pytest.mark.timeout(300)
+@pytest.mark.timeout(150)
 @pytest.mark.num_validators(3)
 async def test_mesh_finalization(node_cluster: NodeCluster) -> None:
     """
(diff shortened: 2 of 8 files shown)


tests/interop/test_multi_node.py
```diff
@@ -100,108 +100,31 @@ async def test_mesh_finalization(node_cluster: NodeCluster) -> None:
     # Each node needs at least 2 peers (the other two nodes).
     # This ensures gossip will reach all nodes.
     # The 15s timeout handles slow handshakes.
```
tcoratger: It's a 30 seconds timeout now no? And not a 15 seconds timeout anymore

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -937,15 +918,15 @@ def update_safe_target(self) -> "Store":
...
+    def aggregate_committee_signatures(self) -> tuple["Store", list[SignedAggregatedAttestation]]:
         """
         Aggregate committee signatures for attestations in committee_signatures.
 
         This method aggregates signatures from the gossip_signatures map.
         Attestations are reconstructed from gossip_signatures using attestation_data_by_root.
 
         Returns:
-            New Store with updated latest_new_aggregated_payloads.
+            Tuple of (new Store with updated payloads, list of new SignedAggregatedAttestation).
```
tcoratger, line 953: Just to avoid variable names in the documentation because these names can change over time, rendering the documentation obsolete

```suggestion
            Tuple of (new store with updated payloads, list of new signed aggregated attestation).
```

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -1042,11 +1026,12 @@ def tick_interval(self, has_proposal: bool, is_aggregator: bool = False) -> "Sto
             is_aggregator: Whether the node is an aggregator.
 
         Returns:
-            New Store with advanced time and interval-specific updates applied.
+            Tuple of (new Store with advanced time, list of new SignedAggregatedAttestation).
```
```suggestion
            Tuple of (new store with advanced time, list of new signed aggregated attestation).
```

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -1079,22 +1066,25 @@ def on_tick(self, time: Uint64, has_proposal: bool, is_aggregator: bool = False)
             is_aggregator: Whether the node is an aggregator.
 
         Returns:
-            New Store with time advanced and all interval actions performed.
+            Tuple of (new Store with time advanced,
+            list of all produced SignedAggregatedAttestation).
```
```suggestion
            Tuple of (new store with time advanced,
            list of all produced signed aggregated attestation).
```

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -1168,8 +1158,11 @@ def get_attestation_target(self) -> Checkpoint:
         #
         # This ensures the target doesn't advance too far ahead of safe target,
         # providing a balance between liveness and safety.
+        #
+        # MODIFIED: We allow the target to be up to 1 slot ahead of safe_target
```
```suggestion
        # We allow the target to be up to 1 slot ahead of the safe target
```

tests/interop/helpers/node_runner.py
```diff
@@ -350,9 +357,21 @@ async def start_node(
...
         event_source.subscribe_gossip_topic(block_topic)
-        event_source.subscribe_gossip_topic(attestation_topic)
+        event_source.subscribe_gossip_topic(aggregation_topic)
+
+        # Determine subnets for our validators and subscribe.
+        #
+        # Validators only subscribe to the subnets they are assigned to.
+        # This matches the Ethereum gossip specification.
+        if validator_indices:
+            from lean_spec.subspecs.chain.config import ATTESTATION_COMMITTEE_COUNT
```
tcoratger: Can we have the import top of the file instead?

tests/interop/test_multi_node.py
```diff
@@ -242,60 +165,23 @@ async def test_mesh_2_2_2_finalization(node_cluster: NodeCluster) -> None:
...
+    # Verify heads converged across nodes.
     #
     # Hub-and-spoke adds latency but should not cause divergence.
-    head_diff = max(head_slots) - min(head_slots)
-    assert head_diff <= 2, f"Head slots diverged too much. Slots: {head_slots}, diff: {head_diff}"
-
-    # ALL nodes must finalize.
-    assert all(slot > 0 for slot in finalized_slots), (
-        f"Not all nodes finalized. Finalized slots: {finalized_slots}"
-    )
```
tcoratger, line 291: From the documentation, I guess you removed all this due to CI limitation?

tests/consensus/devnet/fc/test_attestation_target_selection.py
```diff
@@ -17,39 +17,36 @@ def test_attestation_target_at_genesis_initially(
...
     Why This Matters
     ----------------
-    Initially, the safe target is at genesis (slot 0), so the attestation
-    target walks back from head to genesis.
-
-    This conservative behavior ensures validators don't attest too far ahead
-    before there's sufficient attestation weight to advance the safe target.
+    Initially, the safe target is at genesis (slot 0). The attestation target
+    is allowed up to 1 slot ahead of safe target to ensure the chain can
+    start advancing from genesis.
```
tcoratger: With this new condition, are we sure that all the flow that we are testing is the same as before?

kamilsa: this is probably not safe as we could vote for the new head as for the target in such case, which is not correct.
Let me think about workaround

tests/interop/test_multi_node.py
kamilsa: yes, these tests are flaky on CI. I tried to replace CI host machines to macos M1, but still they could end up with different results every run (see [here](https://github.com/leanEthereum/leanSpec/actions/runs/21950711437/job/63403744516))

If we fix CI we could keep the old logic

tests/consensus/devnet/fc/test_attestation_target_selection.py
kamilsa: removed + Slot(1) from get attestation target

src/lean_spec/subspecs/chain/service.py
```diff
@@ -162,11 +168,18 @@ async def _initial_tick(self) -> Interval | None:
...
                 has_proposal=False,
+                is_aggregator=self.sync_service.is_aggregator,
             )
             self.sync_service.store = new_store
+
+            # Publish any aggregated attestations produced during catch-up.
+            if new_aggregated_attestations:
+                for agg in new_aggregated_attestations:
+                    await self.sync_service.publish_aggregated_attestation(agg)
+
```
unnawut, line 181: If the validator is catching up, i.e. many slots behind head slot, wouldn't this spam the network by publishing many stale aggregated_attestations returned by on_tick() for old slots?

I'm not sure if there's a scenario where a validator catching up can produce a useful aggregation but correct me if I'm wrong.

src/lean_spec/subspecs/sync/service.py
```diff
@@ -402,7 +412,8 @@ async def on_gossip_block(
     async def on_gossip_attestation(
         self,
         attestation: SignedAttestation,
-        peer_id: PeerId,  # noqa: ARG002
+        subnet_id: int,
```
unnawut: I see subnet_id is added but not being used right now. Is there a plan to use it? If not I'm thinking we could remove it just to simplify the specs.

Also applies to `peer_id` in `on_gossip_aggregated_attestation()` below

src/lean_spec/subspecs/sync/service.py
```diff
@@ -452,6 +464,45 @@ async def on_gossip_attestation(
...
+            signed_attestation: The signed aggregated attestation received.
+            peer_id: The peer that propagated the aggregate (unused for now).
+        """
+        if not self._state.accepts_gossip:
+            return
+
+        try:
+            self.store = self.store.on_gossip_aggregated_attestation(signed_attestation)
+        except (AssertionError, KeyError):
+            # Aggregation validation failed.
```
unnawut: Can we log this at least as a warning? It'll be very hard to debug dropped aggregations otherwise. Thanks!

src/lean_spec/subspecs/node/node.py
```diff
@@ -238,6 +238,12 @@ def from_genesis(cls, config: NodeConfig) -> Node:
             is_aggregator=config.is_aggregator,
         )
 
+        # Wire up aggregated attestation publishing.
+        #
+        # SyncService delegates aggregate publishing to NetworkService
+        # via a callback, avoiding a circular dependency.
+        sync_service._publish_agg_fn = network_service.publish_aggregated_attestation
```
unnawut, line 245: Shouldn't this aggregation fn be registered with `ValidatorService` below than the sync service? By current [SyncService definition](https://github.com/leanEthereum/leanSpec/blob/875af64125892f898dcd07b7921ad78ebe8e3eb3/src/lean_spec/subspecs/sync/service.py#L1-L35) I think publishing aggregation is quite far from its scope.

But if we want to decouple aggregator from the rest of validator role, we could also create AggregatorService instead of bundling to SyncService.

Just a suggestion. If it makes sense we can also take this as a separate Github issue.

tests/consensus/devnet/fc/test_attestation_target_selection.py
unnawut, line 1: We could remove these changes as they're not critical to the PR. But no big deal we can avoid this in future big PRs.

PR #395, opened 2026-02-12
tests: refactor state expectation framework

tcoratger:
## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/packages/testing/src/consensus_testing/test_types/state_expectation.py
+++ b/packages/testing/src/consensus_testing/test_types/state_expectation.py
@@ -121,7 +121,7 @@ class StateExpectation(CamelModel):
         for field_name in self.model_fields_set:
             accessor = self._ACCESSORS.get(field_name)
             if accessor is None:
-                continue
+                raise ValueError(f"No accessor defined for field: {field_name}")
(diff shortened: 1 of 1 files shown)


packages/testing/src/consensus_testing/test_types/state_expectation.py
```diff
@@ -91,149 +112,19 @@ def validate_against_state(self, state: "State") -> None:
...
-            elif field_name == "justifications_validators":
-                if state.justifications_validators != expected_value:
-                    raise AssertionError(
-                        f"State validation failed: justifications_validators = "
-                        f"{state.justifications_validators}, expected {expected_value}"
-                    )
+        for field_name in self.model_fields_set:
+            accessor = self._ACCESSORS.get(field_name)
+            if accessor is None:
+                continue
```
fselmo: Didn't check this but just adding a thought in case it helps (take or leave). We could raise here instead of silently ignoring in case this helps us add an accessor for any newly introduced field.

```suggestion
                raise ValueError(f"No accessor defined for field: {field_name}")
```

PR #418, opened 2026-02-25
Add `GET /lean/v0/fork_choice` JSON endpoint to leanSpec

pablodeymo:
## 🗒️ Description

Add `GET /lean/v0/fork_choice` JSON endpoint to leanSpec, matching the endpoint already available in ethlambda's Rust implementation.

The endpoint returns a snapshot of the fork choice tree:

- **nodes**: all blocks at or above the finalized slot, with their root, slot, parent_root, proposer_index, and attestation weight
- **head**: current canonical head root
- **justified**: latest justified checkpoint (slot + root)
- **finalized**: latest finalized checkpoint (slot + root)
- **safe_target**: safe target root
- **validator_count**: number of validators in the head state

### Changes

**New `compute_block_weights` method on Store** (`store.py`): extracts the weight computation loop from `_compute_lmd_ghost_head` into a public method. Walks backward from each validator's latest head vote, accumulating weight per block above the finalized slot.

**New endpoint handler** (`endpoints/fork_choice.py`): follows the same pattern as `checkpoints.py`. Returns 503 when the store is not initialized.

**Route registration** (`routes.py`, `endpoints/__init__.py`): wires up `/lean/v0/fork_choice`.

**⚠️ Note:** This implementation duplicates the logic of weight computation. This should be done in only one place.

**Tests**:
- `test_server.py`: 503 test for the new endpoint (port 15058)
- `test_compute_block_weights.py`: genesis-only (empty weights), linear chain (upward accumulation), and multiple attestations (weight stacking)

## 🔗 Related Issues or PRs

N/A

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/tests/lean_spec/subspecs/forkchoice/test_compute_block_weights.py
+++ b/tests/lean_spec/subspecs/forkchoice/test_compute_block_weights.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from lean_spec.subspecs.containers.attestation import AggregationBits, AttestationData
 from lean_spec.subspecs.containers.checkpoint import Checkpoint
 from lean_spec.subspecs.containers.slot import Slot
-from lean_spec.subspecs.containers.validator import ValidatorIndex
+from lean_spec.subspecs.containers.validator import ValidatorIndex, ValidatorIndices
 from lean_spec.subspecs.forkchoice import Store
 from lean_spec.subspecs.ssz.hash import hash_tree_root
 from lean_spec.subspecs.xmss.aggregation import AggregatedSignatureProof, SignatureKey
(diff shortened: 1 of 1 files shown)


tests/lean_spec/subspecs/api/test_server.py
```diff
@@ -102,3 +102,24 @@ async def test_returns_503_when_store_not_initialized(self) -> None:
...
+
+        try:
+            async with httpx.AsyncClient() as client:
+                response = await client.get("http://127.0.0.1:15058/lean/v0/fork_choice")
+
+                assert response.status_code == 503
+
+        finally:
+            server.stop()
+            await asyncio.sleep(0.1)
```
tcoratger, line 125: Can't we have a test with a 200 OK code as well to test the validity?

PR #419, opened 2026-02-26
fix: allow extra YAML keys in GenesisConfig and log startup failures

ch4r10t33r:
- GenesisConfig: set extra='ignore' so configs from lean-quickstart (ACTIVE_EPOCH, VALIDATOR_COUNT, etc.) load without ValidationError
- __main__: catch exceptions, log with logger.exception, exit 1 so node startup failures are visible instead of silent exit 0

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/tests/lean_spec/snappy/test_snappy.py
+++ b/tests/lean_spec/snappy/test_snappy.py
@@ -37,7 +37,7 @@ def load_test_file(filename: str, size_limit: int = 0) -> bytes:
     return data
 
 
-# Test data files as defined in the C++ test suite (alice29.txt omitted - not in repo)
+# Test data files as defined in the C++ test suite
 TEST_DATA_FILES = [
     ("html", "html", 0),
     ("urls", "urls.10K", 0),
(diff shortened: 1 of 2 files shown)


src/lean_spec/subspecs/genesis/config.py
```diff
@@ -28,19 +28,12 @@ class GenesisConfig(StrictBaseModel):
...
-
-    The genesis configuration solves two fundamental coordination problems:
-
-    - Time Synchronization: All nodes agree on when slots begin
-    - Initial Trust: Bootstrap validators that can produce and attest blocks
-
-    Field names use UPPERCASE to match cross-client YAML convention.
-    Pydantic aliases map them to snake_case Python attributes.
+    Ignores extra YAML keys (e.g. ACTIVE_EPOCH, VALIDATOR_COUNT) so configs
+    from lean-quickstart and other generators load without error.
```
tcoratger: Can we just add this comment as complementary to the previous one that was a general explanation and not remove it?

ch4r10t33r: Fixed.

tcoratger (approved):

PR #426, opened 2026-02-28
devnet-4, recursive aggregation

anshalshukla:
## 🗒️ Description
draft PR for devnet-4
- updates aggregation API to support recursive aggregation
- adds a temporary dummy flag which will be used to later fix the tests once the design is finalized
- the bindings cannot be updated due to incompatibility between leanMultisig xmss and leanSig
- INV_PROOF_SIZE should be a protocol level parameter imo, unlike a aggregator paramater as defined in high level doc

- changes `aggregate_gossip_signatures` to `aggregate` as it will no longer just aggregates gossip signatures but also recursively aggregates children_payloads with them

- changes `aggregate_gossip_signatures` to `aggregate_committee_signatures_and_payloads`, it includes both committee gossip signatures and payloads (no committee) and recursively aggregate them using `aggregate`

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/src/lean_spec/subspecs/containers/state/state.py
+++ b/src/lean_spec/subspecs/containers/state/state.py
@@ -861,17 +861,18 @@ class State(Container):
 
         return final_block, post_state, aggregated_attestations, aggregated_signatures
 
-    def aggregate_gossip_signatures(
+    def aggregate(
         self,
         attestations: Collection[Attestation],
         gossip_signatures: dict[SignatureKey, "Signature"] | None = None,
+        children_payloads: dict[SignatureKey, list[AggregatedSignatureProof]] | None = None,
     ) -> list[tuple[AggregatedAttestation, AggregatedSignatureProof]]:
         """
-        Collect aggregated signatures from gossip network and aggregate them.
+        Collect signatures from gossip network and children payloads to aggregate them.
 
         For each attestation group, attempt to collect individual XMSS signatures
-        from the gossip network. These are fresh signatures that validators
-        broadcast when they attest.
+        from the gossip network and children payloads. These are fresh signatures that validators
+        broadcast when they attest and both new and known children payloads.
 
         Parameters
         ----------
@@ -879,6 +880,8 @@ class State(Container):
             Individual attestations to aggregate and sign.
         gossip_signatures : dict[SignatureKey, Signature] | None
             Per-validator XMSS signatures learned from the gossip network.
+        children_payloads : dict[SignatureKey, list[AggregatedSignatureProof]] | None
+            Aggregated proofs, contains both new and known children payloads.
 
         Returns:
         -------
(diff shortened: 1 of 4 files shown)


src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -925,22 +928,39 @@ def aggregate_gossip_signatures(
...
-                participants = AggregationBits.from_validator_indices(
-                    ValidatorIndices(data=gossip_ids)
+            # Skip validators already covered by gossip
+            gossip_id_set = set(gossip_ids)
+            if children_payloads:
+                for vid in validator_ids:
+                    if vid in gossip_id_set:
+                        continue
+                    key = SignatureKey(vid, data_root)
+                    if (pls := children_payloads.get(key)) is not None:
```
g11tech, line 942: we need to get the one with the maximum coverage with the remaining validator ids and consume them in the greedy manner

g11tech: infact gossips one should be last to be added

g11tech: infact you should now not store by vid-dataroot key, just store payloads by dataroot key i.e. datarootke => list of payloads , list of validator ids

because its simpler to just take them in a greedy fashion and then pick the non covered validator  from validators and then do recursive aggregattion

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -827,19 +827,10 @@ def accept_new_attestations(self) -> "Store":
...
-                # Merge proof lists for the same signature key
-                merged_aggregated_payloads[sig_key] = merged_aggregated_payloads[sig_key] + proofs
-            else:
-                merged_aggregated_payloads[sig_key] = proofs
-
         # Create store with migrated aggregated payloads
         store = self.model_copy(
             update={
-                "latest_known_aggregated_payloads": merged_aggregated_payloads,
+                "latest_known_aggregated_payloads": self._merged_aggregated_payloads(),
```
g11tech, line 833:
```suggestion
                "latest_known_aggregated_payloads": self._merge_aggregated_payloads(),
```

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -951,37 +929,85 @@ def update_safe_target(self) -> "Store":
...
+        self,
+    ) -> dict[SignatureKey, list[AggregatedSignatureProof]]:
+        """Merge new and known aggregated payloads into a single view.
+
+        Starts with a shallow copy of known payloads, then overlays new payloads.
+        When both pools contain proofs for the same key, their lists are concatenated.
         """
-        Aggregate committee signatures for attestations in committee_signatures.
+        merged = dict(self.latest_known_aggregated_payloads)
+        for sig_key, proofs in self.latest_new_aggregated_payloads.items():
```
g11tech, line 941: the keys for aggregated payloads should now be just attestattion data right?

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -925,22 +928,39 @@ def aggregate_gossip_signatures(
...
-            # compact proof that can verify all participants signed the message.
-            if gossip_ids:
-                participants = AggregationBits.from_validator_indices(
-                    ValidatorIndices(data=gossip_ids)
+            # Skip validators already covered by gossip
+            gossip_id_set = set(gossip_ids)
+            if children_payloads:
+                for vid in validator_ids:
+                    if vid in gossip_id_set:
+                        continue
```
unnawut, line 940: nit: Can be collapsed to this. Also more readable:

```suggestion
                for vid in set(validator_ids) - set(gossip_ids):
```

src/lean_spec/subspecs/xmss/types.py
```diff
@@ -113,6 +113,16 @@ class HashTreeOpening(Container):
     """SSZ-compliant list of sibling hashes, from bottom to top."""
 
 
+class BytecodePointOption(SSZUnion):
```
unnawut, line 116: I remember vaguely that SSZUnion type is quite problematic(?, source needed but I think it was around determinism) and there's been proposals to remove it from specs (https://github.com/ethereum/consensus-specs/issues/4763, https://github.com/ethereum/consensus-specs/pull/3906). So I'm not sure we should consider this before adding SSZUnion type especially if we're trying to be more lean than current beacon specs.

Seeing [this match clause in AggregatedXMSS::public_input()](https://github.com/leanEthereum/leanMultisig/blob/main/crates/rec_aggregation/src/lib.rs#L120-L132) and with a bit of AI consultation it looks like replacing the optional with `vec![F::ZERO; bytecode_claim_size]` or similar could work? cc: @TomWambsgans

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -951,37 +929,85 @@ def update_safe_target(self) -> "Store":
...
-            attestation_data = self.attestation_data_by_root.get(data_root)
+        for sig_key in all_keys:
+            attestation_data = self.attestation_data_by_root.get(sig_key.data_root)
             if attestation_data is not None:
                 attestation_list.append(
                     Attestation(validator_id=sig_key.validator_id, data=attestation_data)
                 )
+        return attestation_list
+
+    def aggregate_committee_signatures_and_payloads(self) -> tuple["Store", list[SignedAggregatedAttestation]]:
```
unnawut, line 982: This function is currently unused. I think it should replace `store. aggregate_committee_signatures()` across the entire codebase right?

anshalshukla: right

src/lean_spec/subspecs/xmss/types.py
unnawut: Conclusion from pqinterop chat is we keep leanMultisig API the way it is (with `Option`) but work on leanSpec to avoid SSZUnion

PR #432, opened 2026-03-02
Add frequent node logging: attestations, blocks, justified/finalized, peers

ch4r10t33r:
Adds observability logging to the leanSpec node:

- **Blocks**: INFO when a block is received (peer, slot) and when processed (slot, root, peer)
- **Attestations**: INFO when received (peer, slot, validator); INFO on validation/signature success, WARNING on failure
- **Aggregated attestations**: Same pattern with optional peer_id
- **Periodic status** (every 10s): Peers count, justified slot/root, finalized slot/root

Sync service accepts optional `peer_id` on `on_gossip_attestation` and `on_gossip_aggregated_attestation`; network service passes peer_id from gossip events so logs show the propagating peer.

### DIFF
--- a/src/lean_spec/subspecs/node/node.py
+++ b/src/lean_spec/subspecs/node/node.py
@@ -470,8 +470,8 @@ class Node:
             )
             j = store.latest_justified
             f = store.latest_finalized
-            j_root = j.root.hex()[:16] + "…" if hasattr(j.root, "hex") else str(j.root)[:20]
-            f_root = f.root.hex()[:16] + "…" if hasattr(f.root, "hex") else str(f.root)[:20]
+            j_root = j.root.hex() if hasattr(j.root, "hex") else str(j.root)
+            f_root = f.root.hex() if hasattr(f.root, "hex") else str(f.root)
             logger.info("=" * 64)
             logger.info(
                 "Peers=%s | Justified slot=%s root=%s | Finalized slot=%s root=%s",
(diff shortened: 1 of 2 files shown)


src/lean_spec/subspecs/node/node.py
```diff
@@ -446,6 +453,34 @@ def _install_signal_handlers(self) -> None:
...
+            if self._shutdown.is_set():
+                break
+            store = self.sync_service.store
+            peers_connected = sum(
+                1 for p in self.sync_service.peer_manager.get_all_peers() if p.is_connected()
+            )
+            j = store.latest_justified
+            f = store.latest_finalized
+            j_root = j.root.hex()[:16] + "…" if hasattr(j.root, "hex") else str(j.root)[:20]
+            f_root = f.root.hex()[:16] + "…" if hasattr(f.root, "hex") else str(f.root)[:20]
```
tcoratger: Here why do we log only up to 16 or 20? Can't we just log full or is too long and verbose? I guess you have same concerns for the clients

src/lean_spec/subspecs/sync/service.py
```diff
@@ -440,6 +445,14 @@ async def on_gossip_block(
         #
         # A block may be cached instead of processed if its parent is unknown.
         if result.processed:
+            slot = block.message.block.slot
+            block_root = hash_tree_root(block.message.block)
+            logger.info(
+                "Block processed slot=%s root=%s from peer %s",
+                slot,
+                block_root.hex()[:16] + "…",
```
tcoratger: Same comment here and for all the PR in general. I remember seeing this at a couple of locations in the codebase and I always try to log things in full for clarity.

PR #438, opened 2026-03-07
interop acheived

anshalshukla:
## 🗒️ Description
- Updates `PRF_BYTES_PER_FE` to 16 which fixes the leaf mismatch at the time of verification with signatures generated from keys of lean-sig-cli
- p2p related changes: improved logging, topic name, enr keys, replay buffered events support

- I was able to achieve interop between spec-zeam in a 2 node devnet (with justifications, finalizations)

## ✅ Checklist

    ```console
    uvx tox
    ```


src/lean_spec/subspecs/networking/enr/enr.py
```diff
@@ -140,23 +140,23 @@ def ip6(self) -> str | None:
...
+        raw = self.get(keys.QUIC)
         return Port(int.from_bytes(raw, "big")) if raw else None
 
     @property
-    def udp6_port(self) -> Port | None:
-        """IPv6-specific UDP port. Falls back to udp_port if not set."""
-        raw = self.get(keys.UDP6)
+    def quic6_port(self) -> Port | None:
+        """IPv6-specific QUIC port."""
+        raw = self.get(keys.QUIC6)
```
KolbyML, line 163: We should be keeping udp ports and quic ports no? discv5 doesn't use quic etc

KolbyML: some clients might choose to multiplex both discv5 and quic on 1 udp port, but that would be an option rather than the rule

PR #446, opened 2026-03-10
Devnet 4

anshalshukla:
## 🗒️ Description
Implements recursive aggregation. Spec for devnet-4 (draft):

    - updates aggregation API to support recursive aggregation
    - INV_PROOF_SIZE is defined as a protocol parameter
    - changes aggregate_gossip_signatures to aggregate as it will no longer just aggregates gossip signatures but also recursively aggregates children_payloads with them.
    - changes aggregate_committee_signatures to aggregate_committee_signatures_and_payloads, it includes both committee gossip signatures and payloads (no committee) and recursively aggregate them using aggregate

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/src/lean_spec/subspecs/containers/state/state.py
+++ b/src/lean_spec/subspecs/containers/state/state.py
@@ -759,23 +759,27 @@ class State(Container):
 
         return final_block, post_state, aggregated_attestations, aggregated_signatures
 
-    def _extend_proofs_with_unique_participants(
+    def _extend_proofs_greedily(
             proofs: set[AggregatedSignatureProof] | None,
             selected: list[AggregatedSignatureProof],
             covered: set[ValidatorIndex],
         ) -> None:
             if not proofs:
                 return
-            sorted_proofs = sorted(
-                proofs,
-                key=lambda proof: len(proof.participants.to_validator_indices()),
-                reverse=True,
-            )
-            for proof in sorted_proofs:
-                participants = set(proof.participants.to_validator_indices())
-                if participants - covered:
-                    selected.append(proof)
-                    covered.update(participants)
+            remaining = list(proofs)
+            while remaining:
+                best = max(
+                    remaining,
+                    key=lambda proof: len(
+                        set(proof.participants.to_validator_indices()) - covered
+                    ),
+                )
+                participants = set(best.participants.to_validator_indices())
+                if not (participants - covered):
+                    break
+                selected.append(best)
+                covered.update(participants)
+                remaining.remove(best)
 
     def aggregate(
         self,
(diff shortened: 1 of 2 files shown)


src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -759,79 +759,89 @@ def build_block(
...
+    def _extend_proofs_with_unique_participants(
+            proofs: set[AggregatedSignatureProof] | None,
+            selected: list[AggregatedSignatureProof],
+            covered: set[ValidatorIndex],
+        ) -> None:
+            if not proofs:
+                return
+            sorted_proofs = sorted(
+                proofs,
+                key=lambda proof: len(proof.participants.to_validator_indices()),
```
g11tech: calc and sort them here with participants - covered directly, this is what really will give greedy coverage which imo would be the best coverage algo on average

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -946,65 +946,30 @@ def update_safe_target(self) -> "Store":
         # The head and attestation pools remain unchanged.
         return self.model_copy(update={"safe_target": safe_target})
 
-    def aggregate_committee_signatures(self) -> tuple["Store", list[SignedAggregatedAttestation]]:
+    def aggregate_committee_signatures_and_payloads(self) -> tuple["Store", list[SignedAggregatedAttestation]]:
```
g11tech:
```suggestion
    def aggregate_signatures_and_payloads(self) -> tuple["Store", list[SignedAggregatedAttestation]]:
```

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -759,79 +759,89 @@ def build_block(
...
+            )
+            for proof in sorted_proofs:
+                participants = set(proof.participants.to_validator_indices())
+                if participants - covered:
+                    selected.append(proof)
+                    covered.update(participants)
+
+    def aggregate(
         self,
-        attestations: Collection[Attestation],
```
g11tech, line 764: it needs to take attestation/attestation data as the params

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -759,79 +759,89 @@ def build_block(
...
+        gossip_signatures = gossip_signatures or {}
+        new_payloads = new_payloads or {}
+        known_payloads = known_payloads or {}
 
-            # Get the list of validators who attested to this data.
-            validator_ids = aggregated.aggregation_bits.to_validator_indices()
+        # Use only keys from new_payloads and gossip_signatures
+        # know_payloads can be used to extend the proof with new_payloads and gossip_signatures
+        # but known_payloads are not recursively aggregated into their own proofs
+        attestation_keys = set(new_payloads.keys()) | set(gossip_signatures.keys())
```
g11tech, line 810: new payloads are use for aggregator and it can extend with known payloads if the validator indices in attestations for a particular data is not covered.

and known payloads are used for block proposer which will already have new payloads merged in and shouldn't require a lookup in new payloads (which should be an error condition)

src/lean_spec/subspecs/containers/state/state.py
anshalshukla: We have `attestation_data` from `new_payloads` and `signatures_map`

src/lean_spec/subspecs/containers/state/state.py
anshalshukla: > new payloads are use for aggregator and it can extend with known payloads if the validator indices in attestations for a particular data is not covered.

Yeah, this function does exactly that, it is for aggregation of payloads and signatures for the aggregator.

> and known payloads are used for block proposer which will already have new payloads merged in and shouldn't require a lookup in new payloads (which should be an error condition)

I may not fully understand this, but known payloads refer to payloads that have already been accounted for and included in blocks. While performing aggregation, I use these known payloads if a particular attestation_data exists either in signatures_map or in new_payload, and the same payload is also present in known_payload and can contribute additional participant bits.

anshalshukla:
Adding these changes to #449

PR #449, opened 2026-03-11
separate attestation and proposal keys - Devnet4

anshalshukla:
## 🗒️ Description
- Separates attestation and proposal keys to avoid OTS issues of signing attestation with block as a proposer
- Updates aggregation API to support recursive attestation, restricted behind a flag as bindings are unavailable
- The `store.py` and `state.py` aggregate functions intentionally have redundant code and will be cleaned once the recursive aggregation is supported, right now for clarity each path is implemented separately although some portions can be shared

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
+++ b/packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
@@ -7,7 +7,6 @@ Validates Store responses to blocks, attestations, and time progression.
 
 from __future__ import annotations
 
-import logging
 from typing import ClassVar, Mapping, Self
 
 from pydantic import model_validator
@@ -198,19 +197,16 @@ class ForkChoiceTest(BaseConsensusFixture):
         # Test states use placeholder pubkeys.
         # We must replace them with the key manager's actual keys.
         # Otherwise signature verification will fail.
-        updated_validators = [
-            validator.model_copy(
+        updated_validators = []
+        for i, validator in enumerate(self.anchor_state.validators):
+            idx = ValidatorIndex(i)
+            validator = validator.model_copy(
                 update={
-                    "attestation_pubkey": key_manager[
-                        ValidatorIndex(i)
-                    ].attestation_public.encode_bytes(),
-                    "proposal_pubkey": key_manager[
-                        ValidatorIndex(i)
-                    ].proposal_public.encode_bytes(),
+                    "attestation_pubkey": key_manager[idx].attestation_public.encode_bytes(),
+                    "proposal_pubkey": key_manager[idx].proposal_public.encode_bytes(),
                 }
             )
-            for i, validator in enumerate(self.anchor_state.validators)
-        ]
+            updated_validators.append(validator)
 
         # Updating validators changes the state root.
         # We must also update the anchor block to match.
@@ -299,26 +295,18 @@ class ForkChoiceTest(BaseConsensusFixture):
                         if step.block.gossip_proposer_attestation:
                             proposer_index = block.proposer_index
                             proposer_att_data = store.produce_attestation_data(block.slot)
-                            try:
-                                proposer_gossip_att = SignedAttestation(
-                                    validator_id=proposer_index,
-                                    data=proposer_att_data,
-                                    signature=key_manager.sign_attestation_data(
-                                        proposer_index, proposer_att_data
-                                    ),
-                                )
-                                store = store.on_gossip_attestation(
-                                    proposer_gossip_att,
-                                    scheme=LEAN_ENV_TO_SCHEMES[self.lean_env],
-                                    is_aggregator=True,
-                                )
-                            except Exception as e:
-                                logging.warning(
-                                    "Proposer gossip attestation failed (slot=%s): %s",
-                                    block.slot,
-                                    e,
-                                    exc_info=True,
-                                )
+                            proposer_gossip_att = SignedAttestation(
+                                validator_id=proposer_index,
+                                data=proposer_att_data,
+                                signature=key_manager.sign_attestation_data(
+                                    proposer_index, proposer_att_data
+                                ),
+                            )
+                            store = store.on_gossip_attestation(
+                                proposer_gossip_att,
+                                scheme=LEAN_ENV_TO_SCHEMES[self.lean_env],
+                                is_aggregator=True,
+                            )
 
                     case AttestationStep():
                         # Process a gossip attestation.
--- a/src/lean_spec/subspecs/containers/state/state.py
+++ b/src/lean_spec/subspecs/containers/state/state.py
@@ -827,8 +827,6 @@ class State(Container):
                 ):
                     if entry.validator_id in covered_validators:
                         continue
-                    if int(entry.validator_id) >= len(self.validators):
-                        continue
                     public_key = self.validators[entry.validator_id].get_attestation_pubkey()
                     raw_entries.append((entry.validator_id, public_key, entry.signature))
                     covered_validators.add(entry.validator_id)
@@ -859,8 +857,6 @@ class State(Container):
                 for entry in sorted(
                     gossip_signatures.get(data, set()), key=lambda e: e.validator_id
                 ):
-                    if int(entry.validator_id) >= len(self.validators):
-                        continue
                     public_key = self.validators[entry.validator_id].get_attestation_pubkey()
                     raw_entries.append((entry.validator_id, public_key, entry.signature))
                 if not raw_entries:


packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -200,7 +199,14 @@ def make_fixture(self) -> Self:
...
         updated_validators = [
             validator.model_copy(
-                update={"pubkey": key_manager[ValidatorIndex(i)].public.encode_bytes()}
+                update={
+                    "attestation_pubkey": key_manager[
+                        ValidatorIndex(i)
+                    ].attestation_public.encode_bytes(),
+                    "proposal_pubkey": key_manager[
+                        ValidatorIndex(i)
+                    ].proposal_public.encode_bytes(),
```
tcoratger: I think that this is better to define a variable in order not to have two times `key_manager[ValidatorIndex(i)]`

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -273,21 +278,38 @@ def make_fixture(self) -> Self:
                             self._block_registry[step.block.label] = block
 
                         # Advance time to the block's slot.
-                        # Store rejects blocks from the future.
-                        # This tick includes a block (has proposal).
-                        # Always act as aggregator to ensure gossip signatures are aggregated
```
tcoratger, line 278: Why do we need to delete this doc?

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -258,12 +264,11 @@ def make_fixture(self) -> Self:
...
 
-                        # Store the filled Block for serialization
-                        block = signed_block.message.block
-                        step._filled_block = signed_block.message
+                        # Store the filled block for serialization.
+                        block = signed_block.message
+                        step._filled_block = block
 
                         # Register labeled blocks for fork building.
-                        # Later blocks can reference this one as their parent.
```
tcoratger, line 266: Why do we need to delete this doc?

packages/testing/src/consensus_testing/keys.py
```diff
@@ -282,20 +285,60 @@ def __iter__(self) -> Iterator[ValidatorIndex]:
...
-    def get_public_key(self, idx: ValidatorIndex) -> PublicKey:
-        """Get a validator's public key."""
-        return self[idx].public
+    def get_attestation_public_key(self, idx: ValidatorIndex) -> PublicKey:
+        """Get a validator's attestation public key."""
+        return self[idx].attestation_public
+
+    def get_proposal_public_key(self, idx: ValidatorIndex) -> PublicKey:
+        """Get a validator's proposal public key."""
+        return self[idx].proposal_public
```
tcoratger: I don't think that we need functions for this no? This makes the code more complex given the simplicity of the functions no? We can directly access the elements via `self[idx].attestation_public` and `self[idx].proposal_public`, what do you think?

packages/testing/src/consensus_testing/keys.py
```diff
@@ -282,20 +285,60 @@ def __iter__(self) -> Iterator[ValidatorIndex]:
...
+        prepared interval, caches the updated secret, and produces the signature.
+
+        Args:
+            validator_id: Validator index whose key should be used.
+            slot: The slot to sign for.
+            message: The message bytes to sign.
+            secret_field: Which secret on the key pair should advance.
+        """
+        kp = self[validator_id]
+        sk = getattr(kp, secret_field)
```
tcoratger, line 308: Should work no?

```suggestion
        sk = kp.secret_field
```

src/lean_spec/subspecs/containers/block/block.py
```diff
@@ -142,29 +130,21 @@ def verify_signatures(
         validators = parent_state.validators
 
         # Attestations and signatures are parallel arrays.
-        # - Each attestation says "validators X, Y, Z voted for this data".
-        # - Each signature proves those validators actually signed.
```
tcoratger, line 146: Why do we need to delete this doc?

src/lean_spec/subspecs/containers/block/block.py
```diff
@@ -142,29 +130,21 @@ def verify_signatures(
...
         for aggregated_attestation, aggregated_signature in zip(
             aggregated_attestations, attestation_signatures, strict=True
         ):
-            # Extract which validators participated in this attestation.
-            # The aggregation bits encode validator indices as a bitfield.
```
tcoratger, line 151: Why do we need to delete this doc?

src/lean_spec/subspecs/containers/block/block.py
```diff
@@ -142,29 +130,21 @@ def verify_signatures(
...
             validator_ids = aggregated_attestation.aggregation_bits.to_validator_indices()
 
             # The signed message is the attestation data root.
-            # All validators in this group signed this exact data.
```
tcoratger, line 155: Why do we need to delete this doc?

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -759,78 +758,126 @@ def build_block(
...
+    ) -> None:
+        if not proofs:
+            return
+        remaining = list(proofs)
+        while remaining:
+            best = max(
+                remaining,
+                key=lambda proof: len(set(proof.participants.to_validator_indices()) - covered),
+            )
+            participants = set(best.participants.to_validator_indices())
```
tcoratger, line 775:
```suggestion
            participants = set(best.participants.to_validator_indices())
            best = max(
                remaining,
                key=lambda proof: len(participants) - covered),
            )
            
```

src/lean_spec/subspecs/xmss/aggregation.py
```diff
@@ -44,39 +48,74 @@ class AggregatedSignatureProof(Container):
...
+    # If the bytecode point is not provided, the proof is not recursive.
+    # """
+
     @classmethod
     def aggregate(
         cls,
-        participants: AggregationBits,
-        public_keys: Sequence[PublicKey],
-        signatures: Sequence[Signature],
+        xmss_participants: AggregationBits | None,
```
tcoratger, line 62: Why are we calling these `xmss_participants`, this is better simply `participants` no?

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -581,43 +556,11 @@ def on_block(
...
-            update={
-                "latest_known_aggregated_payloads": block_proofs,
-            }
-        )
+        store = store.model_copy(update={"latest_known_aggregated_payloads": block_proofs})
 
         # Update forkchoice head based on new block and attestations
-        #
-        # IMPORTANT: This must happen BEFORE processing proposer attestation
-        # to prevent the proposer from gaining circular weight advantage.
```
tcoratger, line 593: Why do we need to delete this doc?

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -581,43 +556,11 @@ def on_block(
...
-                ATTESTATION_COMMITTEE_COUNT
-            )
-            if proposer_subnet_id == current_validator_subnet_id:
-                sig = signed_block_with_attestation.signature.proposer_signature
-                new_gossip_sigs.setdefault(proposer_attestation.data, set()).add(
-                    GossipSignatureEntry(proposer_attestation.validator_id, sig)
-                )
-
-        # Update store with proposer signature
-        store = store.model_copy(update={"gossip_signatures": new_gossip_sigs})
```
tcoratger, line 619: The old `on_block()` processed the proposer's attestation into `gossip_signatures` after updating the head. The new `on_block()` does not — it only processes body attestations into `latest_known_aggregated_payloads`. The proposer attestation is now gossipped separately (via `gossip_proposer_attestation` flag in test fixtures).

This means if gossip fails or is delayed, the proposer's attestation contributes zero weight to fork choice. In the old design, it was guaranteed to be processed. Are we sure that is normal?

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -273,21 +278,38 @@ def make_fixture(self) -> Self:
...
+                                        proposer_index, proposer_att_data
+                                    ),
+                                )
+                                store = store.on_gossip_attestation(
+                                    proposer_gossip_att,
+                                    scheme=LEAN_ENV_TO_SCHEMES[self.lean_env],
+                                    is_aggregator=True,
+                                )
+                            except (AssertionError, Exception):
+                                pass
```
tcoratger: I think that this is redundant and too broad. `AssertionError` is a subclass of `Exception`, so listing both is redundant.

More importantly, bare except `Exception: pass` silently swallows OTS reuse errors, serialization bugs, and state corruption. This should at minimum log a warning no?

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -759,78 +758,126 @@ def build_block(
...
+            if not (participants - covered):
+                break
+            selected.append(best)
+            covered.update(participants)
+            remaining.remove(best)
```
tcoratger, line 780: Here we have a lot of mutability, not sure this is the best from a formal verification point of view, in the rest of the code base, that is common to avoid mutable stuffs and prefer copies.

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -759,78 +758,126 @@ def build_block(
...
+
+    def aggregate(
```
tcoratger, line 782: It seems that this function shares a lot of common logic with the `Store` aggregate method no? Are we sure this is the best way to go and why not having a single `aggregate` method?

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
anshalshukla: I've restored most of the doc, I removed some of them that looked obvious or misleading after the changes

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
anshalshukla: Restored

src/lean_spec/subspecs/xmss/aggregation.py
anshalshukla: Because it will contain participants related to raw_xmss and not children proofs. `participants` can look misleading as to include participants from the child proofs as well.

src/lean_spec/subspecs/forkchoice/store.py
anshalshukla: I think this can be misleading, I'm not treating proposer attestation differently anymore and it will be created post block production

src/lean_spec/subspecs/containers/state/state.py
anshalshukla: I think I can further simplify it once the bindings are there and merge the function into one

src/lean_spec/subspecs/containers/state/state.py
anshalshukla: `best` is defined later, I can move it outside the loop but right now it's basically dead code which I intent to merge into aggregate when the bindings are available so can simplify it then

packages/testing/src/consensus_testing/keys.py
anshalshukla: secret_field is not a field in `ValidatorKeyPair` but a Literal taken in as function argument.

src/lean_spec/subspecs/forkchoice/store.py
anshalshukla: We process attestation locally after signing within the `_produce_attestations` function and add it to the `attestation_signatures` map. It's not dependent on gossip that way.

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -282,12 +289,37 @@ def make_fixture(self) -> Self:
...
+                                    is_aggregator=True,
+                                )
+                            except Exception as e:
+                                logging.warning(
+                                    "Proposer gossip attestation failed (slot=%s): %s",
+                                    block.slot,
+                                    e,
+                                    exc_info=True,
+                                )
+
```
unnawut: Since this is part of test setup, should we just let the exception bubble up?

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -759,78 +758,126 @@ def build_block(
...
+                    known_payloads.get(data), child_proofs, covered_validators
+                )
+                raw_entries = []
+                for entry in sorted(
+                    gossip_signatures.get(data, set()), key=lambda e: e.validator_id
+                ):
+                    if entry.validator_id in covered_validators:
+                        continue
+                    if int(entry.validator_id) >= len(self.validators):
+                        continue
```
unnawut: Is this scenario possible? If so should we eject this attestation closer to the networking layer than at aggregation?

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
tcoratger: @anshalshukla This one is not fixed no?

packages/testing/src/consensus_testing/keys.py
tcoratger: Ok we should handle this in a followup no problem.

src/lean_spec/subspecs/containers/state/state.py
anshalshukla: It is possible because anyone can send an attestation in p2p, malicious participants cannot be stopped. We already validate it in networking layer in `on_gossip_attestation`, so removing this checks from other places.

PR #464, opened 2026-03-21
simplify block production

anshalshukla:
## 🗒️ Description
Simplies block production. Uses a more efficient and simpler approach to build blocks.
- Earlier we were calculating attestations from the aggregated_payloads, these attestations were applied on the state until no more attestations can be applied
- With the change in signature and payload maps we don't need to fetch attestations and can directly look at the attestation_data based on the current_justified checkpoint, select proofs greedily and apply them as long as the current_justified checkpoint continues to update

Backwards compatible as it doesn't change anything in state or stf

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/src/lean_spec/subspecs/containers/state/state.py
+++ b/src/lean_spec/subspecs/containers/state/state.py
@@ -675,6 +675,8 @@ class State(Container):
             else:
                 current_justified = self.latest_justified
 
+            processed_att_data: set[AttestationData] = set()
+
             while True:
                 found_entries = False
 
@@ -687,33 +689,20 @@ class State(Container):
                     if att_data.source != current_justified:
                         continue
 
+                    if att_data in processed_att_data:
+                        continue
+                    processed_att_data.add(att_data)
+
                     found_entries = True
 
-                    # Greedy proof selection: pick the proof covering the most
-                    # uncovered validators until all are covered.
                     covered: set[ValidatorIndex] = set()
-                    while True:
-                        best_proof: AggregatedSignatureProof | None = None
-                        best_new_coverage = 0
-
-                        for proof in proofs:
-                            proof_validators = set(proof.participants.to_validator_indices())
-                            new_coverage = len(proof_validators - covered)
-                            if new_coverage > best_new_coverage:
-                                best_new_coverage = new_coverage
-                                best_proof = proof
-
-                        if best_proof is None or best_new_coverage == 0:
-                            break
-
-                        aggregated_attestations.append(
-                            AggregatedAttestation(
-                                aggregation_bits=best_proof.participants,
-                                data=att_data,
-                            )
-                        )
-                        aggregated_signatures.append(best_proof)
-                        covered |= set(best_proof.participants.to_validator_indices())
+                    self._extend_proofs_greedily(
+                        proofs,
+                        aggregated_signatures,
+                        covered,
+                        attestation_data=att_data,
+                        attestations=aggregated_attestations,
+                    )
 
                 if not found_entries:
                     break
@@ -758,6 +747,8 @@ class State(Container):
         proofs: set[AggregatedSignatureProof] | None,
         selected: list[AggregatedSignatureProof],
         covered: set[ValidatorIndex],
+        attestation_data: AttestationData | None = None,
+        attestations: list[AggregatedAttestation] | None = None,
     ) -> None:
         if not proofs:
             return
@@ -773,6 +764,10 @@ class State(Container):
             selected.append(best)
             covered.update(participants)
             remaining.remove(best)
+            if attestation_data is not None and attestations is not None:
+                attestations.append(
+                    AggregatedAttestation(aggregation_bits=best.participants, data=attestation_data)
+                )
 
     def aggregate(
         self,
--- a/src/lean_spec/subspecs/forkchoice/store.py
+++ b/src/lean_spec/subspecs/forkchoice/store.py
@@ -960,7 +960,7 @@ class Store(StrictBaseModel):
                 }
             ), new_aggregates
         else:
-            # Plain aggregation: only attestation_signatures. Remove this block once
+            # Plain aggregation: only attestation_signatures. TODO: Remove this block once
             # bindings support recursive aggregation.
             #
             # Freshly aggregated proofs go directly to latest_known because they are
--- a/tests/lean_spec/subspecs/containers/test_state_aggregation.py
+++ b/tests/lean_spec/subspecs/containers/test_state_aggregation.py
@@ -270,6 +270,286 @@ def test_build_block_state_root_valid_when_signatures_split(
     assert len(result_state.validators.data) == num_validators
 
 
+def test_build_block_greedy_selects_minimum_proofs(
+    container_key_manager: XmssKeyManager,
+) -> None:
+    """Greedy selection picks the minimum set of proofs to cover all validators."""
+    state = make_keyed_genesis_state(4, container_key_manager)
+    parent_header_with_state_root = state.latest_block_header.model_copy(
+        update={"state_root": hash_tree_root(state)}
+    )
+    parent_root = hash_tree_root(parent_header_with_state_root)
+    source = Checkpoint(root=parent_root, slot=Slot(0))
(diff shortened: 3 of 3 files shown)


packages/testing/src/consensus_testing/test_types/store_checks.py
```diff
@@ -593,16 +593,6 @@ def validate_against_store(
...
 
-                # Verify all forks are at the same slot
-                slots = [slot for _, slot, _ in fork_data.values()]
-                if len(set(slots)) > 1:
-                    slot_info = {label: slot for label, (_, slot, _) in fork_data.items()}
-                    raise AssertionError(
-                        f"Step {step_index}: lexicographic_head_among forks have "
-                        f"different slots: {slot_info}. All forks must be at the same "
-                        f"slot to test tiebreaker."
-                    )
```
tcoratger, line 604: Why do we need to remove this? Maybe I missed something?

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -968,8 +968,13 @@ def aggregate(
                 }
             ), new_aggregates
         else:
-            # Plain aggregation: only attestation_signatures; carry forward existing
-            # new payloads. Remove this block once bindings support recursive aggregation.
+            # Plain aggregation: only attestation_signatures. Remove this block once
+            # bindings support recursive aggregation.
```
tcoratger: nit: just for us to track and be searchable via a `TODO`

```suggestion
            # Plain aggregation: only attestation_signatures. TODO: Remove this block once
            # bindings support recursive aggregation.
```

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -634,121 +637,113 @@ def build_block(
...
-            new_attestations: list[Attestation] = []
-
-            for attestation in available_attestations:
-                data = attestation.data
-
-                # Skip if target block is unknown.
-                if data.head.root not in known_block_roots:
-                    continue
+        aggregated_attestations: list[AggregatedAttestation] = []
+        aggregated_signatures: list[AggregatedSignatureProof] = []
```
tcoratger, line 664: Not sure this is normal that these two lists are initialized once before the outer while True loop, since they are accumulated across iterations.

When justification advances and the loop re-enters:
- The covered set is reset per `att_data` on each outer iteration
- Previously selected proofs remain in the accumulator lists which looks weird
- If the same `att_data` matches both the old and new justified checkpoint, its proofs get selected again, creating duplicate attestations in the final block

I think that we should track processed entries with `processed_att_data: set[AttestationData]` and skip already-handled data across iterations.

What do you think?

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -634,121 +637,113 @@ def build_block(
...
+                    found_entries = True
+
+                    # Greedy proof selection: pick the proof covering the most
+                    # uncovered validators until all are covered.
+                    covered: set[ValidatorIndex] = set()
+                    while True:
+                        best_proof: AggregatedSignatureProof | None = None
+                        best_new_coverage = 0
+
+                        for proof in proofs:
```
tcoratger: Here it seems super complex to follow no? With 4 levels of nesting in total, couldn't we simplify? I'm afraid that this is cause a lot of trouble to implement by clients?

```
  while True:                                  # fixed-point
      for att_data, proofs in sorted(...):     # iterate attestation data
          while True:                          # greedy selection
              for proof in proofs:             # scan candidates
```

tests/lean_spec/subspecs/containers/test_state_aggregation.py
tcoratger, line 1: I understand that most of the methods tested in this file no longer exist, but it seems we've lost a lot of test coverage in the process, right? With potentially tricky scenarios that are no longer covered.

packages/testing/src/consensus_testing/test_types/store_checks.py
tcoratger, line 604: Why do we need to delete this?

tests/consensus/devnet/fc/test_fork_choice_head.py
```diff
@@ -382,56 +382,55 @@ def test_head_with_deep_fork_split(
             ),
             # Fork B: slots 2, 3, 4, 5 with more attestations to overtake
```
tcoratger, line 383: This comment looks outdated but I don't fully understand why we need to change slot stuffs here since the process is supposed to be exactly the same as before no?

packages/testing/src/consensus_testing/test_types/store_checks.py
anshalshukla: This should not be a necessary condition for lexicographical comparison. We have multiple blocks at same slot in various testcases but if there are multiple blocks at same slot it is either invalid proposal or double proposal, I've changed the testcase as well as it becomes hard to track store weights when multiple blocks exists for a particular slot.

tests/consensus/devnet/fc/test_fork_choice_head.py
anshalshukla: yeah the process is same, it was failing sporadically. Again as I mentioned earlier, I don't particularly like the idea of having multiple blocks at the same slot as it is not an ideal scenario, I think I had raised it earlier as well but we should update fork choice test cases to limit 1 block per slot. We will eventually have equivocation logic in place anyway, so we would need to remove this behavior regardless.

tests/lean_spec/subspecs/containers/test_state_aggregation.py
anshalshukla: I'll see what coverage is missing now and get them back

src/lean_spec/subspecs/containers/state/state.py
anshalshukla: Makes sense, I'll look into this

src/lean_spec/subspecs/containers/state/state.py
anshalshukla: I'll look into this as well later today

tests/lean_spec/subspecs/containers/test_state_aggregation.py
anshalshukla: I've added few of them back, I'm not much concerned about this as with recursive aggregation I can again look at the coverage and fill the gaps

packages/testing/src/consensus_testing/test_types/store_checks.py
unnawut: Not directly related to these comments but just a quick note that multiple blocks, same slot should not happen in the test cases as those test cases definitely are not for testing adversary cases and neither does our specs. 

I've also fixed some of them in https://github.com/leanEthereum/leanSpec/pull/461 but not all. I've added the rest as an issue at https://github.com/leanEthereum/leanSpec/issues/470

src/lean_spec/subspecs/containers/state/state.py
anshalshukla: This was addressed in the last commit, resolving it

src/lean_spec/subspecs/containers/state/state.py
anshalshukla: This issue was also addressed through _extend_proofs_greedily. Strictly speaking, this still makes the process four loops, since _extend_proofs_greedily itself contains two loops. Previously, the flow had three loops: fixed point, selection, and scanning.

Greedy selection introduces one additional loop, although we can remove it if client teams find it too burdensome to implement.

Looking ahead, once recursive aggregation is available, I plan to aggregate all proofs for a given message into a single proof during block production, even when the validator is not an aggregator. However, if this is done without care, proposers may miss the production interval. To prevent that, we would need an upper bound on the number of signatures/proofs aggregated within a production interval.

Under such a bound, greedy selection appears to be the best strategy for minimizing the number of proofs. My current understanding of LeanVM suggests that recursively aggregating 3 signatures may perform significantly better than aggregating 5, though this remains a hypothesis. It is possible that these gains would be offset by the cost of the extra loop. For that reason, we may choose not to require greedy selection in the spec, and instead allow client teams to implement whichever approach they prefer, anyway the spec should prioritize readability over optimization.

PR #482, opened 2026-03-25
node, networking, forkchoice: import-subnet-ids for subnet network topology

ch4r10t33r:
## Summary

- Add `import_subnet_ids` to control which attestation subnets a node subscribes to and collects from, independently of the aggregator role.
- **Subscription (p2p layer)**: aggregators subscribe to their validator-derived subnet; non-aggregators with no import subnets skip attestation subscriptions entirely (saves bandwidth); `import_subnet_ids` are always subscribed regardless of aggregator flag (for proposer nodes that need attestations for block inclusion); the two sets are additive and deduplicated.
- **Signature collection (forkchoice layer)**: existing aggregator path collects from validators sharing the same subnet; new import-subnet path always collects from explicitly listed subnets regardless of `is_aggregator`.
- Adds `--import-subnet-ids` CLI flag (comma-separated, e.g. `"0,1,2"`) wired through `run_node`, `_init_from_genesis`, `_init_from_checkpoint`, `NodeConfig`, `SyncService`, and `NetworkService`.

## Test plan

- [ ] `uvx tox -e all-checks` passes (ruff, ty, codespell, mdformat)
- [ ] 3 new tests in `TestOnGossipAttestationSubnetFiltering`:
  - `test_import_subnet_stores_without_aggregator` — import subnets work without `is_aggregator`
  - `test_import_subnet_ignored_for_other_subnets` — only explicitly listed subnets are collected
  - `test_import_subnet_combined_with_aggregator` — both paths are additive when used together
- [ ] All 7 tests in `TestOnGossipAttestationSubnetFiltering` pass

### DIFF
--- a/src/lean_spec/__main__.py
+++ b/src/lean_spec/__main__.py
@@ -456,9 +456,6 @@ async def run_node(
     if is_aggregator:
         logger.info("Aggregator mode enabled - node will perform attestation aggregation")
     if aggregate_subnet_ids:
-        if not is_aggregator:
-            logger.error("--aggregate-subnet-ids requires --is-aggregator to be set")
-            sys.exit(1)
         logger.info("Aggregate subnet IDs configured: %s", list(aggregate_subnet_ids))
 
     # Load validator keys if path provided.
@@ -747,8 +744,12 @@ def main() -> None:
     setup_logging(args.verbose, args.no_color)
 
     # Parse --aggregate-subnet-ids from comma-separated string to tuple of SubnetId.
+    # Reject the flag upfront if --is-aggregator is not set.
     aggregate_subnet_ids: tuple[SubnetId, ...] = ()
     if args.aggregate_subnet_ids:
+        if not args.is_aggregator:
+            logger.error("--aggregate-subnet-ids requires --is-aggregator to be set")
+            sys.exit(1)
         try:
             aggregate_subnet_ids = tuple(
                 SubnetId(int(s.strip())) for s in args.aggregate_subnet_ids.split(",") if s.strip()


src/lean_spec/__main__.py
```diff
@@ -494,18 +508,44 @@ async def run_node(
...
+    # Subscription rules:
+    # - All nodes with registered validators subscribe to their validator-derived subnets.
+    #   This forms the mesh network for attestation propagation.
+    # - Aggregator nodes additionally subscribe to any explicit aggregate-subnet-ids.
+    # - Non-aggregator nodes with no validators skip attestation subscriptions entirely.
+    subscription_subnets: set[SubnetId] = set()
+
+    # Always subscribe to validator-derived subnets regardless of aggregator flag.
+    if validator_registry is not None:
+        validator_id = validator_registry.primary_index()
```
anshalshukla: I think we need to loop over through all the registered validators and subscribe to subnets of each of them, on top of that to the subnets mentioned in aggregate_subnet_ids

src/lean_spec/__main__.py
```diff
@@ -494,18 +508,44 @@ async def run_node(
...
+            subscription_subnets.add(validator_id.compute_subnet_id(ATTESTATION_COMMITTEE_COUNT))
+
+    # Explicit aggregate subnets are additive but only meaningful for aggregators.
+    if is_aggregator:
+        if not subscription_subnets:
+            # Aggregator with no registered validators — fall back to subnet 0.
+            subscription_subnets.add(SubnetId(0))
+        subscription_subnets.update(aggregate_subnet_ids)
+
+    for subnet_id in sorted(subscription_subnets):
```
anshalshukla: sorting is not really required

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -373,14 +373,20 @@ def on_gossip_attestation(
...
             assert self.validator_id is not None, "Current validator ID must be set for aggregation"
             current_subnet = self.validator_id.compute_subnet_id(ATTESTATION_COMMITTEE_COUNT)
-            attester_subnet = validator_id.compute_subnet_id(ATTESTATION_COMMITTEE_COUNT)
-            if current_subnet == attester_subnet:
-                new_committee_sigs.setdefault(attestation_data, set()).add(
-                    AttestationSignatureEntry(validator_id, signature)
-                )
+            should_store = current_subnet == attester_subnet
+
+        if should_store:
```
anshalshukla: if we are subscribing to only the subnets that we want to import from then do we need this check? Anything that reaches here can be directly imported in case of aggregator nodes

src/lean_spec/__main__.py
```diff
@@ -443,6 +455,8 @@ async def run_node(
     # Log aggregator mode if enabled
     if is_aggregator:
         logger.info("Aggregator mode enabled - node will perform attestation aggregation")
+    if aggregate_subnet_ids:
```
anshalshukla, line 458: I think we should return an error in case this is set and is_aggregator is not

src/lean_spec/__main__.py
```diff
@@ -690,6 +743,20 @@ def main() -> None:
 
     setup_logging(args.verbose, args.no_color)
 
+    # Parse --aggregate-subnet-ids from comma-separated string to tuple of SubnetId.
+    aggregate_subnet_ids: tuple[SubnetId, ...] = ()
+    if args.aggregate_subnet_ids:
```
anshalshukla, line 749: i think here also we should check is_aggregator before trying to parse

src/lean_spec/__main__.py
```diff
@@ -494,18 +508,44 @@ async def run_node(
...
+
+    for subnet_id in subscription_subnets:
+        attestation_subnet_topic = GossipTopic.attestation_subnet(
+            GOSSIP_FORK_DIGEST, subnet_id
+        ).to_topic_id()
+        event_source.subscribe_gossip_topic(attestation_subnet_topic)
+        logger.info("Subscribed to attestation subnet %d", subnet_id)
+
+    if not subscription_subnets:
+        logger.info("Not subscribing to any attestation subnet")
```
anshalshukla, line 546: nit: since we have a fallback 0 subnet I think this is not necessary

anshalshukla: nvm, it is only for aggregator roles

PR #479, opened 2026-03-25
Add test vectors for store.on_gossip_attestation()

Smartdevs17:
## Summary
- Add test vectors for `store.on_gossip_attestation()` covering all validation rules in `validate_attestation()`
- New `GossipAttestationSpec` type for specifying gossip attestations in tests
- 12 test cases covering:
  - Valid attestation happy path
  - Checkpoint slot mismatch errors (source, target, head)
  - Topology violations (source > target, head < target)
  - Time validation (too far in future)
  - Signature verification
  - Validator lookup
  - Multiple validators and chain extension

## Files Changed
- `packages/testing/src/consensus_testing/test_types/gossip_attestation_spec.py` - New spec type
- `packages/testing/src/consensus_testing/test_types/step_types.py` - Updated AttestationStep
- `packages/testing/src/consensus_testing/test_fixtures/fork_choice.py` - Added attestation builder
- `packages/testing/src/consensus_testing/__init__.py` - Export new type
- `tests/consensus/devnet/fc/test_gossip_attestation_validation.py` - 12 test cases

## Test Coverage
All validation rules in `validate_attestation()`:
1. Availability check (source, target, head blocks exist)
2. Topology check (source slot ≤ target slot ≤ head slot)
3. Consistency check (checkpoint slots match block slots)
4. Time check (not too far in future)

---

**cc @unnawut @tcoratger**

Closes #472

PR #484, opened 2026-03-25
test: refactor test fixture and more tests

tcoratger:
## 🗒️ Description

Followup PR on top of https://github.com/leanEthereum/leanSpec/pull/479

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/tests/consensus/devnet/fc/test_gossip_attestation_validation.py
+++ b/tests/consensus/devnet/fc/test_gossip_attestation_validation.py
@@ -399,7 +399,7 @@ def test_attestation_head_older_than_target_rejected(
     )
 
 
-def test_attestation_source_slot_exceeds_target_rejected(
+def test_attestation_source_slot_override_exceeds_target_rejected(
     fork_choice_test: ForkChoiceTestFiller,
 ) -> None:
     """


tests/consensus/devnet/fc/test_gossip_attestation_validation.py
```diff
@@ -522,3 +571,127 @@ def test_gossip_attestation_chain_extended_after_gossip(
...
+                    target_slot=Slot(2),
+                    target_root_label="block_2",
+                    source_root_override=FAKE_ROOT,
+                    valid_signature=False,
+                ),
+                valid=False,
+                expected_error="Unknown source block",
+            ),
+        ],
+    )
```
tcoratger, line 697: Here I've identified attestation validations paths that were not tested!

tests/consensus/devnet/fc/test_gossip_attestation_validation.py
```diff
@@ -86,6 +87,7 @@ def test_attestation_target_slot_mismatch_rejected(
                     target_root_label="block_2",
                 ),
                 valid=False,
+                expected_error="Target checkpoint slot mismatch",
```
tcoratger, line 90: I think this is nice to check the error message we expect to be sure that we are testing the right error and not another one that we don't expect to happen here.

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -645,85 +676,117 @@ def _build_signed_attestation_from_spec(
...
+            or spec.target_root_override is not None
+            or spec.head_root_override is not None
+            or spec.source_root_override is not None
+        )
+
+        if not expected_valid or has_overrides:
+            attestation_data = self._build_overridden_attestation_data(spec, block_registry)
+        else:
+            # Reuse the spec for honest attestations.
+            attestation_data = store.produce_attestation_data(spec.slot)
```
tcoratger, line 720: Here I think that it is better to reuse the spec method directly.

tests/consensus/devnet/fc/test_gossip_attestation_validation.py
```diff
@@ -387,6 +393,46 @@ def test_attestation_head_older_than_target_rejected(
                     head_root_label="block_1",
                 ),
                 valid=False,
+                expected_error="Head checkpoint must not be older than target",
+            ),
+        ],
+    )
+
+
+def test_attestation_source_slot_exceeds_target_via_override_rejected(
```
unnawut: not sure if via_override is necessary it seems like test impl detail to me

```suggestion
def test_attestation_source_slot_exceeds_target_rejected(
```

PR #485, opened 2026-03-25
testing: rearrange keys file

tcoratger:
## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/packages/testing/src/consensus_testing/keys.py
+++ b/packages/testing/src/consensus_testing/keys.py
@@ -565,7 +565,8 @@ def _generate_single_keypair(
 
     # Generate two independent key pairs: one for attestations, one for proposals.
     #
-    # Separate keys allow signing both roles within the same slot without exhausting a one-time leaf.
+    # Separate keys allow signing both roles within the same slot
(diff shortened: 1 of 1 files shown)


packages/testing/src/consensus_testing/keys.py
```diff
@@ -530,24 +490,15 @@ def _generate_keys(lean_env: str, count: int, max_slot: int) -> None:
         f"({num_slots} slots) using {num_workers} cores..."
     )
 
-    with ProcessPoolExecutor(max_workers=num_workers) as executor:
-        worker_func = partial(_generate_single_keypair, scheme, num_slots)
-        key_pairs = list(executor.map(worker_func, range(count)))
-
-    # Create keys directory (remove old one if it exists)
-    if keys_dir.exists():
-        shutil.rmtree(keys_dir)
```
unnawut, line 539: We don't delete old keys anymore right? There could be case where stale keys from previous run might hang around if we generate less keys than the previous run. Edge case though.

tcoratger: Fixed

PR #487, opened 2026-03-26
fix: validate gossip aggregated attestations before storing

### LINKED ISSUE #473: on_gossip_aggregated_attestation() is not validating attestation rules?
I'm suspecting that [`on_gossip_aggregated_attestation()`](https://github.com/leanEthereum/leanSpec/blob/0cd62a3d4b9cfc3e38561102f6902ef0aabe51f3/src/lean_spec/subspecs/forkchoice/store.py#L392-L458) is validating only the signatures against the expected pubkeys, but ignoring all attestation validity checks.

It should be validating against [`validate_attestation()`](https://github.com/leanEthereum/leanSpec/blob/0cd62a3d4b9cfc3e38561102f6902ef0aabe51f3/src/lean_spec/subspecs/forkchoice/store.py#L272-L320) like [`on_gosssip_attestation()` does](https://github.com/leanEthereum/leanSpec/blob/0cd62a3d4b9cfc3e38561102f6902ef0aabe51f3/src/lean_spec/subspecs/forkchoice/store.py#L356).

shaaibu7:
## 🗒️ Description

 - on_gossip_aggregated_attestation() now validates the aggregated data via validate_attestation() before validator lookup, leanVM proof verification, or payload storage, ensuring the gossip aggregate path enforces the same availability/topology/time invariants as individual attestations.
 
  - Added test_invalid_attestation_data_rejected to prove aggregated attestations that violate validate_attestation() (e.g., source.slot > target.slot) raise before latest_new_aggregated_payloads is mutated.

  ## ✅ Checklist

fixes #473

### DIFF
--- a/tests/consensus/devnet/fc/test_gossip_aggregated_attestation_validation.py
+++ b/tests/consensus/devnet/fc/test_gossip_aggregated_attestation_validation.py
@@ -123,7 +123,7 @@ def test_aggregated_attestation_head_slot_mismatch_rejected(
                     target_slot=Slot(2),
                     target_root_label="block_2",
                     head_root_label="block_1",
-                    head_slot=Slot(1),
+                    head_slot=Slot(5),
                 ),
                 valid=False,
                 expected_error="Head checkpoint slot mismatch",


tests/lean_spec/subspecs/forkchoice/test_store_attestations.py
```diff
@@ -396,6 +396,31 @@ def test_invalid_proof_rejected(self, key_manager: XmssKeyManager) -> None:
...
+    def test_invalid_attestation_data_rejected(self, key_manager: XmssKeyManager) -> None:
+        """
+        Aggregated attestations fail if their data violates validate_attestation().
+        """
+        store, attestation_data = make_store_with_attestation_data(
+            key_manager, num_validators=4, validator_id=ValidatorIndex(0)
+        )
+
+        invalid_source = attestation_data.source.model_copy(
+            update={"slot": Slot(int(attestation_data.source.slot) + 1)}
```
unnawut: This would be more accurate of the error message "Source checkpoint slot must not exceed target"

```suggestion
            update={"slot": Slot(int(attestation_data.target.slot) + 1)}
```

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -414,6 +414,8 @@ def on_gossip_aggregated_attestation(
         data = signed_attestation.data
         proof = signed_attestation.proof
 
+        self.validate_attestation(Attestation(validator_id=ValidatorIndex(0), data=data))
```
unnawut: We shouldn't be assuming the `validator_id` value in the spec here.

We can edit `validate_attestation()` so that it takes in `AttestationData` instead of `Attestation` since it's not using the validator_id anyway.

tests/consensus/devnet/fc/test_gossip_aggregated_attestation_validation.py
```diff
@@ -0,0 +1,160 @@
...
+    return [
+        BlockStep(
+            block=BlockSpec(slot=Slot(1), label="block_1"),
+            checks=StoreChecks(head_slot=Slot(1)),
+        ),
+        BlockStep(
+            block=BlockSpec(slot=Slot(2), label="block_2"),
+            checks=StoreChecks(head_slot=Slot(2)),
+        ),
+    ]
```
tcoratger: I think this is more readable for the reader is you put this directly into each test without a util function like this, even if this is more verbose, we love readability for the tests.

tests/consensus/devnet/fc/test_gossip_aggregated_attestation_validation.py
```diff
@@ -0,0 +1,160 @@
...
+                validator_ids=[ValidatorIndex(1)],
+                slot=Slot(2),
+                target_slot=Slot(2),
+                target_root_label="block_2",
+            ),
+            checks=StoreChecks(head_slot=Slot(2)),
+        ),
+    ]
+
+    fork_choice_test(steps=steps)
```
tcoratger: No need for this, see the other tests, you can do directly something like this (and this should be the same for all the tests below):

```suggestion
    fork_choice_test(steps=
    [
        [
        BlockStep(
            block=BlockSpec(slot=Slot(1), label="block_1"),
            checks=StoreChecks(head_slot=Slot(1)),
        ),
        BlockStep(
            block=BlockSpec(slot=Slot(2), label="block_2"),
            checks=StoreChecks(head_slot=Slot(2)),
        ),
    ],
        GossipAggregatedAttestationStep(
            attestation=GossipAggregatedAttestationSpec(
                validator_ids=[ValidatorIndex(1)],
                slot=Slot(2),
                target_slot=Slot(2),
                target_root_label="block_2",
            ),
            checks=StoreChecks(head_slot=Slot(2)),
        ),
    ])
```

tests/consensus/devnet/fc/test_gossip_aggregated_attestation_validation.py
```diff
@@ -0,0 +1,160 @@
...
+        GossipAggregatedAttestationStep(
+            attestation=GossipAggregatedAttestationSpec(
+                validator_ids=[ValidatorIndex(1)],
+                slot=Slot(2),
+                target_slot=Slot(2),
+                target_root_label="block_2",
+                source_root=Bytes32(b"\xff" * 32),
+                source_slot=Slot(999),
+            ),
+            valid=False,
```
tcoratger: See https://github.com/leanEthereum/leanSpec/pull/484#discussion_r2989947590, I think this is nice to add the expected failure message that we get to be sure that we are testing the right thing and that the test is not failing for something else that we don't expect to happen.

tests/lean_spec/subspecs/forkchoice/test_store_attestations.py
```diff
@@ -396,6 +396,29 @@ def test_invalid_proof_rejected(self, key_manager: XmssKeyManager) -> None:
...
+
+        proof = make_aggregated_proof(key_manager, [ValidatorIndex(1)], invalid_data)
+
+        signed_aggregated = SignedAggregatedAttestation(
+            data=invalid_data,
+            proof=proof,
+        )
+
+        with pytest.raises(AssertionError, match="Source checkpoint slot must not exceed target"):
+            store.on_gossip_aggregated_attestation(signed_aggregated)
```
tcoratger: Since you have a test vector for each case, I think that adding this act just as duplicate no? Especially since I guess that here we are not testing all the scenarios but probably just one and so to test all the scenarios, we would need multiple unit tests which will be a real duplicate of the behavior of the test vectors you added.

```suggestion
```

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -789,3 +808,137 @@ def _build_overridden_attestation_data(
...
+                attestation_data,
+            )
+        else:
+            signature = create_dummy_signature()
+
+        return SignedAttestation(
+            validator_id=spec.validator_id,
+            data=attestation_data,
+            signature=signature,
+        )
```
tcoratger: I've checked locally and that is dead code no?
```suggestion
```

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -789,3 +808,137 @@ def _build_overridden_attestation_data(
...
+
+        source = Checkpoint(root=source_root, slot=source_slot)
+
+        return AttestationData(
+            slot=spec.slot,
+            head=head,
+            target=target,
+            source=source,
+        )
```
tcoratger, line 906: I've checked locally and I feel like the following version is simpler to understand, more compact and I've documented it to explain the concepts one by one.

```suggestion
    def _build_attestation_data_from_gossip_aggregated_spec(
        self,
        spec: GossipAggregatedAttestationSpec,
        block_registry: dict[str, Block],
        state: State,
    ) -> AttestationData:
        """
        Build attestation data for a gossip aggregated attestation.

        Resolves the three checkpoints (target, head, source) from the spec.

        The state provides fallback values for the source checkpoint
        when neither an explicit root nor a label is specified.

        Args:
            spec: Aggregated attestation specification with checkpoint overrides.
            block_registry: Labeled blocks for checkpoint resolution.
            state: Head state providing the latest justified checkpoint as source fallback.

        Returns:
            Attestation data with all three checkpoints resolved.

        Raises:
            ValueError: If the target has neither an explicit root nor
                a label.
        """
        # Resolve the target checkpoint.
        #
        # An explicit root takes highest priority.
        # A label triggers lookup in the block registry.
        # Unlike the other two checkpoints, the target is mandatory.
        if spec.target_root is not None:
            target = Checkpoint(root=spec.target_root, slot=spec.target_slot)
        elif spec.target_root_label is not None:
            target = self._resolve_checkpoint(
                spec.target_root_label, spec.target_slot, block_registry
            )
        else:
            raise ValueError("gossip aggregated attestation spec requires a target root")

        # Resolve the head checkpoint.
        #
        # Priority: explicit root > label > target checkpoint.
        # - When using an explicit root without a slot, the target slot is used.
        # - When no head information is provided at all, the head mirrors the target.
        #
        # This matches honest validator behavior.
        if spec.head_root is not None:
            head = Checkpoint(
                root=spec.head_root,
                slot=spec.head_slot if spec.head_slot is not None else spec.target_slot,
            )
        elif spec.head_root_label is not None:
            head = self._resolve_checkpoint(spec.head_root_label, spec.head_slot, block_registry)
        else:
            head = Checkpoint(
                root=target.root,
                slot=spec.head_slot if spec.head_slot is not None else target.slot,
            )

        # Resolve the source checkpoint.
        #
        # Priority: explicit root > label > latest justified from state.
        # The source represents the most recent justified checkpoint the attester is aware of.
        # When not overridden, the state's latest justified checkpoint provides the correct default.
        if spec.source_root is not None:
            source = Checkpoint(
                root=spec.source_root,
                slot=(
                    spec.source_slot
                    if spec.source_slot is not None
                    else state.latest_justified.slot
                ),
            )
        elif spec.source_root_label is not None:
            source = self._resolve_checkpoint(
                spec.source_root_label, spec.source_slot, block_registry
            )
        else:
            source = Checkpoint(
                root=state.latest_justified.root,
                slot=(
                    spec.source_slot
                    if spec.source_slot is not None
                    else state.latest_justified.slot
                ),
            )

        # Assemble the final attestation data from resolved checkpoints.
        return AttestationData(
            slot=spec.slot,
            head=head,
            target=target,
            source=source,
        )
```

packages/testing/src/consensus_testing/test_fixtures/fork_choice.py
```diff
@@ -789,3 +808,137 @@ def _build_overridden_attestation_data(
...
+
+        if spec.signer_ids and spec.signer_ids != spec.validator_ids:
+            proof = AggregatedSignatureProof(
+                participants=AggregationBits.from_validator_indices(
+                    ValidatorIndices(data=validator_ids)
+                ),
+                proof_data=proof.proof_data,
+            )
+
+        return SignedAggregatedAttestation(data=attestation_data, proof=proof)
```
tcoratger, line 1000: Here I think that we could make things a bit more tractable like this

```suggestion
        def _build_signed_aggregated_attestation_from_spec(
        self,
        spec: GossipAggregatedAttestationSpec,
        store: Store,
        block_registry: dict[str, Block],
        key_manager: XmssKeyManager,
    ) -> SignedAggregatedAttestation:
        """
        Build a signed aggregated attestation from a gossip specification.

        Args:
            spec: Aggregated attestation specification with participant and checkpoint configuration.
            store: Fork choice store providing head state for checkpoint resolution.
            block_registry: Labeled blocks for checkpoint resolution.
            key_manager: XMSS key manager for signing attestation data.

        Returns:
            Signed aggregated attestation ready for gossip processing.
        """
        # Resolve checkpoints into concrete attestation data.
        attestation_data = self._build_attestation_data_from_gossip_aggregated_spec(
            spec,
            block_registry,
            store.states[store.head],
        )

        # Separate "claimed" from "actual" participants.
        #
        # - Claimed validators appear in the proof's participant bitfield.
        # - Actual signers produce the cryptographic material.
        # They default to the same set for honest attestations.
        validator_ids = spec.validator_ids
        signer_ids = spec.signer_ids or spec.validator_ids

        # Path 1: Invalid signature.
        #
        # Build a proof with the correct participant bitfield but zeroed-out proof bytes.
        # This exercises signature verification rejection.
        if not spec.valid_signature:
            proof = AggregatedSignatureProof(
                participants=AggregationBits.from_validator_indices(
                    ValidatorIndices(data=validator_ids)
                ),
                proof_data=ByteListMiB(data=b"\x00" * 32),
            )
            return SignedAggregatedAttestation(data=attestation_data, proof=proof)

        # Path 2: Valid signature.
        #
        # Collect each signer's public key and individual signature over the attestation data.
        # These feed into the aggregation step.
        raw_xmss = [
            (
                key_manager.get_public_keys(vid)[0],
                key_manager.sign_attestation_data(vid, attestation_data),
            )
            for vid in signer_ids
        ]

        # Encode which validators actually produced signatures.
        xmss_participants = AggregationBits.from_validator_indices(
            ValidatorIndices(data=signer_ids)
        )

        # Aggregate individual signatures into a single proof.
        #
        # The proof combines all signer contributions so the verifier
        # can check them in one pass against the attestation data root.
        proof = AggregatedSignatureProof.aggregate(
            xmss_participants=xmss_participants,
            children=[],
            raw_xmss=raw_xmss,
            message=attestation_data.data_root_bytes(),
            slot=attestation_data.slot,
        )

        # Path 3: Participant mismatch.
        #
        # Replace the participant bitfield with different validator indices
        # while keeping the original proof bytes intact.
        # The proof is cryptographically valid for the actual signers,
        # but the claimed participants no longer match.
        # The store must detect and reject this inconsistency.
        if spec.signer_ids and spec.signer_ids != spec.validator_ids:
            proof = AggregatedSignatureProof(
                participants=AggregationBits.from_validator_indices(
                    ValidatorIndices(data=validator_ids)
                ),
                proof_data=proof.proof_data,
            )

        return SignedAggregatedAttestation(data=attestation_data, proof=proof)
```

PR #496, opened 2026-03-27
Update multisig bindings to support recursive aggregation

anshalshukla:
## 🗒️ Description
Updates multisig bindings to make them devnet4 compatible. Other changes are minimal, related to dead code activation, store/state recursive flag removal and minor other test fixes. Few test were commented out
- Client cross compatibility test, can be reintroduced after clients complete the migration
- `test_valid_signature_wrong_validator` the verify API of aggregation no longer accepts public keys which isn't ideal as we need to verify that it does contain the signatures for a particular message at a particular slot for given set of public keys

The changes look elaborate and invasive when compared to main branch but are minimal when compared to tcoratger/poseidon11

## ✅ Checklist

    ```console
    uvx tox
    ```


src/lean_spec/subspecs/xmss/aggregation.py
```diff
@@ -20,6 +20,9 @@
 
 from .containers import PublicKey, Signature
 
+LOG_INV_RATE = 2
+"""Inverse rate exponent for proof generation (1-4, lower = faster but bigger proofs)."""
```
tcoratger: We should say where does this comes from and why we have this magic number here for the reader and future implementers I think

tests/consensus/devnet/fc/test_lexicographic_tiebreaker.py
```diff
@@ -61,53 +59,70 @@ def test_equal_weight_forks_use_lexicographic_tiebreaker(
...
+            #                     slot=Slot(4),
+            #                     target_slot=Slot(4),
+            #                     target_root_label="fork_b_4",
+            #                 ),
+            #             ],
+            #         ),
+            #         checks=StoreChecks(
+            #             lexicographic_head_among=["fork_a_2", "fork_b_4"],
+            #         ),
+            #     ),
```
tcoratger: We should put this test back no? Maybe a modified version of it if needed but we should put it back I think. Why does the recursive signature aggregation change something here? Should be equivalent behaviour in terms of tiebreaking no?

tests/consensus/devnet/verify_signatures/test_invalid_signatures.py
```diff
@@ -100,43 +100,44 @@ def test_invalid_aggregated_attestation_signature(
...
+#                     validator_ids=[ValidatorIndex(0), ValidatorIndex(1)],
+#                     signer_ids=[ValidatorIndex(2), ValidatorIndex(3)],
+#                     slot=Slot(1),
+#                     target_slot=Slot(0),
+#                     target_root_label="genesis",
+#                 ),
+#             ],
+#         ),
+#         expect_exception=AssertionError,
+#     )
```
tcoratger: I guess we can't put this test back right? If this is the case we should have an issue for this to track and explain the problem right?

tests/lean_spec/subspecs/xmss/test_cross_client_compat.py
tcoratger, line 1: I think that we can just remove this file for now since it is super complex to maintain and the compatibility is tested through the rest of the test suite. In the future when we have much more stability we can add something back here.

src/lean_spec/subspecs/forkchoice/store.py
```diff
@@ -919,86 +919,45 @@ def update_safe_target(self) -> "Store":
...
+        # new payloads. Remove this block once bindings support recursive aggregation.
+        aggregated_results = head_state.aggregate(
+            attestation_signatures=self.attestation_signatures,
+            new_payloads=None,
+            known_payloads=None,
+        )
+        new_aggregates = []
+        new_aggregated_payloads = {
+            attestation_data: set(proofs)
+            for attestation_data, proofs in self.latest_new_aggregated_payloads.items()
```
tcoratger: Before didn't we had this instead?

```suggestion
            for attestation_data, proofs in self.latest_known_aggregated_payloads.items()
```

tests/consensus/devnet/fc/test_lexicographic_tiebreaker.py
anshalshukla: A modified version is already in place where we are directly comparing the chains at the fork point instead of comparing them at the tip of the forked chains

```
          ---> A ---> B
Root
          ---> C ---> D
```

Earlier we were comparing B and D lexicographically but ideally the lexicographic comparison should happen at the forking point so A and C should be compared with both branches weight equally, the behavior has been updated I'll remove the comment. This will fix the currently failing production test in CI as well

tests/consensus/devnet/verify_signatures/test_invalid_signatures.py
anshalshukla: This is no more required, the public keys order is no more a requirement. The keys are sorted before aggregation and verification. I'll remove this test completely as it is no more relevant. See: https://github.com/leanEthereum/leanMultisig/pull/174

tests/lean_spec/subspecs/xmss/test_cross_client_compat.py
anshalshukla: Okay

src/lean_spec/subspecs/forkchoice/store.py
anshalshukla: I've updated the aggregation logic here, the plain signature aggregation logic has been replaced by recursive aggregation logic

tcoratger (approved):

src/lean_spec/subspecs/containers/state/state.py
```diff
@@ -786,34 +794,43 @@ def aggregate(
...
-                    continue
-                public_key = self.validators[entry.validator_id].get_attestation_pubkey()
-                raw_entries.append((entry.validator_id, public_key, entry.signature))
-                covered_validators.add(entry.validator_id)
+            # Phase 1: Greedily select child proofs for maximum validator coverage.
+            # New payloads are prioritized over known payloads.
+            child_proofs, covered = self._select_proofs_greedily(new.get(data), known.get(data))
+
+            # Phase 2: Collect raw XMSS signatures for validators not yet covered.
+            # Sorted by validator index as required by the Rust aggregation bindings.
```
anshalshukla: nit: aggregation bindings does not expect in a particular order and does the sorting internally by public key, validator index is unknown to aggregation

PR #499, opened 2026-03-27
Enable attestation support in StateTransitionTest fixture

uink45:
## 🗒️ Description

Wire `BlockSpec.attestations` through to `State.build_block()` in `StateTransitionTest`, which was previously defined in the data model but never connected to block construction.

This enables future test fixtures that exercise `process_attestations()`, justification, and finalization. Currently no exportable state transition vectors exist for these code paths.

Changes (single file: `state_transition.py`):
1. Add block registry with duplicate label rejection to `make_fixture()`
2. Update `_build_block_from_spec()` to build aggregated payloads when `spec.attestations` is provided
3. Add `_build_aggregated_payloads_from_spec()` for signing and aggregating attestations without a Store
4. Add `_resolve_checkpoint()` and `_build_attestation_data_from_spec()` aligned with `ForkChoiceTest`'s approach
5. Compute `max_slot` for XMSS key lifetime from attestation specs

All existing tests pass unchanged.

## 🔗 Related Issues or PRs

N/A

## ✅ Checklist

- [x] Ran `tox` checks to avoid unnecessary CI fails:
    ```console
    uvx tox
    ```
- [x] Considered adding appropriate tests for the changes.
- [ ] Considered updating the online docs in the [./docs/](/leanEthereum/leanSpec/tree/main/docs/) directory.

PR #500, opened 2026-03-28
fix: resolve parent_label from block registry in StateTransitionTest

tcoratger:
## Summary

- `BlockSpec.parent_label` was defined in the data model but never wired into `_build_block_from_spec()`, so it was silently ignored
- Linear chain tests worked accidentally (latest block header matches), but fork tests would get wrong parent roots
- Add `block_registry` to `make_fixture()` and resolve `parent_label` with three-way priority: explicit `parent_root` > `parent_label` > `latest_block_header`
- Reject duplicate labels early with a clear error message

## Test plan

- [x] `uv run fill --fork=devnet --clean -n auto -k state_transition` — all 14 vectors pass
- [x] `uvx tox -e all-checks` — all quality checks pass

### DIFF
--- a/packages/testing/src/consensus_testing/test_fixtures/state_transition.py
+++ b/packages/testing/src/consensus_testing/test_fixtures/state_transition.py
@@ -134,15 +134,25 @@ class StateTransitionTest(BaseConsensusFixture):
 
         # Initialize filled_blocks list that will be populated as we process blocks
         filled_blocks: list[Block] = []
+        block_registry: dict[str, Block] = {}
         try:
             state = self.pre
 
             for block_spec in self.blocks:
                 # Build block and optionally get cached post-state to avoid redundant transitions
-                block, cached_state = self._build_block_from_spec(block_spec, state)
+                block, cached_state = self._build_block_from_spec(block_spec, state, block_registry)
 
                 # Store the filled Block for serialization
                 filled_blocks.append(block)
+
+                # Register labeled blocks for parent resolution
+                if block_spec.label is not None:
+                    if block_spec.label in block_registry:
+                        raise ValueError(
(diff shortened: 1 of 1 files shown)


packages/testing/src/consensus_testing/test_fixtures/state_transition.py
```diff
@@ -220,9 +237,16 @@ def _build_block_from_spec(self, spec: BlockSpec, state: State) -> tuple[Block,
...
+        # Use provided parent root, resolve from label, or compute from state
         if spec.parent_root is not None:
             parent_root = spec.parent_root
+        elif spec.parent_label is not None:
+            if spec.parent_label not in block_registry:
+                raise ValueError(
+                    f"parent_label '{spec.parent_label}' not found - "
+                    f"available: {list(block_registry.keys())}"
+                )
+            parent_root = hash_tree_root(block_registry[spec.parent_label])
```
tcoratger, line 249: @uink45 I think that we should merge this before https://github.com/leanEthereum/leanSpec/pull/499 because I'm not sure that after this, the test you designed will pass, just a label problem with `parent_label="block_1"` which now should fail and needs to be fixed in https://github.com/leanEthereum/leanSpec/pull/499 

cc @unnawut

uink45: Yeah @tcoratger, happy for PR [#500](https://github.com/leanEthereum/leanSpec/pull/500) to be merged first. 

The PR [499](https://github.com/leanEthereum/leanSpec/pull/499) has overlapping block registry changes since I added it for attestation target label resolution. Once #500 is merged I will rebase #499 on top of it to remove the duplicate block registry code and keep just the attestation-specific additions.

PR #505, opened 2026-03-31
test: add consensus-critical justification and finalization test vectors

uink45:
## 🗒️ Description

Add 10 new state transition test vectors to `test_justification.py` exercising consensus-critical edge cases for justification and finalization:

- **Exact 2/3 threshold boundary:** Verifies that 4-of-6 validators, hitting the exact `>=` boundary, is sufficient to justify a target.
- **Below-threshold rejection:** Verifies that 3-of-6 validators, falling below the two-thirds threshold, does not advance justification.
- **Vote accumulation across blocks:** Verifies that pending votes from earlier blocks are preserved and eventually cross the threshold when new validators contribute in a later block.
- **Repeated validator dedup across blocks:** Verifies that the same validators voting twice for the same target across separate blocks do not inflate the tally.
- **Finalization on next justifiable step:** Verifies that justifying the immediately next justifiable slot after the source finalizes the source checkpoint.
- **No finalization with intermediate justifiable slot:** Verifies that justifying a distant target advances the justified checkpoint but does not finalize the source when intermediate justifiable slots exist in between.
- **Mid-block finalized-slot visibility:** Verifies that a second attestation within the same block can see the finalized-slot update caused by the first attestation during mid-block processing.
- **Pronic boundary acceptance:** Verifies that a target at pronic distance (delta 6 = 2×3) from the finalized slot is accepted for justification.
- **Non-justifiable boundary rejection:** Verifies that a target at non-justifiable distance (delta 7) from the finalized slot is rejected despite supermajority support.
- **Square boundary acceptance:** Verifies that a target at square distance (delta 9 = 3²) from the finalized slot is accepted for justification.

## 🔗 Related Issues or PRs

N/A

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/tests/consensus/devnet/state_transition/test_justification.py
+++ b/tests/consensus/devnet/state_transition/test_justification.py
@@ -233,22 +233,22 @@ def test_repeated_validators_do_not_double_count_across_blocks(
 
     Scenario
     --------
-    1. Start from genesis with 6 validators
-    2. Process block_1 at slot 1 and block_2 at slot 2
-    3. Process block_2 with attestations from validators 0 and 1 targeting
-       block_1 at slot 1
-    4. Process block_3 with the same validators 0 and 1 targeting block_1 again
-    5. Process block_4 with new validators 2, 3, 4, and 5 targeting block_2 at slot 2
+    1. Start from genesis with 4 validators
+    2. Process block_1 at slot 1
+    3. Process block_2 at slot 2 with attestations from validators 0 and 1
+       targeting block_1 at slot 1
+    4. Process block_3 at slot 3 with the same validators 0 and 1 targeting
+       block_1 again
 
     Expected Behavior
     -----------------
-    1. Duplicate votes for block_1 do not create new validator weight
-    2. block_1 must not become justified from repeated participants alone
-    3. The later block_2 attestation still uses source slot 0
-    4. latest_justified_slot advances to slot 2 without finalizing slot 1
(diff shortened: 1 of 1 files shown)


tests/consensus/devnet/state_transition/test_justification.py
```diff
@@ -61,3 +61,645 @@ def test_supermajority_attestations_justify_block(
...
+                            ValidatorIndex(4),
+                            ValidatorIndex(5),
+                        ],
+                        slot=Slot(4),
+                        target_slot=Slot(2),
+                        target_root_label="block_2",
+                    ),
+                ],
+            ),
+        ],
```
tcoratger, line 286: Here I don't think that we are testing the duplication thing properly because even if dedup was broken, `latest_justified_slot` would still be 2 because validators 2, 3, 4, 5 voted for this separately.

To truly isolate dedup, we need:
  - 4 validators, validators 0,1 vote in block A, then validators 0,1 vote again in block B (same target). Unique = 2/4. Double-counted = 4/4.
  - 3*2 >= 2*4 → 6 >= 8 → False (correct, no justify)
  - 3*4 >= 2*4 → 12 >= 8 → True (broken dedup would justify)
  - Assert `latest_justified_slot=0`

uink45: Thanks for the feedback. I have updated the PR to account for this.

PR #506, opened 2026-03-31
fix: use head state's justified checkpoint as attestation source

GrapeBaBa:
## Summary

`produce_attestation_data` was using `store.latest_justified` as the attestation source, while `build_block` filters attestations against `head_state.latest_justified`. When `store.latest_justified` diverges from `head_state.latest_justified` (caused by processing a non-head-chain block that crosses the 2/3 supermajority threshold), all attestations get filtered out during block building, producing blocks with 0 attestations.

## Root Cause

`store.latest_justified` can advance past `head_state.latest_justified` when a non-head-chain block's state transition crosses the 2/3 threshold. This creates a mismatch:

- **Voting**: `produce_attestation_data` used `store.latest_justified` → higher checkpoint
- **Block building**: `build_block` filters with `head_state.latest_justified` → lower checkpoint
- **Result**: All attestations fail the source filter → 0-attestation blocks

## Fix

Align with the 3sf-mini reference design: the attestation source comes from `head_state.latest_justified`, not from the store-wide max.

**Ref**: [ethereum/research 3sf-mini/p2p.py](https://github.com/ethereum/research/blob/master/3sf-mini/p2p.py) — `Staker.vote()` uses `self.post_states[self.head].latest_justified_hash`

The store-wide `latest_justified` (global max across all processed blocks) is intentionally used **only for fork choice** (`recompute_head`, `compute_safe_target`), never for voting — matching `get_latest_justified_hash(post_states)` in the reference.

### DIFF
--- a/src/lean_spec/subspecs/forkchoice/store.py
+++ b/src/lean_spec/subspecs/forkchoice/store.py
@@ -97,11 +97,14 @@ class Store(StrictBaseModel):
 
     latest_justified: Checkpoint
     """
-    Highest slot justified checkpoint known to the store.
+    Highest-slot justified checkpoint known to the store.
 
-    LMD GHOST starts from this checkpoint when computing the head.
-
+    This is the global maximum across all forks.
+    Used only for fork choice: LMD GHOST starts here when computing the head.
     Only descendants of this checkpoint are considered viable.
+
+    Attestation production does not use this field.
+    Validators vote with the justified checkpoint from the head state instead.
     """
 
     latest_finalized: Checkpoint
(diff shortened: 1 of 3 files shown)


tcoratger (approved):
Thanks a lot for noticing the bug here @GrapeBaBa!

I've just pushed an additional commit on top with a dedicated test vector that fails on main branch and that passes here so that client teams can also be sure that this behaviour is fine on their side.

unnawut (commented):
Looks great and really good spotting. Plus I really like the informative description, really helpful.

I wonder if this is the reason for frequent reorgs and chain not justifying with longer devnet runs...

PR #509, opened 2026-04-06
test: add consensus-critical justification test vectors

uink45:
## 🗒️ Description

Add 9 new consensus-critical justification tests to `test_justification.py`. All tests use the `StateTransitionTestFiller` pattern, generating JSON fixtures for client implementations.

**New tests:**

| Test | Rule Covered |
|---|---|
| `test_repeated_validator_does_not_double_count_within_same_block` | Same validator in two attestations within one block counts once |
| `test_pending_justification_survives_finalization_rebase` | Pending votes for valid targets survive when finalization rebases the window |
| `test_split_supermajority_aggregations_in_same_block_justify` | Multiple aggregates for the same target merge within a single block |
| `test_odd_validator_threshold_boundary_justifies` | 4/5 validators (80%) meets the 2/3 integer threshold |
| `test_odd_validator_threshold_boundary_does_not_justify` | 3/5 validators (60%) does not meet the 2/3 integer threshold |
| `test_supermajority_with_mismatched_target_root_is_ignored` | Supermajority with wrong canonical root is rejected |
| `test_justification_clears_only_the_resolved_target_votes` | Justifying one target clears only that target's pending votes |
| `test_finalization_prunes_stale_pending_votes_and_rebases_window` | Finalization prunes stale pending votes and shifts the tracking window |
| `test_target_at_or_before_source_is_ignored` | Votes targeting a slot at or before the source are rejected |

**Additional changes:**
- Enhanced `test_votes_accumulate_across_blocks` with pending vote cleanup assertions (`justifications_roots`, `justifications_validators`)
- Renamed `test_exact_two_thirds_threshold_justifies` to `test_even_validator_threshold_boundary` to pair with the new odd-validator counterpart

## 🔗 Related Issues or PRs

N/A

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/tests/consensus/devnet/state_transition/test_justification.py
+++ b/tests/consensus/devnet/state_transition/test_justification.py
@@ -316,10 +316,9 @@ def test_repeated_validator_does_not_double_count_within_same_block(
 
     Expected Behavior
     -----------------
-    1. The two attestations do not merge because their attestation-data slots differ
-    2. Justification tallies are keyed by target root, not by raw attestation count
-    3. Validator 0 contributes only one unit of support toward block_1
-    4. Unique support remains three of six validators, below the threshold
+    1. Justification tallies are keyed by target root, not by raw attestation count
+    2. Validator 0 contributes only one unit of support toward block_1
+    3. Unique support remains three of six validators, below the threshold
     """
     state_transition_test(
         pre=generate_pre_state(num_validators=6),
@@ -520,13 +519,7 @@ def test_pending_justification_survives_finalization_rebase(
             latest_justified_slot=Slot(2),
             latest_finalized_slot=Slot(1),
             justified_slots=JustifiedSlots(data=[]).model_copy(
-                update={
-                    "data": [
-                        Boolean(True),
-                        Boolean(False),
-                        Boolean(False),
-                    ]
-                }
+                update={"data": [Boolean(True), Boolean(False), Boolean(False)]}
             ),
             justifications_validators=JustificationValidators(
                 data=[
@@ -922,6 +915,9 @@ def test_split_supermajority_aggregations_in_same_block_justify(
             slot=Slot(2),
             latest_justified_slot=Slot(1),
             latest_finalized_slot=Slot(0),
+            justified_slots=JustifiedSlots(data=[]).model_copy(update={"data": [Boolean(True)]}),
+            justifications_roots=JustificationRoots(data=[]),
+            justifications_validators=JustificationValidators(data=[]),
         ),
     )
 
@@ -971,6 +967,9 @@ def test_odd_validator_threshold_boundary_justifies(
             slot=Slot(2),
             latest_justified_slot=Slot(1),
             latest_finalized_slot=Slot(0),
+            justified_slots=JustifiedSlots(data=[]).model_copy(update={"data": [Boolean(True)]}),
+            justifications_roots=JustificationRoots(data=[]),
+            justifications_validators=JustificationValidators(data=[]),
         ),
     )
 
@@ -1019,6 +1018,16 @@ def test_odd_validator_threshold_boundary_does_not_justify(
             slot=Slot(2),
             latest_justified_slot=Slot(0),
             latest_finalized_slot=Slot(0),
+            justified_slots=JustifiedSlots(data=[]).model_copy(update={"data": [Boolean(False)]}),
+            justifications_validators=JustificationValidators(
+                data=[
+                    Boolean(True),
+                    Boolean(True),
+                    Boolean(True),
+                    Boolean(False),
+                    Boolean(False),
+                ]
+            ),
         ),
     )
 
@@ -1068,6 +1077,8 @@ def test_supermajority_with_mismatched_target_root_is_ignored(
             slot=Slot(3),
             latest_justified_slot=Slot(0),
             latest_finalized_slot=Slot(0),
+            justifications_roots=JustificationRoots(data=[]),
+            justifications_validators=JustificationValidators(data=[]),
         ),
     )
 


tests/consensus/devnet/state_transition/test_justification.py
```diff
@@ -291,6 +300,75 @@ def test_repeated_validators_do_not_double_count_across_blocks(
...
+                        target_root_label="block_1",
+                    ),
+                ],
+            ),
+        ],
+        post=StateExpectation(
+            slot=Slot(2),
+            latest_justified_slot=Slot(0),
+            latest_finalized_slot=Slot(0),
+            justified_slots=JustifiedSlots(data=[]).model_copy(update={"data": [Boolean(False)]}),
```
tcoratger, line 356: We could have this no?

```suggestion
            justified_slots=JustifiedSlots(data=[Boolean(False)]),
```

tests/consensus/devnet/state_transition/test_justification.py
```diff
@@ -359,6 +437,109 @@ def test_finalization_on_next_justifiable_step(
...
+            latest_finalized_slot=Slot(1),
+            justified_slots=JustifiedSlots(data=[]).model_copy(
+                update={
+                    "data": [
+                        Boolean(True),
+                        Boolean(False),
+                        Boolean(False),
+                    ]
+                }
+            ),
```
tcoratger, line 523: Same thing here we can simplify this and I guess this is the case for other places as well

tests/consensus/devnet/state_transition/test_justification.py
```diff
@@ -291,6 +300,75 @@ def test_repeated_validators_do_not_double_count_across_blocks(
...
+    --------
+    1. Start from genesis with 6 validators
+    2. Process block_1 at slot 1
+    3. Process block_2 at slot 2 with two attestations targeting block_1:
+       - validators 0, 1, and 2 attest with attestation slot 1
+       - validator 0 attests again with attestation slot 2
+
+    Expected Behavior
+    -----------------
+    1. The two attestations do not merge because their attestation-data slots differ
```
tcoratger: We should remove this point because we never test it no?

tests/consensus/devnet/state_transition/test_justification.py
```diff
@@ -685,3 +866,531 @@ def test_square_boundary_acceptance(
...
+                        target_root_label="block_1",
+                    ),
+                ],
+            ),
+        ],
+        post=StateExpectation(
+            slot=Slot(2),
+            latest_justified_slot=Slot(1),
+            latest_finalized_slot=Slot(0),
+        ),
```
tcoratger, line 921: Why don't we test `justifications_validators` and `justified_slots` as well?

tests/consensus/devnet/state_transition/test_justification.py
```diff
@@ -685,3 +866,531 @@ def test_square_boundary_acceptance(
...
```
tcoratger, line 973: Same here why don't we test `justifications_validators` and `justified_slots` as well?

tests/consensus/devnet/state_transition/test_justification.py
```diff
@@ -685,3 +866,531 @@ def test_square_boundary_acceptance(
...
+                        target_root_label="block_1",
+                    ),
+                ],
+            ),
+        ],
+        post=StateExpectation(
+            slot=Slot(2),
+            latest_justified_slot=Slot(0),
+            latest_finalized_slot=Slot(0),
+        ),
```
tcoratger, line 1031: Same here? Why don't we test `justifications_validators` and `justified_slots` as well?

tests/consensus/devnet/state_transition/test_justification.py
```diff
@@ -685,3 +866,531 @@ def test_square_boundary_acceptance(
...
+                        target_root_label="block_2",
+                    ),
+                ],
+            ),
+        ],
+        post=StateExpectation(
+            slot=Slot(3),
+            latest_justified_slot=Slot(0),
+            latest_finalized_slot=Slot(0),
+        ),
```
tcoratger, line 1082: Same here we should probably check the other stuffs right?

tests/consensus/devnet/state_transition/test_justification.py
```diff
@@ -685,3 +866,531 @@ def test_square_boundary_acceptance(
...
+                    ),
+                ],
+            ),
+        ],
+        post=StateExpectation(
+            slot=Slot(4),
+            latest_justified_slot=Slot(1),
+            latest_finalized_slot=Slot(0),
+            justified_slots=JustifiedSlots(data=[]).model_copy(
+                update={"data": [Boolean(True), Boolean(False), Boolean(False)]}
```
tcoratger, line 1164: Same here, no need to model copy

tests/consensus/devnet/state_transition/test_justification.py
```diff
@@ -685,3 +866,531 @@ def test_square_boundary_acceptance(
...
+                        target_root_label="block_5",
+                    ),
+                ],
+            ),
+        ],
+        post=StateExpectation(
+            slot=Slot(6),
+            latest_justified_slot=Slot(5),
+            latest_finalized_slot=Slot(4),
+            justified_slots=JustifiedSlots(data=[]).model_copy(update={"data": [Boolean(True)]}),
```
tcoratger, line 1281: Same here

tests/consensus/devnet/state_transition/test_justification.py
```diff
@@ -685,3 +866,531 @@ def test_square_boundary_acceptance(
...
+            justified_slots=JustifiedSlots(data=[]).model_copy(
+                update={
+                    "data": [
+                        Boolean(True),
+                        Boolean(False),
+                        Boolean(False),
+                        Boolean(True),
+                        Boolean(False),
+                    ]
+                }
```
tcoratger, line 1396: Same here

tests/consensus/devnet/state_transition/test_justification.py
uink45: Thanks for flagging this @tcoratger. 

Regarding the `model_copy` simplification, I realised that when I replace it with `JustifiedSlots(data=[Boolean(False)])` the constructor ends up converting the list into a tuple, while the actual state keeps it as a list. The equality check then fails when I run the state transition tests using `uv run pytest`:

```
AssertionError: State validation failed: justified_slots = data=[Boolean(False)], expected data=(Boolean(False),)
```

To fix this, I can submit a separate PR, however, I was wondering which approach should be taken as the fix could either go in `BaseBitlist._coerce_and_validate` (changing `return tuple(...)` to `return list(...)` to match what `model_copy` produces), or in `validate_against_state`. Do you have a preference on which approach?

tests/consensus/devnet/state_transition/test_justification.py
uink45: Thanks @tcoratger, I have added post state checks for `justifications_validators` and `justified_slots fields`. This has been done for the tests that were missing them, as flagged in the review.

PR #510, opened 2026-04-06
single attestation data

anshalshukla:
## 🗒️ Description
Merge multiple payloads for a particular `attestation_data` while block building.

## ✅ Checklist

    ```console
    uvx tox
    ```


anshalshukla:
I've removed a few testcase which were testing the greedy selection while block building, those can be added back but will need to separate out the core block_build as a separate function.

tcoratger (approved):

PR #527, opened 2026-04-07
test(node): increase orchestrator coverage to 98%

Parth-Vasave:
## 🗒️ Description
This PR increases the test coverage for the consensus orchestrator (`src/lean_spec/subspecs/node/node.py`). The existing test suite was extended to thoroughly cover multiple core components and failure states, elevating the line coverage from 81% up to **98%** (exceeding the target of 90%).

Specifically, test cases were added for:
- Database load flows (`_try_load_from_database`), covering all failure paths (e.g. no database, empty database, missing block/state/checkpoints) and successful load.
- Validator service wiring and publish wrappers.
- The `run()` lifecycle orchestrator, accounting for optional API server and validator service initialization.
- Signal handler installation (`_install_signal_handlers`) across non-main thread contexts.
- Periodic justified/finalized metrics logging (`_log_justified_finalized_periodically`).
- Node shutdown mechanics (`stop()`, `is_running`, and `_wait_shutdown()`).

## 🔗 Related Issues or PRs
Closes https://github.com/leanEthereum/leanSpec/issues/525

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/tests/lean_spec/subspecs/node/test_node.py
+++ b/tests/lean_spec/subspecs/node/test_node.py
@@ -3,12 +3,14 @@
 from __future__ import annotations
 
 import asyncio
+import dataclasses
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 
 from lean_spec.subspecs.api import ApiServerConfig
 from lean_spec.subspecs.chain.config import (
+    ATTESTATION_COMMITTEE_COUNT,
     HISTORICAL_ROOTS_LIMIT,
     INTERVALS_PER_SLOT,
     SECONDS_PER_SLOT,
@@ -96,6 +98,26 @@ def node_config() -> NodeConfig:
         network=MockNetworkRequester(),
         time_fn=lambda: 1704067200.0,
     )
+
+
+def _make_validator_registry() -> ValidatorRegistry:
+    """Create a minimal validator registry with one entry."""
+    registry = ValidatorRegistry()
+    registry.add(
+        ValidatorEntry(
+            index=ValidatorIndex(0),
+            attestation_secret_key=MagicMock(),
+            proposal_secret_key=MagicMock(),
+        )
+    )
+    return registry
+
+
+@pytest.fixture
+def node_with_validator(node_config: NodeConfig) -> Node:
+    """Create a node with a validator registry configured."""
+    config = dataclasses.replace(node_config, validator_registry=_make_validator_registry())
+    return Node.from_genesis(config)
 
 
 class TestNodeFromGenesis:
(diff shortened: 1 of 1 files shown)


tcoratger (commented):
Hey! I've checked with Claude code to review, overall looks good, but couple of comments that I think are pretty important to tackle to make things clean

## Critical

### 1. `test_logging_reports_metrics` and `test_logging_reports_zero_validators_without_service` have no assertions

These tests run the loop and implicitly check it doesn't crash, but the test names claim to verify metrics and validator count respectively. Without actual assertions on values, they provide false confidence in coverage. Either add real assertions (e.g. check a metric value from the registry) or remove them.

### 2. Global `asyncio.sleep` patching is fragile

`patch("asyncio.sleep", ...)` replaces sleep for ALL asyncio internals, not just the target loop. This is a known footgun in async testing. Consider patching at the module level instead (`"lean_spec.subspecs.node.node.asyncio.sleep"`) or restructuring to avoid patching sleep entirely — for example, calling `_log_justified_finalized_periodically` directly with a pre-set shutdown event.

## Major

### 3. Publish wrapper tests don't check arguments

`publish_attestation.assert_awaited_once()` doesn't verify the actual arguments passed. Per project conventions, full equality assertions are preferred over partial checks. The subnet_id computation (the interesting piece of logic in the wrapper) is never verified. Consider:

```python
publish_attestation.assert_awaited_once_with(mock_attestation, 0)
```

### 4. `test_falls_back_to_database_genesis_time` only asserts the mock was called

It should also assert `store.time` equals the expected value, like the sibling test `test_zero_genesis_time_when_database_returns_none` does. Checking the mock was called is necessary but not sufficient.

### 5. `test_run_with_api_server` doesn't verify `run` was awaited

Only `mock_start.assert_awaited_once()` is checked. If the `tg.create_task(self.api_server.run())` line were deleted, the test would still pass. Also assert `mock_run` was awaited to verify the server is actually added to the TaskGroup.

## Minor

### 6. `NodeConfig` boilerplate duplication

The same config block appears ~8 times across the new test classes. There's already a `node_config` fixture that's functionally identical — the new classes should use it instead of duplicating the construction inline.

### 7. `_make_node_with_validator` duplicates existing setup

Same registry + config + `from_genesis` pattern already exists in `TestOptionalServiceWiring`. A shared `node_with_validator` fixture would eliminate this duplication.

### 8. `TestWaitShutdown` tests mirror the implementation

Testing that `_wait_shutdown` calls `.stop()` on optional services is essentially testing `if x is not None: x.stop()`. These are low-value tests — consider whether they're worth keeping or if the effort is better spent on behavioral tests.

## Documentation

For the trickier tests (especially the periodic logging tests with async sleep patching, and the publish wrapper tests), it would really help to add documentation — both a clear docstring explaining what the test verifies and **why**, and line-by-line inline comments explaining the setup logic. When someone reads these tests 6 months from now, they should be able to understand the intent immediately without reverse-engineering the mocks. The project has strong documentation standards for this (every logical step gets a comment block).

## What's great

- `test_from_genesis_resumes_from_database` using a real in-memory SQLite DB is excellent — tests the actual round-trip, not mocks
- The genesis time fallback tests cover genuinely important recovery paths
- `test_wait_shutdown_skips_none_services` is clean and useful

Thanks again for the contribution!

tcoratger (approved):

tcoratger (approved):

PR #587, opened 2026-04-11
test(fc): block builder fixed-point justification loop

### LINKED ISSUE #564: test(fc): block builder fixed-point loop advances justification mid-build
## Context

The block builder in \`State.build_block\` uses a **fixed-point loop** to include attestations. This is subtle and important:

1. The builder collects valid attestation data
2. It processes them, which may **justify a new slot**
3. A newly justified slot may unlock **additional attestation data** that was previously invalid (because its source was unjustified)
4. The loop repeats until no new attestation data is found

This means a single block can contain attestation A (which justifies slot X) and attestation B (whose source is slot X) — even though B was only valid *after* A was processed.

**No spec test filler currently tests this fixed-point behavior.**

## What to test

Write a state transition filler that:

1. Creates a state where one attestation (A) will justify slot X
2. Another attestation (B) has source = slot X (currently unjustified, so B is invalid)
3. Builds a block containing both A and B
4. Verifies that the fixed-point loop includes both:
   - First iteration: finds A (valid), processes it, justifies slot X
   - Second iteration: finds B (now valid because source X is justified), includes it
5. Post-state should show both justification from A and any effect from B

### Key assertions

- Post-state has the slot justified by attestation A
- Post-state also reflects attestation B's contribution (e.g., a second justified slot, or additional votes)
- The block body contains attestations for both A and B

## Where to add the test

Create a new file: \`tests/consensus/devnet/fc/test_block_production.py\` or add to existing if created by issue #563.

## Code skeleton

```python
def test_block_builder_fixed_point_advances_justification(
    fork_choice_test: ForkChoiceTestFiller,
) -> None:
    """Fixed-point loop: justification from attestation A unlocks attestation B."""
    # This requires careful setup:
    # 1. Build chain to slot N
    # 2. Gossip attestation A (source=justified, target=slot X) from supermajority
    #    -> when processed, justifies slot X
    # 3. Gossip attestation B (source=slot X, target=slot Y)
    #    -> invalid until slot X is justified
    # 4. Build block at slot N+1
    #    -> fixed-point: processes A, justifies X, then finds B is now valid
    # 5. Verify block contains both A and B
    pass  # TODO: implement with actual slot/attestation setup
```

> **Note**: This is a more complex test. You need to understand the justification rules (\`is_justifiable_after\`) to pick valid source/target pairs. See the existing justification fillers in \`tests/consensus/devnet/state_transition/test_justification.py\` for reference.

## How to run

```bash
uv run fill --fork=devnet --clean -n auto -k test_block_builder_fixed_point
```

## References

- \`State.build_block\`: \`src/lean_spec/subspecs/containers/state/state.py\` — the fixed-point loop
- \`State.process_attestations\`: same file — how justification happens
- \`Slot.is_justifiable_after\`: \`src/lean_spec/subspecs/containers/slot.py\`
- Existing mid-processing test: \`tests/consensus/devnet/fc/test_finalization_mid_processing.py\`

latifkasuli:
## Summary

- Adds a fork-choice filler that exercises the multi-pass fixed-point loop in `State.build_block`, where processing attestation A advances `current_justified` and unlocks attestation B whose source was previously ineligible.
- Uses `GossipAggregatedAttestationStep` with explicit `source_root_label` overrides and tick-based payload migration to correctly model attestations with different source checkpoints reaching `latest_known_aggregated_payloads` before block production.
- Includes a mid-test `attestation_checks` checkpoint on the `TickStep` to localize future failures between payload migration and fixed-point block building.

## Test plan

- [x] `uv run fill --fork=devnet --clean -n auto -k test_block_builder_fixed_point` passes
- [x] Full suite `uv run fill --fork=devnet --clean -n auto` passes (124/124)
- [x] `ruff format` and `ruff check` clean

Closes #564

### DIFF
--- a/tests/consensus/devnet/fc/test_block_production.py
+++ b/tests/consensus/devnet/fc/test_block_production.py
@@ -1,16 +1,4 @@
-"""Fork Choice: Block Production — Fixed-Point Justification Loop
-
-The block builder in ``State.build_block`` uses a fixed-point loop to select
-attestation data from the store's aggregated payloads. Only entries whose
-``source`` matches the *current* justified checkpoint are eligible. When
-processing selected attestations justifies a new slot, the loop repeats and
-entries that were previously ineligible may now match.
-
-No existing spec test filler exercises this multi-pass behavior through the
-fork-choice block-production path.
-
-Reference: https://github.com/leanEthereum/leanSpec/issues/564
-"""
+"""Fork Choice: Block Production — Fixed-Point Justification Loop"""
 
 import pytest
 from consensus_testing import (
--- a/tests/consensus/devnet/fc/test_block_production.py
+++ b/tests/consensus/devnet/fc/test_block_production.py
@@ -1,4 +1,4 @@
-"""Fork Choice: Block Production — Fixed-Point Justification Loop"""
+"""Fork Choice: Block Production"""
 
 import pytest
 from consensus_testing import (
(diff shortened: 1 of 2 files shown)


tests/consensus/devnet/fc/test_block_production.py
```diff
@@ -0,0 +1,235 @@
...
+attestation data from the store's aggregated payloads. Only entries whose
+``source`` matches the *current* justified checkpoint are eligible. When
+processing selected attestations justifies a new slot, the loop repeats and
+entries that were previously ineligible may now match.
+
+No existing spec test filler exercises this multi-pass behavior through the
+fork-choice block-production path.
+
+Reference: https://github.com/leanEthereum/leanSpec/issues/564
+"""
```
tcoratger:
```suggestion
"""Fork Choice: Block Production — Fixed-Point Justification Loop"""
```

tcoratger (approved):

PR #594, opened 2026-04-12
test: cover skipped attestation with mismatched historical root

### LINKED ISSUE #578: test(st): attestation with root not matching historical block hashes is skipped
## Context

During attestation processing, the attestation's target root (and source root) must match the corresponding entry in the state's \`historical_block_hashes\`. If the root doesn't match the block hash at the referenced slot, the attestation is silently skipped.

This check prevents attestations from referencing non-existent or incorrect blocks. The state maintains a list of block hashes for each slot, and attestations must point to blocks that actually exist in the canonical chain (as seen by this state).

The check in \`process_attestations\`:
\`\`\`python
if att_data.target.root != self.historical_block_hashes[target_index]:
    continue
\`\`\`

## What to test

Write a state transition filler that:

1. Builds a chain with blocks at specific slots
2. Includes an attestation where the target root is a valid \`Bytes32\` but does NOT match the block hash at the target slot in \`historical_block_hashes\`
3. Verifies the attestation is silently skipped (no error, no justification effect)

### Key assertions

- Post-state shows no justification from the mismatched-root attestation
- No errors during processing
- The block itself is processed successfully

## Where to add the test

Add to: \`tests/consensus/devnet/state_transition/test_justification.py\`

## Code skeleton

```python
def test_attestation_with_target_root_not_in_historical_hashes_is_skipped(
    state_transition_test: StateTransitionTestFiller,
) -> None:
    """Attestation with target root that doesn't match historical block hashes is ignored."""
    # Build a chain, then include an attestation with a fabricated target root
    # that doesn't match any block in historical_block_hashes at that slot.
    #
    # The AggregatedAttestationSpec may need a way to override the target root
    # to a specific (incorrect) value. Check available spec fields.
    pass
```

## How to run

```bash
uv run fill --fork=devnet --clean -n auto -k test_attestation_with_target_root_not_in
```

## References

- \`State.process_attestations\`: \`src/lean_spec/subspecs/containers/state/state.py\`
- \`HistoricalBlockHashes\`: \`src/lean_spec/subspecs/containers/state/types.py\`

morelucks:
## Summary

  Add coverage for state-transition attestations whose target root does not match the canonical root recorded in `historical_block_hashes`.

  This extends the aggregated state-transition test helper so tests can inject explicit checkpoint roots without changing the default label-based path, then adds a devnet STF test that
  verifies a fabricated `Bytes32` target root is silently skipped during attestation processing.

  Closes #578

  ## Changes

  - extend `AggregatedAttestationSpec` with optional explicit `target`, `head`, and `source` checkpoint overrides
  - preserve existing helper behavior when overrides are not used
  - add `test_attestation_with_target_root_not_in_historical_hashes_is_skipped`
  - verify the block still processes successfully and no justification state is updated

  ## Verification

  - `uv run fill --fork=devnet --clean -n auto -k test_attestation_with_target_root_not_in`
  - `uv run fill --fork=devnet --clean -n auto -k test_supermajority_with_mismatched_target_root_is_ignored`

### DIFF
--- a/packages/testing/src/consensus_testing/test_types/aggregated_attestation_spec.py
+++ b/packages/testing/src/consensus_testing/test_types/aggregated_attestation_spec.py
@@ -5,11 +5,12 @@ from __future__ import annotations
 from lean_spec.subspecs.containers.attestation import AggregatedAttestation, AttestationData
 from lean_spec.subspecs.containers.block.block import Block
 from lean_spec.subspecs.containers.block.types import AggregatedAttestations
+from lean_spec.subspecs.containers.checkpoint import Checkpoint
 from lean_spec.subspecs.containers.slot import Slot
 from lean_spec.subspecs.containers.state.state import State
 from lean_spec.subspecs.containers.validator import ValidatorIndex, ValidatorIndices
 from lean_spec.subspecs.xmss.aggregation import AggregatedSignatureProof
-from lean_spec.types import ByteListMiB, CamelModel
+from lean_spec.types import ByteListMiB, Bytes32, CamelModel
 
 from ..keys import XmssKeyManager
 from .utils import resolve_checkpoint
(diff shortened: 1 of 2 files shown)


morelucks:
@tcoratger kindly review

tcoratger (approved):

PR #595, opened 2026-04-12
Revert "fix: use head state's justified checkpoint as attestation source"

### LINKED ISSUE #596: in update head, we should also add an `assert` that its downstream from the store's latest justified to make sure the head of chain is always "available" and to bring up any voting that is not consistent
in update head, we should also add an `assert` that its downstream from the store's latest justified to make sure the head of chain is always "available" and to bring up any voting that is not consistent

also where we have checkpoint sync add the spec change that it needs to download one block after the finalized state from the peers which "confirms" the checkpoint state's block as finalized and also helps node "discover" a justified block downstream

_Originally posted by @g11tech in https://github.com/leanEthereum/leanSpec/issues/595#issuecomment-4231159545_

GrapeBaBa:
Reverts leanEthereum/leanSpec#506

Closes https://github.com/leanEthereum/leanSpec/issues/596


g11tech:
in update head, we should also add an `assert` that its downstream from the store's latest justified to make sure the head of chain is always "available" and to bring up any voting that is not consistent

also where we have checkpoint sync add the spec change that it needs to download one block after the finalized state from the peers which "confirms" the checkpoint state's block as finalized and also helps node "discover" a justified block downstream

GrapeBaBa:
> in update head, we should also add an `assert` that its downstream from the store's latest justified to make sure the head of chain is always "available" and to bring up any voting that is not consistent
> 
> also where we have checkpoint sync add the spec change that it needs to download one block after the finalized state from the peers which "confirms" the checkpoint state's block as finalized and also helps node "discover" a justified block downstream

prefer this PR only do revert and create an issue for this.

g11tech (commented):
looks good, although 
```python
self._compute_lmd_ghost_head(
            start_root=self.latest_justified.root,
```
"should" give the head that is downstream from state's latest_justified already so not sure if the assert patch you added is required

however where we NEED to add assert patch is in block production which should asset that the produced block's latest justified is downstream from the current store's latest justified i.e. it includes for e.g. attestations that came in slot4 which you correctly noted should happen as "fix point" of the attestation packing process we have in build block.

and a  corresponding test for build block as well

PR #599, opened 2026-04-13
test: add consensus-critical finalization test vectors

uink45:
## 🗒️ Description

Add 5 new tests that are critical to consensus finalization in a dedicated `test_finalization.py`, consolidate finalization coverage by moving 5 existing vectors out of `test_justification.py`, and add 1 new block header test to `test_genesis.py`. Also extend `StateExpectation` with label based assertions for checkpoint and pending justification roots so every finalization vector pins its checkpoint roots to a specific block rather than only asserting the checkpoint slots. All tests use the `StateTransitionTestFiller` pattern, generating JSON fixtures for client implementations.

**New tests in `test_finalization.py`:**

| Test | Rule Covered |
|---|---|
| `test_non_adjacent_justification_finalizes_across_non_justifiable_gap` | A justified target that is not adjacent to its source still finalizes the source when every intermediate slot is not justifiable |
| `test_no_finalization_when_rebased_boundary_exposes_intermediate_justifiable_slot` | After the finalized boundary has rebased, an intermediate justifiable slot blocks further finalization |
| `test_mid_block_finalized_slot_rejects_target_that_loses_justifiability` | A later attestation within the same block is skipped when the updated `finalized_slot` makes its target no longer justifiable |
| `test_mid_block_resolved_target_does_not_reopen_pending_votes` | A later attestation within the same block whose target was already justified earlier in that block does not reopen a pending tally |
| `test_rebased_finalization_prunes_stale_votes_and_preserves_future_votes` | A second rebase prunes pending votes for targets behind the new finalized slot while preserving votes for targets beyond it |

**New test in `test_genesis.py`:**

| Test | Rule Covered |
|---|---|
| `test_first_post_genesis_block_sets_checkpoint_anchor_roots` | The first block after genesis anchors `latest_justified.root` and `latest_finalized.root` to its parent header root, leaving both slots at zero |

**Moved from `test_justification.py` into `test_finalization.py`:** The following 5 vectors are moved unchanged since their primary invariant is a finalization state change rather than justification eligibility:

- `test_finalization_on_next_justifiable_step`
- `test_pending_justification_survives_finalization_rebase`
- `test_no_finalization_when_intermediate_justifiable_slot_exists`
- `test_mid_block_finalized_slot_visibility`
- `test_finalization_prunes_stale_pending_votes_and_rebases_window`

**Framework addition:**

Three new fields on `StateExpectation` mirror the label based assertion style already used by `StoreChecks`:

- `latest_justified_root_label: str | None`
- `latest_finalized_root_label: str | None`
- `justifications_roots_labels: list[str] | None`

The state transition filler threads its local block registry into post state validation so each label resolves to `hash_tree_root(block_registry[label])`. For the list field, both sides of the comparison are sorted so the assertion stays correct regardless of the storage order used by `State.process_attestations`.

These fields let every finalization vector pin its finalized and justified checkpoint roots to a specific block. Previously the vectors asserted only the checkpoint slots, so a client that finalized the right slot with the wrong block root would still have passed. The two pending tally vectors also pin the surviving tally to its intended block root.

## 🔗 Related Issues or PRs

N/A

## ✅ Checklist

    ```console
    uvx tox
    ```


uink45:
Hi @tcoratger and everyone, confirming this PR is ready for review.

Let me know if there is any feedback for the changes made.

Thanks!

tcoratger (approved):
Thanks a lot, I've some customizations to be done but I'll handle this quickly on a followup PR not to block this one which is important to merge.

PR #624, opened 2026-04-16
framework: auto generate and publish test/prod scheme keys

unnawut:
## 🗒️ Description

This PR introduces auto generation of test and prod keys so when a change is made to the signature scheme, a new set of keys for both test and prod schemes are auto generated and published to https://github.com/leanEthereum/leansig-test-keys.

The leanSpec and leansig-test-keys are still two repos because otherwise we need to create releases on the leanSpec repo. Technically it's simpler to have the keys as releases in leanSpec repo but then we're not doing releases in leanSpec and having one just for the test keys could be confusing. 

## ✅ Checklist

    ```console
    uvx tox
    ```

### DIFF
--- a/.github/workflows/generate-keys.yml
+++ b/.github/workflows/generate-keys.yml
@@ -1,0 +1,69 @@
+name: Generate Keys
+
+on:
+  workflow_dispatch:
+  workflow_call:
+    secrets:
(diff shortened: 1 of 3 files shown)


unnawut:
This only impacts CI for keygen and prod vectors and needs testing on main branch, so I'll go ahead and merge without approval

PR #638, opened 2026-04-18
feat: add multi-fork architecture with ForkProtocol and SpecRunner

tcoratger:
We just start from devnet4 here to begin right now and not go back in time, which would be a mess.

Proposal from @leolara.

## Summary

- Introduce `ForkProtocol(ABC)` + `SpecRunner` for multi-fork dispatch
- Move `State` and `Store` from `subspecs/` into `forks/devnet4/` (fork-specific code)
- `subspecs/` now contains only fork-agnostic shared libraries
- Consumer modules import from `lean_spec.forks` (re-exports, fork-generic)
- Construction goes through `ForkProtocol` (`fork.generate_genesis()`, `fork.create_store()`)
- Rename testing fork `Devnet` → `Devnet4`, add `Devnet5` skeleton
- Adding a new fork: create `Devnet5State(State)` + `Devnet5Spec(Devnet4Spec)`, done

Based on [Multi-Fork Architecture proposal](https://notes.ethereum.org/@leolara/Multi-Fork-Architecture-for-leanSpec).

## Test plan

- [ ] `uvx tox -e all-checks` passes (ruff, format, ty, codespell, mdformat)
- [ ] `uv run pytest --no-cov` passes (all 3267 tests)
- [ ] `uv run fill --fork=devnet4 --clean -n auto` generates fixtures


src/lean_spec/subspecs/genesis/config.py
```diff
@@ -20,7 +20,8 @@
 import yaml
 from pydantic import Field, field_validator, model_validator
 
-from lean_spec.subspecs.containers import State, Validator
+from lean_spec.forks import ForkProtocol, State
```
tcoratger: I'm not happy with the imports of State and Store like this one (couple of them inside the subspecs/ folder) because it's the defaults that are imported here (and therefore for now devnet4). Which obviously not fully generic since we ideally want to be devnet agnostic here!

But first of all this is the easiest solution and we could find a much cleaner solution in some followup PRs (not sure about the best design to adopt here to make a compromise between the generic aspect and strong typing).

src/lean_spec/forks/lstar/spec.py
```diff
@@ -0,0 +1,31 @@
...
+from .containers.state import State
+from .store import Store
+
+
+class LstarSpec(ForkProtocol):
+    """Lstar fork."""
+
+    NAME: ClassVar[str] = "lstar"
+    VERSION: ClassVar[int] = 4
+    NETWORK_NAME: ClassVar[str] = "devnet0"
```
tcoratger, line 17: We should rename this to "lstar" probably. I haven't touch this since this is not the purpose of the PR but this is something we should do in a followup PR

unnawut (approved):
I didn't review line by line but from my skimming it's mostly wrapping the specs under a fork so it looks safe to me to merge and better to try this structure out earlier than later

PR #667, opened 2026-04-22
metrics: simplify and separate metrics logic from specs

unnawut:
## 🗒️ Description

The existing specs have metrics logic mixed with the core specs logic. The most visible example is the metrics code in the [fork choice store](https://github.com/leanEthereum/leanSpec/blob/d56e332dde7370a95e9c2587f801f9f171680c6c/src/lean_spec/subspecs/forkchoice/store.py#L515). This PR attempts to minimize the metrics code in the specs for better readability and also separation of the specs from other objectives.

Changes
- Introduce `observability` subspecs, the vendor-neutral version for reporting metrics. The specs now implement observability instead of prometheus-specific `metrics` subspecs.
  - Although, we still have `MetricsRegistry` containing all the metrics declaration. If this PR passes through I can move these into `observability` subspecs.
  - Once the `metrics` subspecs have only Prometheus-specific code, it can then be moved out of the subspecs whenever #638 settles how we will separate the specs vs. client code.
- Refactored fork choice store so instead of calling and calculating the metrics timing in the fork choice specs, the fork choice store only need to do a `with` statement e.g. `with observe_on_block():` which is a lot easier for specs reader to ignore.
- Also removed `_record_metrics()` from the fork choice specs which seems out of place. The metrics can be recorded by the caller (sync service) which is outside the core specs.
- Removed `BlockLookup` methods that were only used by the metrics but became part of the specs. The methods are moved into their respective callers

Overall our core specs now need not aware of the metrics implementation. It only needs to register through `with observe_on_attestation():` or `with observe_on_block():` and metrics can be calculated and published by the abstracted observer instead. Then we can configure that spec tests can ignore metrics by setting to NullObserver while the python client could inject `PrometheusObserver` into the observability module.

## ✅ Checklist

    ```console
    uvx tox
    ```


unnawut:
Btw. the `with` statement introduced quite a bit of indents. It's a lot easier to review by hiding whitespaces e.g. https://github.com/leanEthereum/leanSpec/pull/667/changes?w=1

tcoratger (approved):

PR #677, opened 2026-04-22
forkchoice: seed anchor checkpoints from the anchor slot

MegaRedHand:
## 🗒️ Description

`Store.from_anchor` previously built the store's justified/finalized checkpoints by copying the slot from the anchor state's embedded `latest_justified`/`latest_finalized` (which can be strictly less than the anchor slot) while overriding only the root to the anchor block root. That produced a slot/root-inconsistent checkpoint: the slot pointed at pre-anchor history while the root identified the anchor block itself.

This PR changes `Store.from_anchor` to use the anchor block as the justified/finalized checkpoints of the `Store`, following [the beacon chain specs](https://github.com/ethereum/consensus-specs/blob/master/specs/phase0/fork-choice.md#get_forkchoice_store).

Also updates the checkpoint-sync tests and a misleading inline comment that described the previous (pre-anchor-slot) behavior.

## ✅ Checklist

    ```console
    uvx tox
    ```
- [X] Considered updating the online docs in the [./docs/](/leanEthereum/leanSpec/tree/main/docs/) directory.

### DIFF
--- a/src/lean_spec/subspecs/forkchoice/store.py
+++ b/src/lean_spec/subspecs/forkchoice/store.py
@@ -210,18 +210,21 @@ class Store(StrictBaseModel):
         # Read the slot at which the anchor block was proposed.
         anchor_slot = anchor_block.slot
 
-        # Initialize checkpoints from this state.
+        # Seed both checkpoints from the anchor block itself.
         #
-        # We explicitly set the root to the anchor block root.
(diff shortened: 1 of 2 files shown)


MegaRedHand:
I believe the CI failure is unrelated

tcoratger (approved):

PR #716, opened 2026-05-13
fix: allow older-but-justified sources in `build_block`

MegaRedHand:
## Problem Summary

The fixed-point loop in `build_block` requires `att.source` to equal `current_justified` exactly. When the store's justified checkpoint has advanced via a sibling fork (e.g. block_6 justifying block_2 while the canonical head is on a fork at block_1), while the current head justified an earlier slot, the head chain's pool holds an attestation that *would* close the gap between them, but its source is different from the head's `latest_justified` checkpoint so it is filtered out and block production aborts with `Fixed-point attestation loop did not converge`.

## Reproduction

`tests/lean_spec/forks/lstar/forkchoice/test_block_production_justification_gap.py` builds the scenario explicitly:

```
                       block_4(4) -- block_5(5)  <-- head
                      /
genesis -- 1 -- 2 -- 3
                      \
                       block_6(6)
```

- `block_4`: 6/8 attest `target=block_1` (justifies slot 1).
- `block_5`: 2/8 attest `target=block_4` (fork-choice weight only).
- `block_6` (sibling of block_4): 6/8 attest `target=block_2` (justifies slot 2 on the store).

Fork choice keeps `block_5` as head; the store's `latest_justified` is now `block_2`, but `block_5`'s post-state still sits at `block_1`. Producing on slot 7 must close the gap.

## PR Description

This PR relaxes that filter to a slot bound: an attestation is selectable as long as its source slot is at or before the head chain's latest justified slot.

```diff
- if att_data.source != current_justified:
+ if att_data.source.slot > current_justified.slot:
      continue
```

Additionally, we also include some other, now required, sanity checks:
- `source.root` matches `source.slot`'s expected root on the chain
- `target.root` matches `target.slot`'s expected root on the chain
- the target isn't already justified (otherwise the attestation is useless), but we allow source=target=0 votes for bootstrapping fork-choice

There was a specific test for the check we removed (`test_build_block_skips_non_matching_source`), which was renamed to `test_build_block_skips_other_chain_source ` and repurposed for the new checks on source and target.

### DIFF
--- a/tests/lean_spec/forks/lstar/state/test_state_aggregation.py
+++ b/tests/lean_spec/forks/lstar/state/test_state_aggregation.py
@@ -275,11 +275,11 @@ def test_build_block_state_root_valid_when_signatures_split(
     assert len(result_state.validators.data) == num_validators
 
 
-def test_build_block_skips_non_matching_source(
+def test_build_block_skips_other_chain_source(
     container_key_manager: XmssKeyManager,
     spec: LstarSpec,
 ) -> None:
-    """Only attestation data whose source matches current_justified is included."""
+    """Only attestation data whose source matches the current chain is included."""
     state = make_keyed_genesis_state(2, container_key_manager)
     parent_header_with_state_root = state.latest_block_header.model_copy(
         update={"state_root": hash_tree_root(state)}
--- a/src/lean_spec/forks/lstar/spec.py
+++ b/src/lean_spec/forks/lstar/spec.py
@@ -1,7 +1,7 @@
 """Lstar fork — identity and construction facade."""
 
 from collections import defaultdict
-from collections.abc import Iterable, Sequence
+from collections.abc import Iterable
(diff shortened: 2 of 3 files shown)


tests/lean_spec/forks/lstar/state/test_state_aggregation.py
```diff
@@ -275,11 +275,11 @@ def test_build_block_state_root_valid_when_signatures_split(
     assert len(result_state.validators.data) == num_validators
 
 
-def test_build_block_skips_non_matching_source(
+def test_build_block_skips_other_chain_source(
```
MegaRedHand, line 278: The original test's purpose no longer applies, but the actual assertions do, so we renamed it.

pablodeymo (approved):

tcoratger (approved):
Thanks a lot :)

PR #717, opened 2026-05-14
Aggregated block proof - devnet5

anshalshukla:
## 🗒️ Description
Implements multi-message recursive block signature aggregation. All the attestation and block signatures are coalesced into a single block proof.

The blocks are deconstructed for additional attestation_data when they can help move the target. These new payloads are only aggregated when the chain is in sync as catching up nodes should not spam the network.

coverage-gate, fill-tests, and interop-tests all switched ubuntu → macos in this PR, and Justfile dropped -n auto from fill-ci. This might be due to inapt compilation target of linux bindings via leanMultisig-py which lead to 4x CI runs in linux based systems. Will be looked upon in followup PR.

## ✅ Checklist

- [X] Ran local quality checks to avoid unnecessary CI fails:
    ```console
    just check
    ```


src/lean_spec/forks/lstar/containers/block/block.py
```diff
@@ -66,21 +62,16 @@ class Block(Container):
...
+    binds every attestation in the body plus the proposer's signature over
+    the block root.
+    """
 
     block: Block
     """The block being signed."""
 
-    signature: BlockSignatures
-    """Aggregated signature payload for the block."""
+    proof: ByteListHalfMiB
```
g11tech: do we intend it to be extended to/aggregated with the CL/EL STF validity proof integration as well ? eventually

g11tech: lets rename ByteListHalfMiB to BlockProof

anshalshukla: I don't think we do, the leanVm is purposed for signature aggregation only so not sure if it can be extended further to include other STF validity proof. @tcoratger @TomWambsgans feel free to correct me here.

src/lean_spec/forks/lstar/spec.py
```diff
@@ -1296,6 +1261,19 @@ def on_block(
...
 
+            # The block body constrains how many distinct AttestationData
+            # entries it may carry.
+            aggregated_attestations = block.body.attestations
+            att_data_set = {att.data for att in aggregated_attestations}
+            assert len(att_data_set) == len(aggregated_attestations), (
+                "Block contains duplicate AttestationData entries; "
+                "each AttestationData must appear at most once"
+            )
+            assert len(att_data_set) <= int(MAX_ATTESTATIONS_DATA), (
```
g11tech, line 1285: should this already been not there in devnet4?

g11tech: nevermind, noticed its moved up from a section below

src/lean_spec/forks/lstar/containers/block/block.py
g11tech: also should we start calling SignedBlock, ProofedBlock or something?

g11tech: I think aggregating other proofs could be a natural extension, but yea lets see what others think about this

if this is just signature, lets stick with the terminology of signature itself instead of proof

TomWambsgans: LeanVM is a candidate for Post-Quantum Data-Availability (blobs). We could then aggregate the Consensus-Layer proof and the Data-Availability-Layer proof together, to stay at 1 proof / block. Enshrined proofs for the Execution-Layer is much harder.

src/lean_spec/subspecs/sync/service.py
```diff
@@ -712,6 +719,130 @@ async def publish_aggregated_attestation(
         """
         await self._publish_agg_fn(signed_attestation)
 
+    async def _maybe_publish_reaggregated_attestations_from_block(
+        self,
+        block: SignedBlock,
+    ) -> None:
```
g11tech: this is essentially publishing old attestations that an "aggregator" may not have,
imo this is not the best way to go, we should keep aggregator focused on new attestations to aggregate

while this splitting and repacking should essentially move to the proposer, because its possible and highly likely that proposer's local view is different from aggregators and instead of aggregators adding to the network load, its better that proposer split the attestations it need from the block by itself and restich them into the block

so everynode's forkchoice will maintain if it saw some attestation data in aggregated payloads or only in block (with index to that block) and then split and restich them if needed

g11tech (commented):
this is my major review of this PR:

https://github.com/leanEthereum/leanSpec/pull/717/changes#r3257938656

I think split and restitching belongs with proposer and not an aggregator, or atleast needs to be backed up by proposer as well incase it didn't see aggregator's restitched payload which could be a better solution I guess

so I recommend such split and restitching in block proposal build block as well

anshalshukla:
The only problem in restitching in block proposal as well is that it will further slow the block building process which will eventually lead to missed slots. Or we will have to bump up the spec for validators.

g11tech:
> The only problem in restitching in block proposal as well is that it will further slow the block building process which will eventually lead to missed slots. Or we will have to bump up the spec for validators.

we still need it as backup otherwise block proposal will fail if proposer can't repack attestations from a branch which moved justification/finalization

so we can still have / expect aggregators to do this job and do heavy lifting but in case of anything that is missing, we still need proposer to be able to do this.

expecting proposers to have a bit more compute itsn't too unrealistic imo, by 2028/2029 we might have more compute as well as optimizations available on a commodity hardware

anshalshukla:
Okay, I've updated it to do deconstruction in onBlock even in case of validators and not just in case of aggregators. I think it is better to continue to do it at the same place and update the store than to look for blocks which can help justify in the build_block function

src/lean_spec/forks/lstar/spec.py
```diff
@@ -862,88 +866,53 @@ def verify_signatures(
...
-            valid = self.sig_scheme.verify(
-                proposer.get_proposal_pubkey(),
-                block.slot,
-                block_root,
-                signatures.proposer_signature,
-            )
-        except (ValueError, IndexError):
-            valid = False
-        assert valid, "Proposer block signature verification failed"
+            type_two.verify(public_keys_per_message=public_keys_per_message)
```
tcoratger: This is the biggest thing I wanted to flag. `TypeTwoMultiSignature.verify(public_keys_per_message=...)` only checks that the supplied pubkey sets match the proof's internal pubkeys — the per-component `(message_i, slot_i)` pairs the snark binds live inside the proof bytes and never get compared against `hash_tree_root(att.data)` / `att.data.slot` from the block body. Compare with `aggregation.py:201-207` where Type-1's `verify` takes `message` and `slot` externally and the Rust side rejects on mismatch.

What this lets a proposer do: take honest Type-1 proofs `(M_i, slot_i, K_i)` and place them in the block with `att_i.data = D_i'` where `hash_tree_root(D_i') ≠ M_i`, as long as the bitfields still resolve to the same `K_i`. The merged Type-2 verifies (pubkeys match, snark is honest), but `process_attestations` then reads attacker-chosen `source/target/head/slot`. Validators get credited for votes they never cast.

Cleanest fix is either exposing `info` after `decompress_without_pubkeys` to Python so the spec can compare per-component, or pushing a `verify_type_2_with_messages(...)` variant into `lean_multisig_py` that does the comparison inside the binding. Probably worth a quick chat with the lean-multisig folks on which side of the boundary they'd rather keep it.

src/lean_spec/forks/lstar/spec.py
```diff
@@ -862,88 +866,53 @@ def verify_signatures(
...
+        except Exception as exc:
+            raise AssertionError(f"Block proof decoding failed: {exc}") from exc
 
-            # The signed message is the attestation data root.
-            # All validators in this group signed this exact data.
-            attestation_data_root = hash_tree_root(aggregated_attestation.data)
+        num_validators = Uint64(len(validators))
+        public_keys_per_message: list[list[PublicKey]] = []
 
+        # Attestation entries: parallel to block.body.attestations.
```
tcoratger: Comment block lost a chunk of pedagogical content without a code reason. Pre-PR was:

```
# Each attestation in the body must have a corresponding signature entry.
# Attestations and signatures are parallel arrays.
# - Each attestation says "validators X, Y, Z voted for this data".
# - Each signature proves those validators actually signed.
```

The "parallel arrays" explanation was the whole reason a new reader could follow the loop below. The original is still 100% accurate today — the loop still builds `public_keys_per_message` parallel to `block.body.attestations`. Restoring it would keep the diff tighter and the function self-explanatory.

src/lean_spec/forks/lstar/spec.py
```diff
@@ -862,88 +866,53 @@ def verify_signatures(
...
-                ) from exc
+            public_keys_per_message.append(
+                [validators[vid].get_attestation_pubkey() for vid in validator_ids]
+            )
 
-        # Verify the proposer's signature over the block root.
-        #
-        # The proposer signs hash_tree_root(block) with their proposal key.
-        # This proves the proposer endorsed this specific block.
+        # Proposer entry: bound to block root with a singleton participant set.
```
tcoratger: Similar to the comment above — the proposer-signature explanation got compressed quite hard. Pre-PR was:

```
# Verify the proposer's signature over the block root.
#
# The proposer signs hash_tree_root(block) with their proposal key.
# This proves the proposer endorsed this specific block.
```

New line reads more like an internal note than teaching content. "Bound to block root with a singleton participant set" is true but assumes the reader already knows what's happening. The longer version was newcomer-friendly and still accurate.

src/lean_spec/forks/lstar/spec.py
```diff
@@ -1305,41 +1287,20 @@ def on_block(
...
-                "each AttestationData must appear at most once"
-            )
-            assert Uint8(len(att_data_set)) <= MAX_ATTESTATIONS_DATA, (
-                f"Block contains {len(att_data_set)} distinct AttestationData entries; "
-                f"maximum is {MAX_ATTESTATIONS_DATA}"
-            )
-
             # Copy the aggregated proof map for updates
             # Shallow-copy the dict and its inner sets to preserve immutability
             # Block attestations go directly to "known" payloads
```
tcoratger: Two things tangled together here:

1. The comment says "Block attestations go directly to known payloads (like `is_from_block=True` in the spec)" — but the loop just below only does `block_proofs.setdefault(att.data, set())`. It inserts an empty set per attestation and never adds the per-attestation Type-1 proof. Pre-PR was `.setdefault(att.data, set()).add(proof)`. Downstream, `extract_attestations_from_aggregated_payloads` iterates `for proof in proofs` — an empty inner set yields zero votes, so block-imported attestations contribute nothing to fork-choice head/safe-target weight on receiving nodes.

2. The local-store write-back you added in `_maybe_publish_reaggregated_attestations_from_block` (8f9bdc1a) updates `latest_new_aggregated_payloads`, not `latest_known`, and only runs on the gossip path (see the inline at `sync/service.py:543`).

Either restoring the `.add(proof)` here (requires the per-attestation Type-1s to be available at this point — they aren't, since the block carries a merged Type-2), or routing the deconstruction through here so `latest_known` ends up populated, would close the gap. Until then, the comment is misleading.

src/lean_spec/subspecs/sync/service.py
```diff
@@ -712,6 +720,141 @@ async def publish_aggregated_attestation(
...
+        1. Extract that data's Type-1 proof out of the block's Type-2 proof.
+        2. Merge it with all local partial Type-1 proofs for the same data into
+           one Type-1 proof whose participant bits are the union.
+        3. Gossip the combined aggregate.
+
+        If the node has never seen the data locally, the extracted Type-1 is
+        gossiped as-is. This propagates the strongest available aggregate
+        without a full re-aggregation pass.
+        """
+        is_validtor = self.store.validator_id
```
tcoratger: Two small things:

- Typo: `is_validtor` → `is_validator`.
- `not validator_id` evaluates truthy when `validator_id == ValidatorIndex(0)` (it's a `Uint64` subclass, so `0` is falsy). So validator 0 silently skips deconstruction. The pre-fix line was `validator_id is not None and self.is_aggregator` which used `is not None` and didn't have this issue. A `validator_id is None` check keeps the semantics correct for index 0.

src/lean_spec/subspecs/sync/service.py
```diff
@@ -533,6 +540,7 @@ async def on_gossip_block(
             )
             self.store = new_store
             self._replay_pending_attestations()
+            await self._maybe_publish_reaggregated_attestations_from_block(block)
```
tcoratger: Just flagging: this is the only callsite for `_maybe_publish_reaggregated_attestations_from_block`. It's tied to gossip-block import — blocks that arrive via `_head_sync` / backfill don't go through here, so a node catching up never runs the deconstruction. Combined with the `latest_known_aggregated_payloads` empty-set seeding in `spec.py:1296-1299`, a freshly-syncing node ends up with zero block-imported attestation weight in fork choice. Probably worth covering the sync path too — closest spot is wherever the processed-store from `_head_sync` lands.

tests/lean_spec/subspecs/sync/test_reaggregate_from_block.py
```diff
@@ -0,0 +1,125 @@
...
+Exercises the decision/guard logic of
+`SyncService._maybe_publish_reaggregated_attestations_from_block`: an
+aggregator that imports a block extracts each attestation's Type-1 proof out
+of the merged Type-2 proof, merges it with locally held partial aggregates
+for the same data, and gossips the combined Type-1.
+
+These tests cover only the paths reachable without a Type-2 split. The split
+primitive (`split_type_2_by_msg`) is unsupported by the test-mode prover in
+the current `lean_multisig_py` build, so the proof-extraction path is not
+exercised here.
```
tcoratger: Heads up: this docstring is correct about the test-mode prover limitation, but it means none of the three tests below actually exercise the split-and-merge in `_maybe_publish_reaggregated_attestations_from_block`. They all assert `published == []` after early-exit gates, so the ~50 lines that do the actual work are uncovered. Worth mentioning in the PR description so it doesn't read as "feature is tested". A smoke test using prod-mode bindings, or waiting for the test-mode `split_type_2_by_msg` support, would close this.

tests/lean_spec/subspecs/sync/test_reaggregate_from_block.py
```diff
@@ -0,0 +1,125 @@
...
+    assert published == []
+
+
+async def test_noop_when_not_aggregator(peer_id: PeerId, key_manager: XmssKeyManager) -> None:
+    """A non-aggregator node never re-aggregates or publishes."""
+    base_store, signed_block, _ = _setup(
+        key_manager, block_participants=[ValidatorIndex(1), ValidatorIndex(2)]
+    )
+    service, published = _aggregator_service(peer_id, base_store)
+    service.is_aggregator = False
```
tcoratger: Post-`8f9bdc1a` the function no longer checks `self.is_aggregator` — the gate is now `if not self.store.validator_id: return`. So setting `service.is_aggregator = False` here doesn't actually exercise what the test name suggests. The test still passes because the early return happens for a different reason (or because of the `validator_id == 0` falsy bug — see the inline at `sync/service.py:743`). Worth renaming + reworking after the deconstruction gate is settled.

packages/testing/src/consensus_testing/keys.py
```diff
@@ -591,18 +595,19 @@ def build_attestation_signatures(
                 for vid in validator_ids
             ]
 
-            # Produce a single aggregated proof that the leanVM can verify
+            # Produce a single aggregated proof that the leanVm can verify
```
tcoratger: Tiny inconsistency: this comment now says `leanVm`, but elsewhere in the file (and in the codebase) it's `leanVM`. Pre-PR had `leanVM` here too. Just a one-letter casing slip in the reformat.

packages/testing/src/consensus_testing/test_types/block_spec.py
```diff
@@ -241,37 +243,92 @@ def build_attestations(
...
             )
+            proof_bytes = merged.encode_bytes()
         else:
-            proposer_signature = create_dummy_signature()
+            placeholder = ByteList512KiB(data=b"")
+            if not self.valid_signature:
+                # Burn the dummy signature creation to mirror the legacy
+                # shape; the merged blob carries no real bytes anyway.
+                dummy_signature = create_dummy_signature()
+                del dummy_signature
```
tcoratger: This block constructs `dummy_signature` only to `del` it, with a comment explaining it's there to mirror the legacy shape. The merged blob carries no real bytes anyway (the placeholder is already `b""`), so the construct doesn't affect anything. Removing the four lines + the import (if `create_dummy_signature` then becomes unused) would trim the diff.

tests/consensus/lstar/fc/test_attestation_target_selection.py
```diff
@@ -444,7 +444,7 @@ def test_attestation_target_justifiable_constraint(
     }
 
     steps = []
-    for i in range(1, 31):
+    for i in range(1, 16):
```
tcoratger, line 432: The loop shrank from `range(1, 31)` to `range(1, 16)`, but the `expected_targets` dict at lines 413-444 still has entries for slots 16-30. Those keys are now dead data — the test only walks 1-15. Either restoring the original range (and explaining the reason for the shrink) or trimming `expected_targets` to slots 1-15 would clean it up. As-is the slot 16-30 entries advertise coverage that isn't there (pronic 4×5=20, perfect squares 16/25, etc.).

.github/workflows/ci.yml
```diff
@@ -78,7 +78,7 @@ jobs:
 
   coverage-gate:
     name: Coverage gate - Python 3.12
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
```
tcoratger, line 87: Just flagging for the PR description: `coverage-gate`, `fill-tests`, and `interop-tests` all switched ubuntu → macos in this PR, and `Justfile` dropped `-n auto` from `fill-ci`. GH Actions macos runners are roughly 10× the ubuntu cost, so it'd be great to capture the why somewhere — if it's xdist racing on `setup_prover`'s global state in `lean_multisig_py`, that's a real upstream bug worth a follow-up issue rather than a CI workaround. (No action needed on the PR if there's a known reason, just a context line for future reviewers so this doesn't get unwound by accident.)

Justfile
```diff
@@ -83,7 +83,7 @@ test-consensus *args:
 # Canonical CI fixture run; contributors should use `uv run fill` directly
 [group('tests'), private]
 fill-ci *args:
-    uv run --group test fill --fork=Lstar --clean -n auto "$@"
+    uv run --group test fill --fork=Lstar --clean "$@"
```
tcoratger, line 90: Same context as the `ci.yml` note: dropping `-n auto` is a meaningful CI slowdown. If parallel runs of `fill` race somewhere (probably the `setup_prover(mode=mode)` global init in `lean_multisig_py`), that's worth tracking down — happy to look at it as a follow-up if helpful.

src/lean_spec/subspecs/xmss/aggregation.py
```diff
@@ -1,27 +1,37 @@
-"""Signature aggregation for the Lean Ethereum consensus specification."""
+"""
+Signature aggregation for the Lean Ethereum consensus specification.
+Multi-signature aggregation containers and helpers.
+
+Two proof shapes:
+
+- Type-1: many validators, one message (one AttestationData, or one block root).
+- Type-2: a merge of N Type-1 proofs, each over a distinct message.
+"""
```
tcoratger, line 8: Module docstring reads like two top-lines stacked:

```
Signature aggregation for the Lean Ethereum consensus specification.
Multi-signature aggregation containers and helpers.
```

Probably a merge artifact. Could collapse to one sentence + the Type-1/Type-2 explanation, e.g.:

```
Multi-signature aggregation for the Lean Ethereum consensus spec.

- Type-1: many validators on one message (one AttestationData, or one block root).
- Type-2: a merge of N Type-1 proofs over distinct messages.
```

packages/testing/src/consensus_testing/test_fixtures/verify_signatures.py
```diff
@@ -60,10 +55,6 @@ class VerifySignaturesTest(BaseConsensusFixture):
 
     Supported operations:
 
-    - `{"operation": "drop_last_signature"}`: Remove the last entry
-      from the block's attestation_signatures list. Produces a signed
-      block whose signature-group count is one less than its
-      attestation count.
     - `{"operation": "set_proposer_index", "value": int}`: Rewrite
```
tcoratger, line 66: Heads up: the `drop_last_signature` tamper op went away with this PR (and `tests/consensus/lstar/verify_signatures/test_structural_rejections.py` was deleted). The old test pinned "signature group count must match attestation count". The Type-2 envelope has no separate signature group anymore, but the analogue exists — "proof component count ≠ body attestations + 1" should be rejected, and that path isn't currently tested. A replacement tamper that truncates the proof blob (or otherwise breaks `TypeTwoMultiSignature.decode_bytes`) would keep the structural-rejection coverage from disappearing.

tcoratger (commented):
Hey @anshalshukla, thanks a lot for this. I've just went through a first careful review through a Claude discussion. I've noticed a couple of points that I think are valid. Please let me know what you think about them. I've let Claude pushed the comments since he is very clear and precise in his explanations.

-----

the 8f9bdc1a move out of the aggregator-only branch is good, but there's a small bug in the new gate (see the inline at `sync/service.py:743`). Also that helper currently only fires on the gossip path — blocks that arrive via sync/backfill skip it — so a node catching up still doesn't end up with per-attestation Type-1s.

**The verify_signatures binding** — the proof binds `(message_i, slot_i, pubkeys_i)` per component internally, but the spec never compares those internal `(message_i, slot_i)` against `hash_tree_root(att_i.data)` / `att_i.data.slot` from the block body. The snark verifies that the supplied pubkeys signed the proof's internal messages — not that those messages match the body. Details inline at `spec.py:913`. Worth a sync with the lean-multisig folks on whether they'd rather take a `verify_type_2_with_messages(...)` upstream, or expose `info` after `decompress_without_pubkeys` so the spec can compare in Python.

**The `latest_known_aggregated_payloads` seeding** — `on_block` still inserts empty sets per attestation (`spec.py:1296-1299`) and the comment block above still describes the pre-PR behavior. The local write-back you added in `_maybe_publish_reaggregated_attestations_from_block` lands in `latest_new_aggregated_payloads`, not `latest_known`, so on import the block-imported votes still don't feed fork-choice weight on the receiver. Inline at `spec.py:1292`.

**Test coverage** — `test_reaggregate_from_block.py`'s own header acknowledges the happy path can't be exercised because `split_type_2_by_msg` isn't in test-mode bindings yet. The three test bodies all assert `published == []`, so the ~50 lines that actually do the split/merge are uncovered. Might be worth a line in the PR description so it doesn't slip through as "tested".

**CI** — `coverage-gate`, `fill-tests`, `interop-tests` switched to macos-latest and `Justfile` dropped `-n auto` from `fill-ci`. GH Actions macos runners are roughly 10× the ubuntu cost — if the switch was working around a flake or OOM, would be great to capture the why in the PR description (xdist + `setup_prover`'s global side-effects in `lean_multisig_py` is the likely suspect).

The rest are smaller — comments/docstrings that drifted and a couple of typos. Left inline so they're easy to scroll through.

src/lean_spec/forks/lstar/store.py
```diff
@@ -128,18 +128,22 @@ class Store(StrictBaseModel, Generic[StateT, BlockT]):
...
     )
     """
     Aggregated signature proofs pending processing.
 
     These payloads are "new" and do not yet contribute to fork choice.
     They migrate to known payloads via interval ticks.
-    Populated from blocks or gossip aggregated attestations.
+    Populated from gossip aggregated attestations only; blocks no longer
+    feed individual proofs into this map (the block-level proof is a
+    merged Type-2 blob that the spec verifies as a whole). Aggregators will
```
tcoratger: Small docstring nit: "Aggregators will still extract the individual per-message proofs from the block-level Type-2 and gossip aggregated attestations." — future tense, but `_maybe_publish_reaggregated_attestations_from_block` exists now and (after 8f9bdc1a) runs for any validator, not just aggregators. Either drop the future tense and describe the current behavior, or trim to the part that's still aspirational, so the docstring matches what the code does today.

anshalshukla:
@tcoratger you can take a look again

Justfile
anshalshukla: I'll look at it as a followup, but I suspect it was due to leanMultisig high RAM demand as even in few of the local runs I encountered a slower test completion but def worth looking into setup_prover as well

tcoratger (approved):

PR #745, opened 2026-05-21
test framework: close download tempfile before extracting tarball

unnawut:
## 🗒️ Description

Prod key extraction on latest main branch [is failing](https://github.com/leanEthereum/leanSpec/actions/runs/26214049841/job/77135357943) because the previous flow held the temp file open while tarfile reopened the same path with a separate file descriptor. So there's chance that the tar file content was not fully written before it's being decompressed.

The new code closes the temp-file write context before opening it again for reading, so the kernel sees the full tarball on disk.


unnawut:
Merging without approval since this has low side effect

PR #748, opened 2026-05-21
lstar/aggregate: skip trivial 1-raw-sig + 0-children case

### LINKED ISSUE #747: lstar/aggregate: skip trivial 1-raw-sig + 0-children case to avoid pointless STARK proving
## Problem

`lstar/spec.py aggregate()` currently skips an `AttestationData` only when there are zero raw sigs and fewer than two children:

```python
# A lone child proof is already a valid proof — nothing to do.
if not raw_entries and len(child_proofs) < 2:
    continue
```

So with `1 raw sig + 0 children`, the spec aggregates and produces a `SignedAggregatedAttestation` whose proof is a recursive STARK over a single signature.

That work is wasted:

- The proof contains exactly one validator. It cannot move ⅔ quorum, so it gives consensus no signal it didn't already have.
- The raw XMSS sig is already on the per-subnet `attestation_signatures` gossip topic at sign time. Any peer aggregator can fold it in as a raw entry next round.
- The recursive STARK prover (`xmss/aggregation.py:aggregate`) is roughly constant-cost in input size — building a 1-validator proof costs the same as building a 32-validator proof.

## What we measured

On the multi-client devnet, two zeam aggregators (one validator each on their duty subnet, no peer payloads arriving in `latest_new_aggregated_payloads`) sat in the `1 raw + 0 children` case for 100% of slots over a 10-minute window. The result:

- Each aggregator spent ~10.8 s of FFI per worker run on a 1-validator "aggregate".
- ~50% of slot triggers got dropped as `in_flight` because the worker hadn't finished by the next slot.
- Every published aggregate covered exactly one validator.

Full numbers: blockblaz/zeam#907.

## Proposed change

Extend the skip predicate in `aggregate()`:

```python
# Skip cases where running the prover provides no consensus value:
#   - 0 raw + 0 children: nothing to aggregate.
#   - 0 raw + 1 child: lone child is already a valid proof; nothing to do.
#   - 1 raw + 0 children: a single-validator "aggregate" carries no
#     information the raw gossip sig doesn't already carry, and the
#     prover cost is constant in input size.
if not child_proofs and len(raw_entries) <= 1:
    continue
if not raw_entries and len(child_proofs) < 2:
    continue
```

Bookkeeping below the loop already keeps gossip sigs that were not consumed by an aggregation, so the lone sig stays in `store.attestation_signatures` and gets folded in by a future round once `raw_entries >= 2` or any child shows up.

## Why this is correctness-preserving

- The wire format and verifier are unchanged.
- The raw sig stays on the gossip topic — peers don't lose visibility of the vote.
- The store invariant (consumed sigs are pruned, untouched sigs remain) is unchanged.
- A spec-faithful client today produces a 1-validator aggregate that another aggregator could fold in as a child next round. With the change, peers still see the raw sig and fold it in directly. Same coverage, much cheaper.

## Acceptance criteria

- [ ] `aggregate()` returns no aggregation for the `1 raw + 0 children` case.
- [ ] Existing aggregation tests that happen to land on `1 raw + 0 children` are updated to `>= 2 raw` or include a child, with an explanatory comment.
- [ ] A new test asserts that `aggregate()` on a store with a single gossip sig and empty payload maps produces zero aggregations and leaves the sig in `store.attestation_signatures` for a future pass.

## Related

- blockblaz/zeam#907 — devnet investigation that surfaced this case
- blockblaz/zeam#908 — zeam-side fast path. Currently a deviation from this spec; will become spec-compliant once this change lands.

ch4r10t33r:
Closes #747.

## Summary

Extend the early-skip predicate in `lstar/spec.py:aggregate()`:

```python
# Skip cases where running the prover provides no consensus value:
#   - 0 raw + 0 children: nothing to aggregate.
#   - 0 raw + 1 child:    a lone child proof is already a valid proof.
#   - 1 raw + 0 children: a single-validator "aggregate" carries no
#                         information the raw gossip sig doesn't already
#                         carry — the sig is already on the gossip topic.
if not child_proofs and len(raw_entries) <= 1:
    continue
if not raw_entries and len(child_proofs) < 2:
    continue
```

## Why

A 1-validator aggregate can't move ⅔ quorum and the raw XMSS sig is already on the per-subnet `attestation_signatures` gossip topic at sign time, so any peer aggregator can fold it in as a raw entry next round. The recursive STARK prover is roughly constant-cost in input size — building a 1-validator proof costs the same as a 32-validator one.

On the multi-client devnet, two zeam aggregators sat in this case for 100% of slots over 10 minutes, spending ~10.8 s of FFI per worker run on single-validator proofs and dropping ~50% of slot triggers as `in_flight`. Full numbers in [blockblaz/zeam#907](https://github.com/blockblaz/zeam/issues/907).

## Correctness

- Wire format and verifier unchanged.
- The lone gossip sig stays in `store.attestation_signatures` because the existing bookkeeping only prunes sigs whose `att_data` produced an aggregate. A future round folds it in once another sig or a child shows up.
- Same end-state coverage as before, much cheaper to get there.

## Tests

- New: `test_aggregate_skips_single_gossip_sig_with_no_children` — asserts no aggregation is produced for the 1-sig + 0-children case and that the sig survives in the store.
- Updated: `test_multiple_attestation_data_grouped_separately` — seeded with two raw sigs per `att_data` so each group is non-trivial. The test's intent (two distinct `AttestationData` → two distinct proofs) is preserved.

`uv run pytest tests/lean_spec/forks/lstar/state tests/lean_spec/forks/lstar/forkchoice` → 47 passed.

## Related

- [blockblaz/zeam#907](https://github.com/blockblaz/zeam/issues/907) — devnet investigation
- [blockblaz/zeam#908](https://github.com/blockblaz/zeam/pull/908) — zeam-side fast path. Becomes spec-compliant once this lands.

### DIFF
--- a/packages/testing/src/consensus_testing/test_types/block_spec.py
+++ b/packages/testing/src/consensus_testing/test_types/block_spec.py
@@ -523,7 +523,7 @@ class BlockSpec(CamelModel):
         # Aggregation runs on a local clone: gossip pools mutate here, but the
         # caller's gossip-signature view must not be consumed by this simulated
         # build. Only the freshly aggregated Type-1 payloads propagate back.
-        aggregation_store, _ = spec.aggregate(store)
+        aggregation_store, _ = spec.aggregate(store, skip_trivial_inputs=False)
         merged_store = spec.accept_new_attestations(aggregation_store)
 
         # Build the block through the spec's State.build_block().
--- a/src/lean_spec/forks/lstar/spec.py
+++ b/src/lean_spec/forks/lstar/spec.py
@@ -1603,12 +1603,26 @@ class LstarSpec(ForkProtocol):
         # The head and attestation pools remain unchanged.
         return store.model_copy(update={"safe_target": safe_target})
 
-    def aggregate(self, store: LstarStore) -> tuple[LstarStore, list[SignedAggregatedAttestation]]:
+    def aggregate(
+        self,
+        store: LstarStore,
+        *,
+        skip_trivial_inputs: bool = True,
+    ) -> tuple[LstarStore, list[SignedAggregatedAttestation]]:
         """Turn raw validator votes into compact aggregated attestations.
 
         Validators cast individual signatures over gossip. Before those
         votes can influence fork choice or be included in a block, they
         must be combined into compact cryptographic proofs.
+
+        ``skip_trivial_inputs`` (default ``True``) is an **aggregator-role**
+        policy: when set, the ``1 raw + 0 children`` shape is skipped because
+        a single-validator "aggregate" carries no consensus signal beyond the
+        raw gossip sig already on the network (see issue #747). Interval-2
+        aggregator ticks use the default. Block-building callers that must
+        fold every chosen ``att_data`` into ``latest_known_aggregated_payloads``
+        (including lone gossip sigs) pass ``skip_trivial_inputs=False`` —
+        see ``BlockSpec`` in the consensus-testing filler.
 
         The store holds three pools of attestation evidence:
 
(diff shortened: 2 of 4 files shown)


koko1123:
@ch4r10t33r I owe you an apology — I picked #747 up from the issue tracker and completely missed this PR, and ended up submitting a duplicate implementation as #1191. Closing mine in deference to yours: you wrote the issue, measured the devnet impact, and implemented it first.

Main has moved quite a bit since May (the `spec/forks` restructure, and fork behavior is now tested via consensus vectors under `tests/consensus/` rather than pytest units), so in case it helps with the rebase, my branch is current against today's main and you should feel free to take anything from it, no attribution needed:

https://github.com/koko1123/leanSpec/tree/skip-trivial-single-signature-aggregation

A few notes from having effectively done that rebase, in case they save you time:

- `aggregate()` now lives in `src/lean_spec/spec/forks/lstar/aggregation.py`.
- The predicate can collapse to a single condition: `len(raw_signatures) + len(child_proofs) < 2` is exactly equivalent to the two-condition form in the issue — both skip precisely the `R + C <= 1` cases (verified by enumeration).
- On the harness dependency we both hit in `build_signed_block_with_store`: an alternative to the `skip_trivial_inputs` kwarg is to keep the spec signature unchanged and build the skipped single-signature proofs directly in the harness, mirroring how `_sign_block` already builds the 1-signature proposal component. That sidesteps any spec-purity question in review and keeps the 12 single-vote fork-choice scenarios' weights intact.
- The skip case can be covered with a fork-choice consensus vector: a single raw signature, a tick across interval 2 as aggregator, then assert the new-aggregate pool stays empty and the signature stays in the raw pool. There is a ready one on my branch (`test_interval_2_aggregator_skips_single_raw_signature`).

Happy to review the rebase or help any other way — this is worth landing either way for the zeam aggregators.

PR #753, opened 2026-05-21
metrics: block-proposal attestation build observability

ch4r10t33r:
## Summary

Adds cross-client `lean_block_proposal_*` metrics for block-proposal attestation selection (`build_block` fixed-point loop), aligned with [zeam #914](https://github.com/blockblaz/zeam/pull/914) `getProposalAttestations` instrumentation:

- `lean_block_proposal_attestation_build_phase_seconds{phase}` — `select_payloads`, `compact`, `stf_simulate`
- `lean_block_proposal_attestation_builds_total` — completed selection runs
- `lean_block_proposal_child_payloads_consumed_total` — greedy child payloads selected (before compaction)
- `lean_block_proposal_attestation_data_selected` — distinct `AttestationData` in the block body
- `lean_block_proposal_aggregates_selected` — aggregate proofs after compaction

**Note:** leanSpec runs `compact` once after the fixed-point loop; zeam runs compaction inside each loop iteration before candidate STF — phase timings are not directly comparable across clients.

## Test plan

- [x] `uv run pytest tests/lean_spec/subspecs/metrics/test_registry.py --no-cov`
- [ ] CI green on fork PR


src/lean_spec/forks/lstar/spec.py
```diff
@@ -794,11 +805,17 @@ def build_block(
                     post_state.latest_justified != current_justified
                     or post_state.latest_finalized.slot != current_finalized_slot
                 ):
+                    metrics.lean_block_proposal_attestation_build_phase_seconds.labels(
+                        phase="stf_simulate",
+                    ).observe(time.perf_counter() - stf_start)
```
turuslan, line 810: This line exists in both `if` and `else`.
Can it be placed before `if`?

PR #777, opened 2026-05-26
framework: cache decoded SecretKey instead of decode/encode bytes for every sign

unnawut:
## 🗒️ Description

The existing `_secret_state` cached re-encoded compact bytes (~2.7 KB per entry) to avoid retaining the full ~370 MB (pre-#776, they're even smaller now) decoded Python tree. But on every sign, the cache hit still ran `SecretKey.decode_bytes`, which is as expensive as decoding from JSON (~4 s per call in prod scheme), and each sign also re-encoded back to compact bytes.

This PR makes the key manager hold the decoded object directly in `_secret_state` instead. First sign per (validator, role) per worker still pays the JSON-to-object cost once, but subsequent signs reuse the live object and just advance and sign in microseconds.

Since the github runner has 4 cores and 14GB ram, we'd be using around `4 workers * 8 keys * 370 MB = 11.84 GB` assuming all workers initialize all validators. And most test cases use up to 4 and only a few use 8, so there's still quite a bit of breathing room.

Timing on my machine is ~24% faster (before: 14:58m, after: 11:22m). [Tested on my fork](https://github.com/unnawut/leanSpec/actions/runs/26430749502), the CI goes down from ~1h to 32m.

### DIFF
--- a/packages/testing/src/consensus_testing/keys.py
+++ b/packages/testing/src/consensus_testing/keys.py
@@ -180,7 +180,7 @@ class XmssKeyManager:
 
     - Raw JSON (lightweight hex strings, ~2.7 KB per validator)
     - Deserialized public keys only (avoids the heavy secret key objects)
-    - Advanced secret key state as compact SSZ bytes
+    - Advanced secret key state as live Python objects
     """
 
     __slots__ = (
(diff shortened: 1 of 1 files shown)


unnawut:
merging since this is low risk CI change

PR #781, opened 2026-05-26
fix: keep `latest_justified` monotonic across same-block attestations

MegaRedHand:
## Summary

- Adds a state-transition regression test for a block carrying three supermajority attestations targeting slots 4, 9, and 6 in that body order. Each slot is justifiable from finalized genesis (immediate / square / pronic), and each aggregate has 3/4 votes.
- Patches `process_attestations` so `latest_justified` only advances forward. Before the fix, attestations resolved in body order and unconditionally overwrote `latest_justified`, so the slot-6 attestation dragged the checkpoint from slot 9 back to slot 6.
- Order is forced via `forced_attestations` because the block builder normally sorts attestations by `target.slot` ascending. Slots 4 and 9 go through the builder; slot 6 is appended last so the on-chain body order becomes `[4, 9, 6]`.

The per-slot `justified_slots` bitfield is unchanged: each justified target is still recorded. Only the "current checkpoint" pointer is now monotonic in slot.

## Test plan

- [x] New test `test_same_block_multi_target_attestations_advance_to_highest_slot` fails on the unpatched spec (`latest_justified_slot = 6, expected 9`) and passes with the patch.
- [x] Full `tests/consensus/lstar/state_transition/test_justification.py` + `test_finalization.py` suite: 31/31 pass with the patch.
- [x] `uv run fill --fork=lstar` generates the fixture for the new test after the fix.
- [x] `ruff check` + `ruff format --check` clean on modified files.

### DIFF
--- a/src/lean_spec/forks/lstar/spec.py
+++ b/src/lean_spec/forks/lstar/spec.py
@@ -536,7 +536,12 @@ class LstarSpec(ForkProtocol):
                 # The block becomes justified
                 #
                 # The chain now considers this block part of its safe head.
-                latest_justified = target
+                # Only advance the checkpoint forward.
+                # Attestations within a block can resolve in any order, and
+                # an earlier target processed after a later one must not
+                # drag latest_justified backwards.
+                if target.slot > latest_justified.slot:
+                    latest_justified = target
                 justified_slots = justified_slots.with_justified(
                     finalized_slot,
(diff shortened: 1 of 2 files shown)


pablodeymo (approved):
LGTM

tcoratger (approved):
Thanks a lot!

PR #791, opened 2026-05-28
feat: tiered attestation scoring

MegaRedHand:
## 🗒️ Description

Previously, we relaxed the block building algorithm to also take into account attestations with an older source. This fixed a devnet breaking bug but also made block building much less efficient, with useless attestations now being packed in blocks.

This PR changes the block building algorithm to order attestations by usefulness, with attestations that finalize a slot being included before any that only justify a slot, and those being preferred before attestations that only add weight. This is implemented by grouping attestations by tiers (Finalize, Justify, Build), and scoring them by coverage.

## ✅ Checklist

- [X] Ran local quality checks to avoid unnecessary CI fails:
    ```console
    just check
    ```


src/lean_spec/spec/forks/lstar/containers/state.py
```diff
@@ -97,6 +138,18 @@ def extend_to_slot(self, finalized_slot: Slot, target_slot: Slot) -> Self:
...
+        """
+        Advance the tracking window by dropping slots that became finalized.
+
+        A non-positive delta keeps the tracking window unchanged.
+        """
+        if delta <= 0:
+            return self
+
+        # Drop the leading entries that fell behind the finalized boundary.
+        return type(self)(data=self.data[delta:])
```
tcoratger: I think that we removed it very recently, probably not useful for just one usage

src/lean_spec/spec/forks/lstar/containers/state.py
```diff
@@ -67,6 +67,47 @@ def is_slot_justified(self, finalized_slot: Slot, target_slot: Slot) -> Boolean:
...
+            raise IndexError(
+                f"Slot {target_slot} is outside the tracked range "
+                f"(finalized_boundary={finalized_slot}, tracked_length={len(self)})"
+            )
+
+        # Clone the data, flip the one bit, and wrap it in a new instance.
+        new_data = list(self.data)
+        new_data[relative_index] = value
+
+        return type(self)(data=new_data)
```
tcoratger: Since this is used only 1 time, probably better to inline the logic directly from a readability standpoint.

src/lean_spec/spec/forks/lstar/block_production.py
```diff
@@ -13,19 +15,363 @@
...
+
+
+def _is_genesis_self_vote(attestation_data: AttestationData) -> bool:
+    """
+    Whether the vote anchors both source and target at slot 0.
+
+    Genesis self-votes cannot justify or finalize, but they carry
+    fork-choice signal, so selection treats them specially.
+    """
+    return attestation_data.source.slot == Slot(0) and attestation_data.target.slot == Slot(0)
```
tcoratger: Maybe for this also we can inline the logic directly at the appropriate locations and not have a dedicated function.

src/lean_spec/spec/forks/lstar/block_production.py
```diff
@@ -13,19 +15,363 @@
...
+            best_candidate = (attestation_data, entry_score, new_voters)
+            best_candidate_key = candidate_key
+
+    return best_candidate
 
 
 class BlockProductionMixin(LstarSpecBase):
     """Proposer-side block building for the lstar fork."""
 
+    def _select_attestations(
```
tcoratger: I think that for this, this is used only 1 time, this is better to just inline the logic directly at the location, like we did previously

tcoratger (commented):
In general my comment here is that we should minimize the number of functions. For example, when a function is used only one time, this probably mean that we should delete it and inline the logic at the appropriate location directly.

I've seen also a couple of standalone functions outside of the objects. Most of the time this may be confusing for the reader since everything else is inside dedicated objects. You can check the py architect subagent if you have Claude, he should be used to this kind of architecture.

src/lean_spec/spec/forks/lstar/containers/state.py
MegaRedHand: Done

src/lean_spec/spec/forks/lstar/containers/state.py
MegaRedHand: Done

src/lean_spec/spec/forks/lstar/block_production.py
MegaRedHand: Done

src/lean_spec/spec/forks/lstar/block_production.py
MegaRedHand: I inlined it, along with the filter and scoring helpers

MegaRedHand:
@tcoratger I moved the new algorithm as an alternative to `BlockProductionMixin` in #1149. I'll leave this in draft since the diff might be useful for reviewers

PR #820, opened 2026-06-02
fix: ignore stale finalized payloads in fork choice

latifkasuli:
## Summary

Fixes a fork-choice vote extraction edge case where a stale aggregated payload whose attested head is already finalized could mask the same validator's still-relevant vote above `latest_finalized`.

Fork-choice consumers now filter payloads by `attestation_data.head.slot > latest_finalized.slot` before doing the per-validator latest-message selection. Votes whose target is finalized still count when their head is above finalization, since fork-choice weight follows the attested head.

`prune_stale_attestation_data()` uses the same head-slot boundary so pruning and fork-choice extraction stay aligned.

## Root Cause

Fork choice extracts one latest attestation per validator by comparing `AttestationData.slot`. If a validator has:

- a fresh vote above finalization, and
- a later-slot stale payload whose head is already finalized,

the stale payload previously won the latest-message comparison. Since fork-choice weight accumulation stops at `latest_finalized`, that selected stale payload contributes no weight above the finalized boundary, effectively hiding the fresh vote.

Filtering by target slot was too strong because an attestation can have a finalized target while its head remains above finalization and fork-choice relevant.

## Validation

- `uv run pytest tests/lean_spec/spec/forks/lstar/forkchoice -q --no-cov`
- `uv run pytest tests/lean_spec/spec/forks/lstar/forkchoice/test_validator.py tests/lean_spec/spec/forks/lstar/state/test_state_aggregation.py -q --no-cov`
- `uv run --group test fill tests/consensus/lstar/fc/test_store_pruning.py --fork=Lstar --clean -q`
- `uv run --group lint ruff check src/lean_spec/spec/forks/lstar/fork_choice.py tests/lean_spec/spec/forks/lstar/forkchoice/test_compute_block_weights.py tests/lean_spec/spec/forks/lstar/forkchoice/test_store_pruning.py`
- `uv run --group lint ruff format --check src/lean_spec/spec/forks/lstar/fork_choice.py tests/lean_spec/spec/forks/lstar/forkchoice/test_compute_block_weights.py tests/lean_spec/spec/forks/lstar/forkchoice/test_store_pruning.py`
- `uv run --group lint ty check src/lean_spec/spec/forks/lstar/fork_choice.py tests/lean_spec/spec/forks/lstar/forkchoice/test_compute_block_weights.py tests/lean_spec/spec/forks/lstar/forkchoice/test_store_pruning.py`
- `git diff --check`

### DIFF
--- a/src/lean_spec/spec/forks/lstar/fork_choice.py
+++ b/src/lean_spec/spec/forks/lstar/fork_choice.py
@@ -434,17 +434,6 @@ class ForkChoiceMixin(LstarSpecContract):
                         attestations[validator_index] = attestation_data
         return attestations
 
-    def _fork_choice_payloads(
-        self,
-        store: LstarStore,
-        aggregated_payloads: dict[AttestationData, set[SingleMessageAggregate]],
-    ) -> dict[AttestationData, set[SingleMessageAggregate]]:
-        return {
-            attestation_data: proofs
-            for attestation_data, proofs in aggregated_payloads.items()
-            if attestation_data.target.slot > store.latest_finalized.slot
-        }
-
     def _accumulate_ancestor_weights(
         self,
         store: LstarStore,
@@ -478,7 +467,12 @@ class ForkChoiceMixin(LstarSpecContract):
         for every ancestor above the finalized slot.
         """
         attestations = self.extract_attestations_from_aggregated_payloads(
-            store, self._fork_choice_payloads(store, store.latest_known_aggregated_payloads)
+            store,
+            {
+                attestation_data: proofs
+                for attestation_data, proofs in store.latest_known_aggregated_payloads.items()
+                if attestation_data.target.slot > store.latest_finalized.slot
+            },
         )
 
         weights = self._accumulate_ancestor_weights(
@@ -560,7 +554,12 @@ class ForkChoiceMixin(LstarSpecContract):
         """
         # Extract attestations from known aggregated payloads
         attestations = self.extract_attestations_from_aggregated_payloads(
-            store, self._fork_choice_payloads(store, store.latest_known_aggregated_payloads)
+            store,
+            {
+                attestation_data: proofs
+                for attestation_data, proofs in store.latest_known_aggregated_payloads.items()
+                if attestation_data.target.slot > store.latest_finalized.slot
+            },
         )
 
         # Run LMD-GHOST fork choice algorithm.
@@ -647,7 +646,11 @@ class ForkChoiceMixin(LstarSpecContract):
         # "Known" is excluded by design.
         attestations = self.extract_attestations_from_aggregated_payloads(
             store,
-            self._fork_choice_payloads(store, store.latest_new_aggregated_payloads),
+            {
+                attestation_data: proofs
+                for attestation_data, proofs in store.latest_new_aggregated_payloads.items()
+                if attestation_data.target.slot > store.latest_finalized.slot
+            },
         )
 
         # Run LMD GHOST with the supermajority threshold.
--- a/src/lean_spec/spec/forks/lstar/fork_choice.py
+++ b/src/lean_spec/spec/forks/lstar/fork_choice.py
@@ -434,6 +434,17 @@ class ForkChoiceMixin(LstarSpecContract):
                         attestations[validator_index] = attestation_data
         return attestations
 
+    def _fork_choice_payloads(
+        self,
+        store: LstarStore,
+        aggregated_payloads: dict[AttestationData, set[SingleMessageAggregate]],
+    ) -> dict[AttestationData, set[SingleMessageAggregate]]:
+        return {
+            attestation_data: proofs
+            for attestation_data, proofs in aggregated_payloads.items()
+            if attestation_data.target.slot > store.latest_finalized.slot
+        }
+
     def _accumulate_ancestor_weights(
         self,
         store: LstarStore,
@@ -467,7 +478,7 @@ class ForkChoiceMixin(LstarSpecContract):
         for every ancestor above the finalized slot.
         """
         attestations = self.extract_attestations_from_aggregated_payloads(
-            store, store.latest_known_aggregated_payloads
+            store, self._fork_choice_payloads(store, store.latest_known_aggregated_payloads)
         )
 
         weights = self._accumulate_ancestor_weights(
@@ -549,7 +560,7 @@ class ForkChoiceMixin(LstarSpecContract):
         """
         # Extract attestations from known aggregated payloads
         attestations = self.extract_attestations_from_aggregated_payloads(
-            store, store.latest_known_aggregated_payloads
+            store, self._fork_choice_payloads(store, store.latest_known_aggregated_payloads)
         )
 
         # Run LMD-GHOST fork choice algorithm.
@@ -636,7 +647,7 @@ class ForkChoiceMixin(LstarSpecContract):
         # "Known" is excluded by design.
         attestations = self.extract_attestations_from_aggregated_payloads(
             store,
-            store.latest_new_aggregated_payloads,
+            self._fork_choice_payloads(store, store.latest_new_aggregated_payloads),
         )
 
         # Run LMD GHOST with the supermajority threshold.
--- a/src/lean_spec/spec/forks/lstar/fork_choice.py
+++ b/src/lean_spec/spec/forks/lstar/fork_choice.py
@@ -102,9 +102,9 @@ class ForkChoiceMixin(LstarSpecContract):
     def prune_stale_attestation_data(self, store: LstarStore) -> LstarStore:
         """Remove attestation data that can no longer influence fork choice.
 
-        An attestation becomes stale when its target checkpoint falls at or before
+        An attestation becomes stale when its head checkpoint falls at or before
         the finalized slot. Such attestations cannot affect chain selection since
-        the target is already finalized.
+        the head is already finalized.
 
         Pruning removes all attestation-related data:
 
@@ -114,22 +114,22 @@ class ForkChoiceMixin(LstarSpecContract):
         """
         # Filter out stale entries from all attestation-related mappings.
         #
-        # Each mapping is keyed by attestation data, so we check membership by slot
-        # against the finalized slot.
+        # Each mapping is keyed by attestation data, so we check the attested
+        # head slot against the finalized slot.
         store.attestation_signatures = {
             attestation_data: signatures
             for attestation_data, signatures in store.attestation_signatures.items()
-            if attestation_data.target.slot > store.latest_finalized.slot
+            if attestation_data.head.slot > store.latest_finalized.slot
         }
         store.latest_new_aggregated_payloads = {
             attestation_data: proofs
             for attestation_data, proofs in store.latest_new_aggregated_payloads.items()
-            if attestation_data.target.slot > store.latest_finalized.slot
+            if attestation_data.head.slot > store.latest_finalized.slot
         }
         store.latest_known_aggregated_payloads = {
             attestation_data: proofs
             for attestation_data, proofs in store.latest_known_aggregated_payloads.items()
-            if attestation_data.target.slot > store.latest_finalized.slot
+            if attestation_data.head.slot > store.latest_finalized.slot
         }
         return store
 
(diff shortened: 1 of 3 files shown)


src/lean_spec/spec/forks/lstar/fork_choice.py
```diff
@@ -434,6 +434,17 @@ def extract_attestations_from_aggregated_payloads(
...
+    def _fork_choice_payloads(
+        self,
+        store: LstarStore,
+        aggregated_payloads: dict[AttestationData, set[SingleMessageAggregate]],
+    ) -> dict[AttestationData, set[SingleMessageAggregate]]:
+        return {
+            attestation_data: proofs
+            for attestation_data, proofs in aggregated_payloads.items()
+            if attestation_data.target.slot > store.latest_finalized.slot
+        }
```
tcoratger: Can we avoid having a super small function like this in the spec? I think from a readability standpoint (even if this is a bit of code duplication), in the spec context, this is better to just inline the logic at the 3 appropriate locations?

tcoratger (commented):
Thanks a lot!

Please let me know when the PR is ready for review? 

I think that you PRs are on draft, please put them as ready for review when you feel they are ready so that someone can have a look :)

latifkasuli:
Addressed the review comment in `8beaa56` by removing `_fork_choice_payloads()` and inlining the finalized-slot filter at the three fork-choice call sites.

Validation:
- `uv run pytest tests/lean_spec/spec/forks/lstar/forkchoice/test_compute_block_weights.py -k stale_latest_message -q --no-cov`
- `uv run pytest tests/lean_spec/spec/forks/lstar/forkchoice/test_compute_block_weights.py tests/lean_spec/spec/forks/lstar/forkchoice/test_store_pruning.py tests/lean_spec/spec/forks/lstar/forkchoice/test_attestation_target.py tests/lean_spec/spec/forks/lstar/forkchoice/test_store_attestations.py -q --no-cov`
- `uv run --group lint ruff check src/lean_spec/spec/forks/lstar/fork_choice.py tests/lean_spec/spec/forks/lstar/forkchoice/test_compute_block_weights.py`
- `uv run --group lint ruff format --check src/lean_spec/spec/forks/lstar/fork_choice.py tests/lean_spec/spec/forks/lstar/forkchoice/test_compute_block_weights.py`
- `uv run --group lint ty check src/lean_spec/spec/forks/lstar/fork_choice.py tests/lean_spec/spec/forks/lstar/forkchoice/test_compute_block_weights.py`
- `git diff --check`

src/lean_spec/spec/forks/lstar/fork_choice.py
```diff
@@ -467,7 +467,12 @@ def compute_block_weights(self, store: LstarStore) -> dict[Bytes32, int]:
         for every ancestor above the finalized slot.
         """
         attestations = self.extract_attestations_from_aggregated_payloads(
-            store, store.latest_known_aggregated_payloads
+            store,
+            {
+                attestation_data: proofs
+                for attestation_data, proofs in store.latest_known_aggregated_payloads.items()
+                if attestation_data.target.slot > store.latest_finalized.slot
```
tcoratger: This is the heart of the fix, and I want to think out loud with you about one detail. 🙂

The filter keeps a vote when its **target** is above the finalized slot. But just below, the weight comes from `_accumulate_ancestor_weights`, which walks upward from the vote's **head**. So the field we *filter on* (target) and the field that *actually earns weight* (head) are different — and they're not always the same slot.

Remember `target` is chosen by walking **backwards** from `head` (in `validator_duties.py`, `get_attestation_target`), so `target.slot <= head.slot`, sometimes strictly less. And `Slot.is_justifiable_after` treats distance `0` as justifiable, so a target landing *exactly on* the finalized slot is perfectly valid.

Here's a tiny picture of where it bites. Imagine slots 2–5 were skipped (totally normal):

```
slot:    0          1                      6
       genesis ── block_F ───────────── block_H
                 (finalized)              (head)
```

An honest validator voting for `block_H` walks back from slot 6 looking for a justifiable ancestor, and the only one it finds is `block_F` at the finalized slot. So its vote is:

```
head   = block_H @ 6   (above finalized — should add weight)
target = block_F @ 1   (exactly at finalized — still a valid target)
```

`validate_attestation` happily accepts this (it only asks `head.slot >= target.slot`). Now watch the two sides disagree:

```
weight walk:   block_H @ 6  →  6 > 1 ✓  → +1 to block_H     (this vote *should* count)
this filter:   target.slot 1 > finalized 1  →  False  → vote dropped  (so it never does)
```

I actually ran this against the branch and `compute_block_weights` returns `{}` instead of `{block_H: 1}` — the honest vote's weight just disappears, which can tip head selection.

What do you think about filtering on the field the weight function actually uses?

```python
if attestation_data.head.slot > store.latest_finalized.slot
```

The nice part: this still kills the bug you're fixing. In your test the stale vote has `head == target == finalized`, so `head.slot > finalized.slot` is also false and it's still dropped. I tried it locally and both your test and the little scenario above go green together. 🎯

One more breadcrumb: `prune_stale_attestation_data` (a bit higher up, not in this diff) uses the same `target.slot`-based rule, and its docstring says such votes "cannot affect chain selection" — which is the exact assumption the example above bends. So if we do move to `head`, it'd be worth aligning that spot too.

And the honest open question: if forfeiting weight for a finalized-target vote is actually an *intended* 3SF rule, then the code is fine and it's the docstrings that should say so out loud. Do you happen to know which way that's meant to go?

src/lean_spec/spec/forks/lstar/fork_choice.py
```diff
@@ -549,7 +554,12 @@ def update_head(self, store: LstarStore) -> LstarStore:
         """
         # Extract attestations from known aggregated payloads
         attestations = self.extract_attestations_from_aggregated_payloads(
-            store, store.latest_known_aggregated_payloads
+            store,
+            {
+                attestation_data: proofs
+                for attestation_data, proofs in store.latest_known_aggregated_payloads.items()
+                if attestation_data.target.slot > store.latest_finalized.slot
```
tcoratger: Same underlying thought as the comment over in `compute_block_weights` — this filter keys on `target.slot` while the weight walk below keys on `head.slot`, so it'd be good to keep them in sync once we settle the target-vs-head question. 🙂

Two small extra things while we're here:
- there's no regression test that drives this masking scenario through head selection specifically, so a guard test here would be lovely (sketch is in the test-file comment);
- the line just above ("Extract attestations from known aggregated payloads") describes the call but not the new filtering, so it might be a nice spot to say *why* we drop these votes.

src/lean_spec/spec/forks/lstar/fork_choice.py
```diff
@@ -636,7 +646,11 @@ def update_safe_target(self, store: LstarStore) -> LstarStore:
         # "Known" is excluded by design.
         attestations = self.extract_attestations_from_aggregated_payloads(
             store,
-            store.latest_new_aggregated_payloads,
+            {
+                attestation_data: proofs
+                for attestation_data, proofs in store.latest_new_aggregated_payloads.items()
+                if attestation_data.target.slot > store.latest_finalized.slot
```
tcoratger: And the third sibling — same `target` vs `head` thought applies on the "new" pool too.

This one feeds the safe target, which is what validators lean on to decide what's safe to attest to, so dropping a legitimately-weighted vote here has a slightly higher blast radius. A small guard test for safe-target would be reassuring once we agree on the rule. 🌿

tests/lean_spec/spec/forks/lstar/forkchoice/test_compute_block_weights.py
```diff
@@ -81,6 +81,65 @@ def test_linear_chain_weight_accumulates_upward(spec: LstarSpec, base_store: Sto
...
+    )
+
+    base_store.latest_known_aggregated_payloads = {
+        fresh_vote: {_make_empty_proof([ValidatorIndex(0)])},
+        stale_vote: {_make_empty_proof([ValidatorIndex(0)])},
+    }
+
+    weights = spec.compute_block_weights(base_store)
+
+    assert weights == {block2_root: 1}
```
tcoratger, line 223: This test is a great anchor for the bug you're fixing — it's exactly the "stale message masks a fresh one" case, so I'd keep it as-is. 👍

To round things out, here's a companion **test vector** I'd suggest adding right alongside it. It encodes the mirror-image situation from my inline note: a vote whose head is above finalized but whose target sits at the finalized slot. It should still earn weight.

Heads up: on the branch **as it stands today this test fails** (`compute_block_weights` returns `{}`), which is precisely the behavior I was flagging. Once the filter keys on `head.slot`, it passes — so it doubles as the regression guard that keeps this from creeping back later. 🙂

```python
def test_fresh_head_with_finalized_target_still_counts(
    spec: LstarSpec, base_store: Store
) -> None:
    """A vote with a head above finalized still adds weight when its target sits at the finalized slot."""
    genesis_root = base_store.head

    block1 = make_signed_block(
        slot=Slot(1),
        proposer_index=ValidatorIndex(0),
        parent_root=genesis_root,
        state_root=make_bytes32(10),
    )
    block1_root = hash_tree_root(block1.block)

    block2 = make_signed_block(
        slot=Slot(2),
        proposer_index=ValidatorIndex(1),
        parent_root=block1_root,
        state_root=make_bytes32(20),
    )
    block2_root = hash_tree_root(block2.block)

    base_store.blocks = {
        **base_store.blocks,
        block1_root: block1.block,
        block2_root: block2.block,
    }
    genesis_state = base_store.states[genesis_root]
    base_store.states = {
        **base_store.states,
        block1_root: genesis_state,
        block2_root: genesis_state,
    }
    base_store.head = block2_root
    base_store.latest_finalized = Checkpoint(root=block1_root, slot=Slot(1))

    # Head is above the finalized slot, so the upward walk reaches block2 and adds weight.
    # Target sits exactly at the finalized slot, which is still a valid, justifiable target.
    fresh_vote_with_old_target = AttestationData(
        slot=Slot(2),
        head=Checkpoint(root=block2_root, slot=Slot(2)),
        target=Checkpoint(root=block1_root, slot=Slot(1)),
        source=Checkpoint(root=genesis_root, slot=Slot(0)),
    )
    base_store.latest_known_aggregated_payloads = {
        fresh_vote_with_old_target: {_make_empty_proof([ValidatorIndex(0)])},
    }

    weights = spec.compute_block_weights(base_store)

    assert weights == {block2_root: 1}
```

And if you wanted to take the maintainer's earlier "maybe a test vector?" nudge all the way to cross-client land, both of these (the masking case and this mirror case) would make lovely consensus fillers under `tests/consensus/lstar/fc/`, modeled on `test_store_pruning.py` — build the finalized chain with `BlockStep`, drop the aggregates in with `GossipAggregatedAttestationStep`, then assert the chosen head with `StoreChecks`. Totally optional, just flagging the path. 🙏

tcoratger (commented):
Hey @latifkasuli, thanks a lot for chasing this one down!


I've discussed with Claude locally and I think I've couple of comments here

latifkasuli:
Addressed the finalized-target review by moving the fork-choice boundary from target slot to head slot:

- `compute_block_weights()`, `update_head()`, and `update_safe_target()` now filter on `attestation_data.head.slot > store.latest_finalized.slot`.
- `prune_stale_attestation_data()` now uses the same head-slot predicate and its docstring/comments are aligned with that rule.
- Added regressions for the case where `target == latest_finalized` but `head > latest_finalized`, covering block weights, `update_head()`, `update_safe_target()`, and pruning.

Validation run locally:

- `uv run pytest tests/lean_spec/spec/forks/lstar/forkchoice -q --no-cov`
- `uv run pytest tests/lean_spec/spec/forks/lstar/forkchoice/test_validator.py tests/lean_spec/spec/forks/lstar/state/test_state_aggregation.py -q --no-cov`
- `uv run --group test fill tests/consensus/lstar/fc/test_store_pruning.py --fork=Lstar --clean -q`
- `uv run --group lint ruff check src/lean_spec/spec/forks/lstar/fork_choice.py tests/lean_spec/spec/forks/lstar/forkchoice/test_compute_block_weights.py tests/lean_spec/spec/forks/lstar/forkchoice/test_store_pruning.py`
- `uv run --group lint ruff format --check src/lean_spec/spec/forks/lstar/fork_choice.py tests/lean_spec/spec/forks/lstar/forkchoice/test_compute_block_weights.py tests/lean_spec/spec/forks/lstar/forkchoice/test_store_pruning.py`
- `uv run --group lint ty check src/lean_spec/spec/forks/lstar/fork_choice.py tests/lean_spec/spec/forks/lstar/forkchoice/test_compute_block_weights.py tests/lean_spec/spec/forks/lstar/forkchoice/test_store_pruning.py`
- `git diff --check`

tcoratger (approved):

PR #854, opened 2026-06-06
Fix production fixture archive generation

KolbyML:
## 🗒️ Description
Test assets were being generated without assets for test asset releases.

I believe this will fix it, but I have no way to test it


unnawut (approved):
the fix works for me!

PR #875, opened 2026-06-07
refactor(testing): scenario builder and shared selective-check base

tcoratger:
## Summary

PR 10 from the `packages/testing/` audit — fork choice authoring ergonomics, in two parts. Net **−245 lines**.

### 1. Dedup the selective-validation models

`StoreChecks` drove its scalar, label-root, and pool target-slot checks through three parallel `if "field" in fields:` ladders (~75 lines); `AttestationCheck` repeated a fourth as a field-name `if`-chain; `StateExpectation` already had the clean accessor-table pattern. A shared **`SelectiveCheck`** base now iterates only the fields a test set against a subclass accessor table, and each repetitive group collapses to one table-driven loop:

- `_SCALAR_ACCESSORS` (9 fields), `_LABEL_ROOT_ACCESSORS` (4), `_POOL_TARGET_SLOT_ACCESSORS` (3) on `StoreChecks`
- a module-level slot-accessor table for `AttestationCheck`
- `StateExpectation` and `StoreChecks` both inherit the base

Genuinely bespoke checks (lexicographic head, reorg depth, block-body structure, `filled_block_root_label`) stay explicit — they aren't one-line lookups. `store_checks.py` drops 632 → 584; the duplication is gone without flattening the complex checks. No behavior change (error-message field labels now read as the field name, e.g. `head_slot` rather than `head.slot` — surfaced only on failure, never in vectors).

### 2. Fluent `ForkChoiceScenario` builder

A linear chain with attestations read as ~120 lines of nested `BlockStep(block=BlockSpec(...), checks=StoreChecks(...))`. The builder expresses the same step list through chained calls:

```python
ForkChoiceScenario()
  .block(1, label="block_1").expect(head_slot=Slot(1))
  .tick_to_unix_time(10)
  .attest(by=[1, 2], slot=2, target_label="block_2", target_slot=2)
  .block(3, label="block_3").expect(head_slot=Slot(3), block_attestation_count=1)
  .steps()
```

- `.expect(**fields)` attaches a `StoreChecks` to the preceding step — real field names, no alias layer that could drift.
- Two tick methods (`.tick(interval)` / `.tick_to_unix_time(time)`) name the targets separately, hiding `TickStep`'s mutual-exclusivity foot-gun.
- `.chain(slots)` appends a labeled block per slot.
- The builder emits **exactly** the frozen steps the explicit form does (verified by JSON-equality against the original before rewriting). The explicit `steps=[...]` form stays fully supported; the builder is opt-in.

`test_produce_block_includes_pending_attestations` is rewritten through it as the worked example. The other three block-production tests keep the explicit form, where heavy per-step overrides read more clearly.

### 3. Drop wrapper noise

`validator_indices=[ValidatorIndex(0), ValidatorIndex(1), ...]` → `[0, 1, 2]` across the fork choice tests. The spec types already coerce plain ints to `ValidatorIndex` at construction (verified identical), so the wrappers were pure ceremony. Scoped to validator-index lists in `fc/`; `Slot(...)` is left as domain-meaningful.

## Test plan

- `just check` passes (lint, format, ty, codespell, mdformat, lock)
- Builder/explicit step equality proven via JSON comparison
- Single-test fill smoke on the builder-converted test passes


tcoratger:
Superseded: reduced to the dedup-only change in a new PR. The fluent builder and the wrapper sweep are dropped.

PR #897, opened 2026-06-09
test: add test vector for block production with merging aggregates

unnawut:
## 🗒️ Description

Add a test vector that asserts if a block producer has 2 aggregates of the same message, they get merged successfully for block inclusion.

Since proof bytes are not deterministic, this vector checks that all attesters are accounted for in the produced block.

I was also trying to add split and merge but seems like it needs a leanVM commit bump so I'll defer that to later.

### DIFF
--- a/tests/consensus/lstar/fork_choice/test_signature_aggregation.py
+++ b/tests/consensus/lstar/fork_choice/test_signature_aggregation.py
@@ -8,19 +8,22 @@ from consensus_testing import (
     BlockSpec,
     BlockStep,
     ForkChoiceTestFiller,
+    GossipAggregatedAttestationStep,
     StoreChecks,
+    TickStep,
 )
-from lean_spec.spec.forks import Slot, ValidatorIndex
+from lean_spec.spec.forks import Interval, Slot, ValidatorIndex
 
 pytestmark = pytest.mark.valid_until("Lstar")
 
 
 @pytest.mark.real_crypto(smoke=True)
-def test_multiple_specs_same_target_merge_into_one(
+def test_multiple_attestations_same_target_merge_into_one(
     fork_choice_test: ForkChoiceTestFiller,
 ) -> None:
     """
-    Two attestations sharing one target merge into a single aggregation.
+    Two attestations in the known pool sharing one target get merged
+    into a single aggregation.
 
     Given
     -----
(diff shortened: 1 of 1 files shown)


tcoratger:
Thanks a lot, can you just rebase after https://github.com/leanEthereum/leanSpec/pull/895 and follow the new doc writer conventions for the documentation of the test vectors just for uniformity? :)

Also I haven't checked in details but would be nice to ensure that this test doesn't already exist with the new coverage added in the PR.

unnawut:
Accidentally pushed. I want to do a bit of touch up before marking for review again

unnawut:
> Also I haven't checked in details but would be nice to ensure that this test doesn't already exist with the new coverage added in the PR.

@tcoratger I checked and yep there's `test_multiple_attestations_same_target_merge_into_one` that's very similar, but it merges the leaf signatures, not a recursion. So I think this new one is still relevant because it's forcing 2 proof generation and then recurse them on the next block.

I've moved my new test into the same file as yours though (test_signature_aggregation.py) so they stay close together.

tcoratger (commented):
Thanks for adding this! 🙏 I went through it carefully and the scenario is really nicely constructed. The `TickStep` to slot 1 / interval 3 is a neat touch: it pushes the gossips past the aggregate phase so the two overlapping proofs land in the known pool unmerged, which is exactly what forces the recursive in-block fold — so this genuinely covers a different path from the leaf-merge sibling. 👍

Just a few small things below, mostly around the test-vector docstring conventions, plus one question about crypto coverage. Nothing blocking — feel free to push back on any of them. 🚀

tests/consensus/lstar/fork_choice/test_signature_aggregation.py
```diff
@@ -83,6 +86,91 @@ def test_multiple_specs_same_target_merge_into_one(
     )
 
 
+def test_overlapping_proofs_same_target_recursively_merge_into_one(
```
tcoratger, line 90: One thing worth double-checking: the leaf-merge sibling just above carries `@pytest.mark.real_crypto(smoke=True)`, but this new test doesn't. In the default mocked lane the prover is stubbed out, so the real recursive fold/split only runs in the slower real-crypto lane — which means the fast lane here is really validating the bitfield bookkeeping (overlapping voters collapsing into one attestation) rather than the actual recursive proof.

Since the recursion is kind of the whole point of this vector, would it make sense to add `@pytest.mark.real_crypto(smoke=True)` to mirror the sibling, so the fast lane exercises the real fold too?

(If you intentionally left it off because real recursive fold isn't smoke-lane-stable yet — related to the leanVM bump you mentioned — that's totally fair, I just wanted to flag the asymmetry. 🙂)

tests/consensus/lstar/fork_choice/test_signature_aggregation.py
```diff
@@ -83,6 +86,91 @@ def test_multiple_specs_same_target_merge_into_one(
...
+    When
+    ----
+    - block_2 is built on block_1, carrying no votes of its own.
+
+    Then
+    ----
+    - the builder takes the first proof.
+    - the builder takes the second proof for its one uncovered voter.
+    - the builder folds both proofs into one attestation.
+    - block_2 holds 1 aggregated attestation covering V0, V1, V2, V3.
```
tcoratger: Small docstring-convention nit: these bullets describe the builder's internal selection steps, but the Then section is meant to mirror the assertions one-to-one — and nothing here actually observes selection order or "one uncovered voter" (the checks are just `block_attestation_count=1`, `participants={0,1,2,3}`, and head at slot 2).

Could we swap them for outcome-only bullets, like the sibling has at lines 43-44? Something like:

```
    - block_2 holds 1 aggregated attestation.
    - that aggregation covers V0, V1, V2, V3.
    - head is block_2 at slot 2.
```

That also naturally splits the count and the participant set onto their own lines, which matches the one-fact-per-line style. 🙂

tests/consensus/lstar/fork_choice/test_signature_aggregation.py
```diff
@@ -83,6 +86,91 @@ def test_multiple_specs_same_target_merge_into_one(
     )
 
 
+def test_overlapping_proofs_same_target_recursively_merge_into_one(
+    fork_choice_test: ForkChoiceTestFiller,
+) -> None:
+    """
+    Two overlapping proofs (enforced by 2 separate gossips) for one target
+    fold into a single in-block attestation.
```
tcoratger: Tiny one: the test-vector convention only allows a parenthetical when it glosses a number, so "(enforced by 2 separate gossips)" is slightly off here (and uses "2" as a word). Since the Given section already says both proofs arrive by gossip, I think we can just drop the parenthetical and keep the summary as a clean one-liner. 👍

tests/consensus/lstar/fork_choice/test_signature_aggregation.py
```diff
@@ -83,6 +86,91 @@ def test_multiple_specs_same_target_merge_into_one(
...
+    ----
+    - the builder takes the first proof.
+    - the builder takes the second proof for its one uncovered voter.
+    - the builder folds both proofs into one attestation.
+    - block_2 holds 1 aggregated attestation covering V0, V1, V2, V3.
+    - head is block_2 at slot 2.
+
+    Timing
+    ------
+    - the proofs are gossipped at slot 1, interval 3 (past the aggregate phase).
```
tcoratger: Optional: this Timing line mostly restates the Given bullet about the clock ticking past the aggregate phase. No strong opinion — maybe keep the concrete interval here and drop the Given line (or vice versa), just so the fact lives in one place. Happy either way. 🙂

tests/consensus/lstar/fork_choice/test_signature_aggregation.py
unnawut: I overlooked this. I enabled real crypto now and impressed the test filler is still low 4 minutes so I think yes it makes sense to use real proof for this.

unnawut:
@tcoratger Applied your comments and re-reviewed. Please have another look! I'm still not too familiar with Given/When/Then format so feel free to guide me more

tcoratger (approved):

PR #899, opened 2026-06-09
Add state transition vectors for aggregation bits bounds handling

MegaRedHand:
## Motivation

An audit finding in ethlambda (fixed in [lambdaclass/ethlambda#425](https://github.com/lambdaclass/ethlambda/pull/425)) showed that clients can diverge on blocks carrying malformed `aggregation_bits`: ethlambda skipped an attestation whose bitfield was longer than the validator registry, while zeam and lantern reject such blocks. A skipped attestation also left a spurious all-false entry in the post-state — a cross-client state-divergence vector. The spec had no vectors pinning this behavior.

## Spec change

`process_attestations` previously crashed with a bare `IndexError` when a set bit pointed past the validator registry. That exception is not an `AssertionError`, so rejection handlers cannot catch it, and it carries no language-neutral reason for fixtures. It now raises `SpecRejectionError(VALIDATOR_INDEX_OUT_OF_RANGE)`, matching the equivalent check in `verify_signatures`. The check sits after the vote filters, so filtered attestations are still skipped without their bits being touched. The pre-existing `EMPTY_AGGREGATION_BITS` rejection is unchanged, now resolved once before the tally insert.

## New vectors (`state_transition` format)

| Vector | Expectation |
|---|---|
| Set bit at index >= validator count | Rejected: `VALIDATOR_INDEX_OUT_OF_RANGE` |
| All-false bits (registry-sized) | Rejected: `EMPTY_AGGREGATION_BITS` |
| Zero-length bitfield | Rejected: `EMPTY_AGGREGATION_BITS` |
| Oversized bitfield, all set bits in range | Valid; pinned post-state root fails any client that skips the attestation |

The last vector is the sharp one: validated against ethlambda, it passes on the fixed branch and fails on pre-fix code with a state root mismatch. The rejection vectors discriminate via the emitted `rejectionReason`, so client runners should assert it rather than only checking that the block was rejected.

## Framework changes

- `AggregatedAttestationSpec` gains a raw `aggregation_bits` override, honored only by the unsigned forced-attestation path of the state transition format. Deriving bits from validator indices can only produce tight bitfields, so empty or padded bitfields were previously inexpressible.
- The placeholder zero state root of rejection tests now applies only to the final block, so earlier setup blocks process normally instead of failing with `STATE_ROOT_MISMATCH`. All existing rejection tests are single-block, where this changes nothing (verified: full fill passes unchanged).

## Testing

- `uv run pytest`: 2872 passed
- `uv run fill --fork=lstar -n auto`: 530 fillers pass, determinism check passes
- `just check` clean
- New fixtures run against ethlambda: 4/4 pass on the fix branch; the oversized-bits vector fails on pre-fix code as intended

### DIFF
--- a/packages/testing/src/consensus_testing/test_fixtures/state_transition.py
+++ b/packages/testing/src/consensus_testing/test_fixtures/state_transition.py
@@ -9,7 +9,7 @@ from consensus_testing.keys import XmssKeyManager
 from consensus_testing.test_fixtures.base import BaseConsensusFixture, BaseTestSpec
 from consensus_testing.test_types import AggregatedAttestationSpec, BlockSpec, StateExpectation
 from lean_spec.spec.crypto.merkleization import hash_tree_root
-from lean_spec.spec.forks import AggregationBits, SpecRejectionError
+from lean_spec.spec.forks import SpecRejectionError
 from lean_spec.spec.forks.lstar.containers import (
     AggregatedAttestation,
     AggregatedAttestations,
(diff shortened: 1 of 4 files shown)


tcoratger:
Thanks a lot, can you just rebase after https://github.com/leanEthereum/leanSpec/pull/895 and follow the new doc writer conventions for the documentation of the test vectors just for uniformity? :)

Also I haven't checked in details but would be nice to ensure that this test doesn't already exist with the new coverage added in the PR. I imagine this test coverage was not present before since it helped you guys to spot the bug.

tcoratger (approved):
Thanks for flagging this!

PR #915, opened 2026-06-10
chore: add vulture dead-code detection and remove findings

leolara:
## What

Adds [vulture](https://github.com/jendrikseipp/vulture) for dead-code detection, then removes the dead code it found. Split into two commits so the tooling and the removals can be reviewed independently.

### Commit 1 — add vulture dead-code detection
- Adds `vulture` to the `lint` dependency group and a `just deadcode` recipe.
- Configures vulture in `pyproject.toml` to scan `src`, `packages`, `tests`, and the whitelist together (most spec code is exercised only through the test suite, so the tests must be in scope or every test-only-used symbol reads as a false positive).
- Adds a tight `vulture_whitelist.py` that names each indirectly-used symbol explicitly (framework dispatch, serialization, interpreter protocols) instead of relying on wide glob ignores. Excluded from ruff and ty since it is data, not executable code.
- `just deadcode` is intentionally **not** wired into `just check`: vulture has unavoidable false positives on a Pydantic-heavy codebase, so gating CI on it would be noisy.

### Commit 2 — remove dead code flagged by vulture
Removes the six symbols `just deadcode` reported and that were confirmed unused:
- a write-only last-RPC timestamp on the gossipsub peer state
- a mock response-stream flag that was set but never asserted
- an unused parametrize column in the merkleization tests
- an uncalled mock outbound-stream method
- two uncalled test helpers (a key-file stub writer and an interop finalization waiter)

Three reported symbols were **kept** as confirmed false positives and recorded in the whitelist instead of removed:
- `subscribed` — essential event field, compared via dataclass equality in tests
- `REAL` — constructed via `CryptoMode("real")` from the `--crypto` CLI option
- `MESSAGE_LENGTH` — documented XMSS protocol parameter

## Verification
- `just deadcode` → clean (exit 0)
- `just check` → passes (lint, format, ty, codespell, mdformat, lock)
- affected unit tests pass

## Notes
Draft: the whitelist makes ignores explicit, which means a newly-added Pydantic validator/serializer or fixture will trip vulture until added to `vulture_whitelist.py`. That is the intended trade-off for tight, greppable ignores.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -37,6 +37,11 @@ fix:
 typecheck *args:
     uv run --group lint ty check "$@"
 
+# Detect dead code with vulture (paths and ignores configured in pyproject.toml)
+[group('quality')]
+deadcode *args:
+    uv run --group lint vulture "$@"
+
 # Spell check source, tests, packages, and docs
 [group('quality')]
 spellcheck *args:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,6 +59,9 @@ packages = ["src/lean_spec"]
 
 [tool.ruff]
 line-length = 100
+# The vulture whitelist is data for dead-code detection, not executable code.
+# It references symbols by bare name, which is not valid standalone Python.
+extend-exclude = ["vulture_whitelist.py"]
 
 [tool.ruff.format]
 docstring-code-format = true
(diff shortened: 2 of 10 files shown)


tcoratger (approved):
Great!

Issue #712, opened 2026-05-11
Checkpoint-sync API ships state without anchor block; spec-required `Store.blocks[anchor_root]` cannot be populated


ch4r10t33r:
## Summary

`Store.create_store(state, anchor_block, ...)` (`forks/lstar/spec.py:879`) requires **both** the finalized state and the anchor block, and asserts `anchor_block.state_root == hash_tree_root(state)`. The store is then seeded with `blocks={anchor_root: anchor_block}` (`spec.py:936`).

But the only checkpoint-sync API endpoint defined by the spec — `/lean/v0/states/finalized` (`subspecs/api/routes.py`, served by `endpoints/states.py`) — returns **only** the `State`. There is no spec-defined endpoint that returns the matching `Block` / `SignedBlock`.

A node bootstrapping via checkpoint sync therefore has no spec-conformant way to obtain the anchor block, even though `create_store` mandates it.

## Concrete impact

1. **`create_store` cannot be called.** A checkpoint-syncing node has the state but no anchor block, so it cannot construct the fork-choice store per the spec.

2. **Implementations resort to fabrication.** Real clients work around this by synthesising a placeholder `SignedBlock` from `state.latest_block_header` fields with an empty body and zero signature. Example: zeam PR https://github.com/blockblaz/zeam/pull/857 added `storeSyntheticAnchorBlock` for exactly this reason.

3. **Wire-protocol violation.** Such synthetic blocks have:
   - `hash_tree_root(synthetic) != anchor_root` (because the real anchor block at slot N>0 may contain attestations; synthetic body is empty)
   - `proposer_signature == 0`

   When a peer requests the anchor via `BlocksByRoot` (`/leanconsensus/req/blocks_by_root/1/ssz_snappy`) and verifies the returned root, it MUST reject and score-penalize the responder. Checkpoint-synced nodes become un-syncable to other peers for the anchor block.

4. **Genesis is the only safe case** (slot 0, empty body — synthetic happens to match real). Every checkpoint sync at slot > 0 with attestations in the anchor body breaks.

## Proposed fix

Spec must define an endpoint that ships the anchor block alongside the state. Two options:

**(a) Separate endpoint** — add `/lean/v0/blocks/finalized` returning `SignedBlock` SSZ. Mirrors Beacon API's `/eth/v2/beacon/blocks/{block_id}`. Checkpoint-sync clients fetch `(state, signed_block)` as two requests.

[...]

PR #974, opened 2026-06-12
feat(api): restore /lean/v0/blocks/finalized for checkpoint-sync anchor block

MegaRedHand:
## Summary

Restores the `/lean/v0/blocks/finalized` endpoint that #713 introduced and #751 removed as dead code, makes the node's checkpoint-sync client consume it, and wires the live node to serve it.

The endpoint's consumers were external, which is why the dead-code sweep missed them:

- The **hive lean simulator** gates every checkpoint-sync scenario on it: its helper runner feature-detects the `signed_block_getter` parameter on `ApiServer` ([lean_spec_client_runner.py](https://github.com/ethereum/hive/blob/master/simulators/lean/helper/lean_spec_client_runner.py)) and its readiness probe polls `/lean/v0/blocks/finalized` ([helper.rs](https://github.com/ethereum/hive/blob/master/simulators/lean/src/utils/helper.rs)). Since #751 shipped, the probe 404s forever and **all checkpoint-sync-based reqresp tests fail for every client** on [hive.leanroadmap.org](https://hive.leanroadmap.org): 17 of 21 tests per client in the latest reqresp run. These failures are easy to miss in the UI, since they happen during setup, before the client-under-test starts, so they disappear from per-client filtered views.
- Client implementations serve the endpoint for checkpoint-sync interop, mirroring this spec (e.g. ethlambda).

## Restore the endpoint (server side)

This branch was rebased onto the `node/api` refactor that landed in main (`handlers.py` / `context.py` / `responses.py`, replacing the old `routes.py` + `endpoints/` modules). The endpoint is integrated into that new structure rather than the original file layout:

- `node/api/handlers.py`: a `finalized_block` handler method, alongside the other endpoint handlers.
- `node/api/context.py`: an optional `signed_block_getter` on `ApiContext`, with a `require_signed_block_getter` accessor that raises 503 when unset. The fork-choice store retains only unsigned blocks, so the embedding node injects the signed-block source.
- `node/api/server.py`: the `signed_block_getter` field on `ApiServer`, threaded into the context, and the route registration. The injection interface matches #713 and hive's feature detection.

## Consume it during checkpoint sync (client side)

`Anchor.from_checkpoint` previously rebuilt the anchor block from `state.latest_block_header` with an **empty body**. `create_store` keys the head, the checkpoints, and the block map by `hash_tree_root(anchor_block)`, so whenever the finalized block carried attestations the rebuilt anchor root diverged from the finalized root the rest of the network agrees on. This is the gap #712 originally described.

The new flow, replacing the header reconstruction entirely:

1. Fetch `/lean/v0/blocks/finalized`. The block is fetched first because it is small: a source that cannot serve it (503/404) aborts checkpoint sync before the multi-megabyte state download starts.
2. Fetch `/lean/v0/states/finalized` and run the existing structural and genesis-time checks.
3. Verify the pairing: `block.state_root == hash_tree_root(state)`. A mismatch raises, since it means the source advanced finalization between the two requests and a retry is the fix.
4. `create_store(state, block)`, now keyed by the network's true finalized root.

## Wire the live node to serve it

The server accepted a `signed_block_getter` but the node never supplied one, so on a real node the endpoint always returned 503 (only the test/conformance servers wired a source). `Node.from_genesis` now injects a source that looks up the block by root in the live store and wraps it in an **empty proof**.

The empty proof is intentional, not a placeholder:

- The store and database retain only unsigned blocks; the original full-block proof is deconstructed into per-attestation proofs at import and is not reconstructable.
- The consuming side (`Anchor.from_checkpoint`) uses only `signed_block.block` and never verifies the proof, so an empty one is sufficient for the anchor pair. This mirrors the genesis anchor that no proposer ever signed.
- No current consumer (this spec, the hive simulator, other clients' checkpoint-sync paths) verifies the anchor proof. Serving a real proof would require a short-term signed-block cache near the network layer; that is a best-effort optimization with no functional consumer today, so it is deliberately out of scope.

This is a deliberate departure from #713, which served a mock proposer signature (in the older `BlockSignatures` envelope, since replaced by the merged `MultiMessageAggregate`) and whose comment said real nodes should retain the actual signed block for the finalized root. With no consumer verifying the proof, the empty proof avoids that retention machinery until a verifier exists.

## Tests

- `tests/node/api/endpoints/test_blocks.py`: contract tests (200, content type, SSZ round-trip, block state-root matches the finalized state's hash tree root).
- `tests/node/api/test_server.py`: 503 without store, 503 without signed-block source, 404 when the source has no block, 200 + anchor-root match.
- `tests/node/test_node.py`: the node wires the signed-block source (returns the finalized block in an empty proof, returns `None` for an unknown root).
- `tests/node/sync/test_checkpoint_sync.py`: transport/HTTP/corrupt-SSZ error wrapping for the block fetch, plus a live-server round-trip and the 503 path.
- `tests/node/test_anchor.py`: anchor keyed by the fetched block's root, abort on block-fetch failure, abort on state-fetch failure, raise on state/block pairing mismatch.
- `tests/node/api/endpoints/conftest.py`: the conformance server wires a signed-block source that wraps the store's anchor block with an empty proof.

`just check` and the node/api test trees are green.


src/lean_spec/node/sync/checkpoint_sync.py
```diff
@@ -21,6 +21,13 @@
 FINALIZED_STATE_ENDPOINT: Final = "/lean/v0/states/finalized"
 """Beacon API path for the finalized state."""
 
+FINALIZED_BLOCK_ENDPOINT: Final = "/lean/v0/blocks/finalized"
+"""API endpoint for fetching the signed block matching the finalized state.
+
+Together with the state, this forms the anchor pair for store creation.
+Checkpoint sync requires the source to serve it.
+"""
```
tcoratger:
```suggestion
"""API endpoint for fetching the signed block matching the finalized state."""
```

src/lean_spec/node/sync/checkpoint_sync.py
```diff
@@ -79,6 +86,56 @@ async def fetch_finalized_state(url: str, state_class: type[State]) -> State:
...
 
+async def fetch_finalized_block(url: str) -> SignedBlock:
+    """
+    Fetch the signed block matching the finalized state via checkpoint sync.
+
+    The returned block carries the real body, so its hash tree root equals
+    the finalized root the rest of the network agrees on.
+    The caller must verify the block's state root against the fetched state
+    before pairing them into a store.
+
```
tcoratger:
```suggestion
    Fetch the signed block matching the finalized state via checkpoint sync.

```

src/lean_spec/node/sync/checkpoint_sync.py
```diff
@@ -79,6 +86,56 @@ async def fetch_finalized_state(url: str, state_class: type[State]) -> State:
...
+    Args:
+        url: Base URL of the node API (e.g., "http://localhost:5052").
```
tcoratger:
```suggestion
        url: Base URL of the node API.
```

src/lean_spec/node/api/context.py
```diff
@@ -30,6 +31,14 @@ class ApiContext:
     aggregator_role_control: AggregatorRoleControl | None
     """Holder of the aggregator flag, or None when aggregator control is unwired."""
 
+    signed_block_getter: Callable[[Bytes32], SignedBlock | None] | None
+    """
+    Callable returning the signed block for a block root, or None when unwired.
+
+    The fork-choice store retains only unsigned blocks.
+    Serving the checkpoint-sync anchor block needs a separate signed-block source.
+    """
```
tcoratger:
```suggestion
    """Callable returning the signed block for a block root, or None when unwired."""
```

src/lean_spec/node/api/handlers.py
```diff
@@ -129,6 +129,38 @@ async def finalized_state(self, request: web.Request) -> web.Response:
...
+        Return the finalized signed block as SSZ bytes.
+
+        Together with the finalized state endpoint, this gives a
+        checkpoint-syncing peer the (state, signed block) anchor pair.
+        External consumers bootstrap their fork-choice store from this pair.
+
+        The fork-choice store holds only unsigned blocks.
+        Serving a signed block therefore needs a separate signed-block source.
+        Nodes without such a source answer with 503.
+
```
tcoratger:
```suggestion
        Return the finalized signed block as SSZ bytes.

```

src/lean_spec/node/api/server.py
```diff
@@ -40,6 +41,16 @@ class ApiServer:
...
 
+    signed_block_getter: Callable[[Bytes32], SignedBlock | None] | None = None
+    """
+    Optional callable returning the signed block for a block root.
+
+    The fork-choice store retains only unsigned blocks, so serving the
+    checkpoint-sync anchor block needs a separate signed-block source.
+    The embedding node injects one here.
+    When absent, the finalized block endpoint returns 503.
+    """
```
tcoratger:
```suggestion
    """Optional callable returning the signed block for a block root."""
```

tests/node/api/endpoints/test_blocks.py
```diff
@@ -0,0 +1,54 @@
...
+        """Finalized block endpoint returns octet-stream content type."""
+        response = get_finalized_block(server_url)
+        content_type = response.headers.get("content-type", "")
+        assert "application/octet-stream" in content_type
+
+    def test_ssz_deserializes(self, server_url: str) -> None:
+        """Finalized block SSZ bytes deserialize to a valid SignedBlock object."""
+        response = get_finalized_block(server_url)
+        signed_block = SignedBlock.decode_bytes(response.content)
+        assert signed_block is not None
```
tcoratger: Maybe we can remove this test since in the next one we already assert on `signed_block`, so no need here to assert that it is not None

tests/node/api/endpoints/test_blocks.py
```diff
@@ -0,0 +1,54 @@
...
+        block_response = get_finalized_block(server_url)
+        signed_block = SignedBlock.decode_bytes(block_response.content)
+
+        state_response = httpx.get(
+            f"{server_url}/lean/v0/states/finalized",
+            headers={"Accept": "application/octet-stream"},
+        )
+        state = State.decode_bytes(state_response.content)
+
+        assert signed_block.block.state_root == hash_tree_root(state)
```
tcoratger: Here this is probably better to assert on the full signed_block with `assert signed_block = ....`

tests/node/api/test_server.py
```diff
@@ -88,6 +94,102 @@ async def test_returns_404_when_finalized_state_missing(self, base_store: Store)
             await server.aclose()
 
 
+class TestFinalizedBlockEndpoint:
+    """Tests for the /lean/v0/blocks/finalized endpoint."""
+
+    @staticmethod
+    def _signed_block_getter_for(store: Store) -> Callable[[Bytes32], SignedBlock | None]:
```
tcoratger: Small thought: this little "look up a block, wrap it in an empty proof" helper shows up in a few places now (here, the endpoints `conftest`, `test_checkpoint_sync`, `test_anchor`'s `_signed_genesis_block`, and the getter in `node.py`). Maybe we could move one version into `consensus_testing` and reuse it everywhere? Would trim the setup a bit. Totally optional though!

tests/node/api/test_server.py
```diff
@@ -88,6 +94,102 @@ async def test_returns_404_when_finalized_state_missing(self, base_store: Store)
...
+                response = await client.get("http://127.0.0.1:15071/lean/v0/blocks/finalized")
+
+                assert response.status_code == 503
+
+        finally:
+            await server.aclose()
+
+    async def test_returns_503_without_signed_block_source(self, base_store: Store) -> None:
+        """Endpoint returns 503 when no signed-block source is configured."""
+        config = ApiServerConfig(port=15075)
```
tcoratger: Tiny heads-up: this port (15075) is also used in `tests/node/sync/test_checkpoint_sync.py`, and 15071 is reused within this file. Under xdist they could occasionally race for the same bind. If it's easy, `port=0` (we already do that in `test_node.py`) lets the OS pick a free port and sidesteps the collision. No worries if it feels out of scope here.

tests/node/sync/test_checkpoint_sync.py
```diff
@@ -217,3 +294,46 @@ async def test_client_fetches_and_deserializes_state(self, base_store: Store) ->
...
+
+    async def test_block_fetch_raises_when_source_unconfigured(self, base_store: Store) -> None:
+        """A server without a signed-block source yields the 503 error path."""
+        config = ApiServerConfig(port=15076)
+        server = ApiServer(config=config, store_getter=lambda: base_store)
+
+        await server.start()
+
+        try:
+            with pytest.raises(CheckpointSyncError, match="HTTP error 503"):
```
tcoratger: Minor: we usually assert the full message with `==` rather than a `match=` fragment, so the rest of the string can't drift unnoticed. Same idea for the `startswith(...)` one a few lines up (line 273), if its message is stable. Something like:

```python
with pytest.raises(CheckpointSyncError) as exception_info:
    await fetch_finalized_block("http://127.0.0.1:15076")
assert str(exception_info.value) == "HTTP error 503: ..."
```

tests/node/test_anchor.py
```diff
@@ -130,3 +187,36 @@ async def test_success_builds_store_and_status(self) -> None:
...
+                "lean_spec.node.anchor.fetch_finalized_state",
+                new_callable=AsyncMock,
+                return_value=checkpoint_state,
+            ),
+            patch(
+                "lean_spec.node.anchor.fetch_finalized_block",
+                new_callable=AsyncMock,
+                return_value=mismatched_block,
+            ),
+            pytest.raises(CheckpointSyncError, match="mismatch"),
```
tcoratger: Same small thing as in the checkpoint-sync tests — could we assert the full message here too? e.g. `== "anchor block / state mismatch; source advanced finalization between requests, retry"`. Keeps the exact wording pinned. 🙂

tests/node/test_anchor.py
```diff
@@ -130,3 +187,36 @@ async def test_success_builds_store_and_status(self) -> None:
         assert anchor.store is not None
         assert anchor.validators == checkpoint_state.validators
         assert anchor.initial_status.finalized == anchor.store.latest_finalized
+        # The anchor is keyed by the fetched block's root, so the store's
+        # finalized checkpoint matches the root the network agrees on.
+        assert anchor.store.latest_finalized.root == hash_tree_root(signed_block.block)
```
tcoratger, line 192: Nice assertion! One small idea: since the heart of this change is that the store is keyed by the *fetched* block, maybe also assert the block actually landed in the store, e.g. `assert hash_tree_root(signed_block.block) in anchor.store.blocks`? Just makes the intent extra explicit.

src/lean_spec/node/anchor.py
```diff
@@ -96,6 +96,10 @@ async def from_checkpoint(
             CheckpointSyncError: For every failure mode covering transport,
```
tcoratger, line 96: Small doc nit (feel free to fold into your other docstring tweaks): the method now also raises on the block/state pairing mismatch, so it might be worth adding that to the `Raises:` list here — it's a retryable case, so handy for callers to know about.

tcoratger (commented):
Thanks for restoring this — and the anchor-root fix (fetching the real signed block instead of rebuilding it from the header with an empty body) is a really nice catch 🙌

I left a few small, totally optional suggestions inline, mostly around the tests. Nothing blocking — the source side looks good to me. The main themes:
- a couple of exception assertions use `match=` / `startswith`, which we usually swap for a full `==` on the message
- the "wrap a block in an empty proof" helper is repeated in a few places, so maybe a small shared helper could simplify the setup
- a couple of hardcoded ports overlap, which can get flaky under parallel runs

Feel free to take or leave any of these! 🙂

MegaRedHand:
@tcoratger all comments were addressed 🫡

tcoratger (approved):

PR #1001, opened 2026-06-13
fix(forkchoice): derive finalized from canonical head, not an independent max

### LINKED ISSUE #1000: Fork choice: store.latest_finalized can latch a losing fork's higher finalized, freezing finalization
## Summary

`on_block` advances `store.latest_finalized` by an **independent monotonic max** over every imported block's post-state, decoupled from `store.latest_justified` and the head:

```python
# src/lean_spec/spec/forks/lstar/fork_choice.py  (on_block)
latest_finalized = store.latest_finalized.advance_to(post_state.latest_finalized)
```

A fork that finalizes a **higher** slot but then **loses head selection** leaves its finalized checkpoint latched in the store. `store.latest_finalized` can then sit on a dead branch that is **not an ancestor of the head**, while `store.latest_justified` and `head` are on a different branch whose state finalized is **lower**.

## Impact — finalization freeze (liveness)

`get_attestation_target` decides target justifiability against `store.latest_finalized` (`is_justifiable_after(store.latest_finalized.slot, ...)`). But the canonical state transition (`process_attestations`) validates each attestation's target against the **state's** `latest_finalized` (the head chain's, which is lower). When the two finalized **slots** disagree the justifiable-slot lattice is offset, so **every advancing target the validators vote is rejected** (`target not justifiable`) and justification/finalization **freezes permanently** — even with full (2/3+) attestation coverage and a live, advancing head.

This was first observed as a network-wide finalization freeze in a long-running multi-subnet devnet (a client mirroring this store update): head and justification kept advancing while finalized stayed pinned. Root cause: `store.latest_finalized` (higher, from a block that had been reorged out) ≠ the canonical head state's `latest_finalized` (lower).

## Root cause / when it was introduced

The original forkchoice store (#53) and the 3sf-mini reference derive the finalized checkpoint from the **chosen (head) state**, so justified and finalized stay consistent:
- 3sf-mini `p2p.py`: a staker reads `latest_finalized_hash` from `post_states[head]` (the head's state); there is **no** max-over-states for finalized (only `get_latest_justified_hash` maxes, and only for *justified*, the LMD anchor).
- leanSpec #53 `update_head`: `latest_finalized = self.states[new_head].latest_finalized` (the head's state).

**#194** (`a1ae0916`) moved the finalized update into `on_block` as an **independent max over post-states** — which is what allows the desync. #727 renamed that max to `advance_to`; #817 relocated it.

## Reproduction

Regression test `test_losing_fork_higher_finalized_does_not_latch` (see PR): a dead fork finalizes slot 4, then a heavier fork justifies slot 7 (winning the head) but finalizes only slot 1. On the pre-fix code `store.latest_finalized` latches `dead_4` (slot 4, **not** an ancestor of the head); it should track the head's state (slot 1 on `block_1`).

## Fix

Derive `store.latest_finalized` from the canonical head's chain in **`update_head`**, so it is recomputed on every head change (block import, acceptance ticks, proposals): take the finalized **slot** named by the head's state, and recover its **root** as the head's own ancestor at that slot. This keeps the checkpoint on the head chain and makes `store.latest_finalized.slot == head-state.latest_finalized.slot`, so the slot a validator attests against matches the slot the state transition validates against. It restores #53 / 3sf-mini semantics (finalized read from the head state, not a max).

`produce_block_with_signatures` carried a **second** copy of the independent max (it advanced `store.latest_finalized` from the produced block's own state and pruned against it, stranding/over-pruning on a reorged proposal); that write is removed — `update_head` owns the finalized checkpoint.

Note: the fork choice is **not** clamped to finalized descendants. leanSpec finalization is order-dependent, so a fork can finalize a higher slot yet lose head selection; clamping LMD-GHOST to the finalized subtree would latch the head onto that losing fork. PR #1001.

---

## Observed devnet trace (the original network-wide freeze)

The freeze was first seen in a multi-node beam-sim run, where a finalized checkpoint was observed to **revert** (finalized reached slot 27, then fell back to slot 18 — a protocol-violating finalized reorg). The complete trace that pinned it as an implementation bug:

```
[s=38 i=3] block 38 (proposer=6, parent=slot 36) imported (on_block)
           -> can_target_finalize(27) = true
           -> block 38 post-state finalized = 27
           -> fed into fcStore (monotonic max) -> store.finalized = 27
[s=38 i=3] fork choice reorg: old_head=37 -> new_head=38      (block 38 becomes head)
[s=38 i=4] fork choice reorg: old_head=38 -> new_head=37      (block 38 reorged out, dropped)
[s=39 i=1] head = {slot: 37}; block 39 (proposer=7) built on slot 37
[s=39 i=1] prestate justified = {slot: 30}  finalized = {slot: 18}   (canonical back to finalized 18)
```

### Fork tree

```
   ... - block 36 (common ancestor)
            |
            +-- branch A:  block 38 (proposer 6, parent 36)        31/32 votes finalize 27, then reorged out
            |
            +-- branch B:  block 37 - block 39 - ...               canonical, justified climbs to 108, finalized 18
```

Both forks branch off block 36. Branch A finalizes a higher slot (27) but loses head selection; branch B wins the head (its justified climbs to 108) yet finalizes only slot 18.

### What happened, step by step

1. **block 38 (branch A) legitimately finalizes 27** with 31/32 votes — a real network supermajority.
2. **block 38 immediately loses head selection** (head 38 -> 37 at the next interval). The canonical chain follows branch B (36 -> 37 -> 39 -> ...), which never finalizes 27 and stays at 18.
3. **`store.latest_finalized`, the independent monotonic max, latches the orphaned block 38's `finalized = 27`** and never reverts after the reorg.
4. **`store.finalized = 27` (dead fork) != canonical `state.finalized = 18` (branch B).** Validators derive attestation targets against 27; the canonical state transition validates them against 18; `is_justifiable_after` disagrees, so every advancing target is rejected and finalization freezes.

### Why branch B never finalizes 27

On branch B, target slot 30 was justified first by votes from a **different** source. The later `source = 27` votes then hit `target_already_justified` and are skipped, so branch B does not finalize 27. This is the order-dependence of 3sf finalization: the same votes finalize different things depending on processing order, so a losing fork can finalize a higher slot than the canonical chain.

This is the concrete, real-network instance of the minimal `test_losing_fork_higher_finalized_does_not_latch` regression above: branch A (block 38, finalize 27) is the dead fork; branch B (the heavier chain that wins the head and finalizes only 18) is the heavy fork.

GrapeBaBa:
## Problem

Fixes #1000.

`store.latest_finalized` was advanced by an **independent monotonic max** over every imported block's post-state, decoupled from the head. A fork that finalizes a **higher** slot but then **loses head selection** left its finalized checkpoint latched in the store, so `store.latest_finalized` could sit on a branch that is **not** the head's while `head` / `latest_justified` were on another.

`get_attestation_target` derives the attestation target against `store.latest_finalized`, while `process_attestations` validates targets against the canonical head state's `latest_finalized` — **both using only the slot** (`is_justifiable_after`). Once the two finalized **slots** diverged, every advancing target was rejected and **finalization froze** — first observed as a network-wide freeze in a long-running multi-subnet devnet (store/banner finalized 27, canonical state 18).

`#194` introduced the independent max; the original store (#53) and the 3sf-mini reference derive finalized from the chosen head's state (full trace in #1000).

## Fix

Derive `store.latest_finalized` from the canonical head's chain, in **`update_head`**, so it is recomputed on **every** head change — block import, acceptance ticks, and proposals:

```python
def _finalized_on_head_chain(self, store):
    finalized_slot = store.states[store.head].latest_finalized.slot
    finalized_root = store.head
    while store.blocks[finalized_root].slot > finalized_slot:
        finalized_root = store.blocks[finalized_root].parent_root
    return Checkpoint(root=finalized_root, slot=finalized_slot)
```

- `store.latest_finalized.slot == head_state.latest_finalized.slot` ⇒ the slot a validator attests against matches the slot the state transition validates against — the freeze is gone.
- The **root** is the head's ancestor at that slot, so it names a real block even for a checkpoint-sync anchor (whose state stores a placeholder it cannot resolve to its own block).

**This restores 3sf-mini.** In `3sf-mini/p2p.py` the staker reads finalized from the head state, not a max:
```python
@property
def latest_finalized_hash(self):
    return self.post_states[self.head].latest_finalized_hash
# in vote():  (state = self.post_states[self.head])
while not is_justifiable_slot(state.latest_finalized_slot, target_block.slot):
```
`#194`'s max-over-blocks was the deviation. (3sf-mini *does* keep a max — `get_latest_justified_hash` — but only for **justified**, the LMD anchor, never for finalized. leanSpec's `store.latest_justified` matches that, and the attestation source keeps using it per #595.)

### DIFF
--- a/src/lean_spec/spec/forks/lstar/fork_choice.py
+++ b/src/lean_spec/spec/forks/lstar/fork_choice.py
@@ -805,19 +805,23 @@ class ForkChoiceMixin(LstarSpecBase):
             attestations=latest_votes,
         )
         # Invariant: the finalized checkpoint stays on the head's chain.
-        # Its root is the head's own ancestor at the finalized slot.
-        # A checkpoint-sync anchor state holds a placeholder for its own block.
-        # The ancestor walk recovers the real block.
+        # Climb from the head to its ancestor at the finalized slot.
         finalized_slot = store.states[new_head].latest_finalized.slot
         finalized_root = new_head
-        while store.blocks[finalized_root].slot > finalized_slot:
-            finalized_root = store.blocks[finalized_root].parent_root
-        return store.model_copy(
-            update={
-                "head": new_head,
-                "latest_finalized": Checkpoint(root=finalized_root, slot=finalized_slot),
-            }
-        )
+        while finalized_root in store.blocks and store.blocks[finalized_root].slot > finalized_slot:
+            parent_root = store.blocks[finalized_root].parent_root
+            if parent_root not in store.blocks:
+                break
+            finalized_root = parent_root
+
+        # A fresh checkpoint-sync anchor stores no block at that slot.
+        # Keep the trusted anchor there rather than emit an unresolved root.
+        if store.blocks[finalized_root].slot == finalized_slot:
+            latest_finalized = Checkpoint(root=finalized_root, slot=finalized_slot)
+        else:
+            latest_finalized = store.latest_finalized
+
+        return store.model_copy(update={"head": new_head, "latest_finalized": latest_finalized})
 
     def accept_new_attestations(self, store: LstarStore) -> LstarStore:
         """
--- a/src/lean_spec/spec/forks/lstar/fork_choice.py
+++ b/src/lean_spec/spec/forks/lstar/fork_choice.py
@@ -633,11 +633,6 @@ class ForkChoiceMixin(LstarSpecBase):
             )
 
             # Recompute the head now that the block and its votes are in the store.
-            # update_head derives the finalized checkpoint from the new head's chain,
-            # never an independent max over every imported block, so latest_finalized
-            # always names a block on the canonical chain and the slot a validator
-            # attests against matches the slot the state transition validates against
-            # -- finalization cannot freeze.
             store = self.update_head(store)
 
             # Prune stale vote data, but only when finalization advanced past the snapshot.
--- a/src/lean_spec/spec/forks/lstar/fork_choice.py
+++ b/src/lean_spec/spec/forks/lstar/fork_choice.py
@@ -84,28 +84,6 @@ class ForkChoiceMixin(LstarSpecBase):
         # The chain left the known tree before reaching the ancestor's slot.
         return False
 
-    def _finalized_on_head_chain(self, store: LstarStore) -> Checkpoint:
-        """
-        Return the finalized checkpoint that lies on the canonical head's chain.
-
-        The head state names the finalized slot. Its root is taken as the head's
-        own ancestor at that slot, so the checkpoint always lies on the head chain
-        and names a block the store holds -- even when the state carries a
-        placeholder root it cannot resolve to its own block, as a checkpoint-sync
-        anchor state does.
-
-        Args:
-            store: Fork-choice store holding the head and the block tree.
-
-        Returns:
-            The finalized checkpoint on the head's chain.
-        """
-        finalized_slot = store.states[store.head].latest_finalized.slot
-        finalized_root = store.head
-        while store.blocks[finalized_root].slot > finalized_slot:
-            finalized_root = store.blocks[finalized_root].parent_root
-        return Checkpoint(root=finalized_root, slot=finalized_slot)
-
     def create_store(
         self,
         state: SpecStateType,
@@ -826,12 +804,20 @@ class ForkChoiceMixin(LstarSpecBase):
             start_root=store.latest_justified.root,
             attestations=latest_votes,
         )
-        store = store.model_copy(update={"head": new_head})
-
-        # The finalized checkpoint travels with the head. Deriving it from the new
-        # head's chain on every recompute keeps it on the canonical chain, so it
-        # never lags a head that moved nor pins a fork that lost head selection.
-        return store.model_copy(update={"latest_finalized": self._finalized_on_head_chain(store)})
+        # Invariant: the finalized checkpoint stays on the head's chain.
+        # Its root is the head's own ancestor at the finalized slot.
+        # A checkpoint-sync anchor state holds a placeholder for its own block.
+        # The ancestor walk recovers the real block.
+        finalized_slot = store.states[new_head].latest_finalized.slot
+        finalized_root = new_head
+        while store.blocks[finalized_root].slot > finalized_slot:
+            finalized_root = store.blocks[finalized_root].parent_root
+        return store.model_copy(
+            update={
+                "head": new_head,
+                "latest_finalized": Checkpoint(root=finalized_root, slot=finalized_slot),
+            }
+        )
 
     def accept_new_attestations(self, store: LstarStore) -> LstarStore:
         """
--- a/src/lean_spec/spec/forks/lstar/fork_choice.py
+++ b/src/lean_spec/spec/forks/lstar/fork_choice.py
@@ -607,9 +607,6 @@ class ForkChoiceMixin(LstarSpecBase):
             #
             #     store slot 5, candidate slot 7  ->  take candidate
             #     store slot 5, candidate slot 5  ->  keep store
-            #
-            # The finalized checkpoint is NOT taken as an independent max here; it is
-            # derived from the canonical head's state after head selection (below).
             latest_justified = store.latest_justified.advance_to(post_state.latest_justified)
 
             # Seed each block-carried vote into the known pool with an empty proof set.
--- a/src/lean_spec/spec/forks/lstar/fork_choice.py
+++ b/src/lean_spec/spec/forks/lstar/fork_choice.py
@@ -84,6 +84,28 @@ class ForkChoiceMixin(LstarSpecBase):
         # The chain left the known tree before reaching the ancestor's slot.
         return False
 
+    def _finalized_on_head_chain(self, store: LstarStore) -> Checkpoint:
+        """
+        Return the finalized checkpoint that lies on the canonical head's chain.
+
+        The head state names the finalized slot. Its root is taken as the head's
+        own ancestor at that slot, so the checkpoint always lies on the head chain
+        and names a block the store holds -- even when the state carries a
+        placeholder root it cannot resolve to its own block, as a checkpoint-sync
+        anchor state does.
+
+        Args:
+            store: Fork-choice store holding the head and the block tree.
+
+        Returns:
+            The finalized checkpoint on the head's chain.
+        """
+        finalized_slot = store.states[store.head].latest_finalized.slot
+        finalized_root = store.head
+        while store.blocks[finalized_root].slot > finalized_slot:
+            finalized_root = store.blocks[finalized_root].parent_root
+        return Checkpoint(root=finalized_root, slot=finalized_slot)
+
     def create_store(
         self,
         state: SpecStateType,
@@ -578,15 +600,17 @@ class ForkChoiceMixin(LstarSpecBase):
             # Run the state transition from the parent state to this block's post-state.
             post_state = self.state_transition(parent_state, block)
 
-            # Advance the justified and finalized checkpoints from the post-state.
+            # Advance the justified checkpoint from the post-state.
             #
             # A candidate wins only when its slot is strictly higher than the store's.
             # On a slot tie the store keeps its checkpoint, avoiding a silent root swap:
             #
             #     store slot 5, candidate slot 7  ->  take candidate
             #     store slot 5, candidate slot 5  ->  keep store
+            #
+            # The finalized checkpoint is NOT taken as an independent max here; it is
+            # derived from the canonical head's state after head selection (below).
             latest_justified = store.latest_justified.advance_to(post_state.latest_justified)
-            latest_finalized = store.latest_finalized.advance_to(post_state.latest_finalized)
 
             # Seed each block-carried vote into the known pool with an empty proof set.
             #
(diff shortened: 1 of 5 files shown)


src/lean_spec/spec/forks/lstar/fork_choice.py
```diff
@@ -84,6 +84,28 @@ def _checkpoint_is_ancestor(
...
+            store: Fork-choice store holding the head and the block tree.
+
+        Returns:
+            The finalized checkpoint on the head's chain.
+        """
+        finalized_slot = store.states[store.head].latest_finalized.slot
+        finalized_root = store.head
+        while store.blocks[finalized_root].slot > finalized_slot:
+            finalized_root = store.blocks[finalized_root].parent_root
+        return Checkpoint(root=finalized_root, slot=finalized_slot)
```
tcoratger: Please avoid having a function for this since this is used only one time, just inline the logic at the appropriate location :)

src/lean_spec/spec/forks/lstar/fork_choice.py
```diff
@@ -578,15 +600,17 @@ def on_block(
...
+            # Advance the justified checkpoint from the post-state.
             #
             # A candidate wins only when its slot is strictly higher than the store's.
             # On a slot tie the store keeps its checkpoint, avoiding a silent root swap:
             #
             #     store slot 5, candidate slot 7  ->  take candidate
             #     store slot 5, candidate slot 5  ->  keep store
+            #
+            # The finalized checkpoint is NOT taken as an independent max here; it is
+            # derived from the canonical head's state after head selection (below).
```
tcoratger: The line just after this doesn't treat finalized thing so this looks confusing

```suggestion
```

src/lean_spec/spec/forks/lstar/fork_choice.py
```diff
@@ -607,12 +631,16 @@ def on_block(
...
                     "latest_known_aggregated_payloads": new_known_aggregated_payloads,
                 }
             )
 
             # Recompute the head now that the block and its votes are in the store.
+            # update_head derives the finalized checkpoint from the new head's chain,
+            # never an independent max over every imported block, so latest_finalized
+            # always names a block on the canonical chain and the slot a validator
+            # attests against matches the slot the state transition validates against
+            # -- finalization cannot freeze.
```
tcoratger: Too verbose

```suggestion
```

src/lean_spec/spec/forks/lstar/fork_choice.py
```diff
@@ -809,7 +834,12 @@ def update_head(self, store: LstarStore) -> LstarStore:
             start_root=store.latest_justified.root,
             attestations=latest_votes,
         )
-        return store.model_copy(update={"head": new_head})
+        store = store.model_copy(update={"head": new_head})
+
+        # The finalized checkpoint travels with the head. Deriving it from the new
+        # head's chain on every recompute keeps it on the canonical chain, so it
+        # never lags a head that moved nor pins a fork that lost head selection.
+        return store.model_copy(update={"latest_finalized": self._finalized_on_head_chain(store)})
```
tcoratger: You can do it inside the same model_copy and the documentation is too verbose, so difficult to read, check the rules of doc writer subagent

src/lean_spec/spec/forks/lstar/validator_duties.py
```diff
@@ -199,29 +199,21 @@ def produce_block_with_signatures(
...
-        # Checkpoint advances must be propagated manually here.
-        #
-        # Tie semantics mirror the block-import path.
-        # A candidate needs a strictly higher slot to replace the store's view.
+        # Locally produced blocks bypass normal block processing, so the justified
+        # checkpoint is advanced manually. The finalized checkpoint is not written
+        # here: update_head derives it from the canonical head, and get_proposal_head
+        # ran update_head just above, so the store's finalized already sits on the head
+        # this block extends. Pinning it here from the produced block's own state would
+        # strand it -- and prune against it -- if the proposal later lost head selection.
```
tcoratger: Too verbose, check the rules of doc writer

src/lean_spec/spec/forks/lstar/fork_choice.py
```diff
@@ -806,7 +804,20 @@ def update_head(self, store: LstarStore) -> LstarStore:
             start_root=store.latest_justified.root,
             attestations=latest_votes,
         )
-        return store.model_copy(update={"head": new_head})
+        # Invariant: the finalized checkpoint stays on the head's chain.
+        # Its root is the head's own ancestor at the finalized slot.
+        # A checkpoint-sync anchor state holds a placeholder for its own block.
+        # The ancestor walk recovers the real block.
```
tcoratger: Minor: this says the walk "recovers the real block" for a checkpoint-sync anchor, but at a fresh anchor those ancestors aren't in the store yet — which is exactly the crash flagged below. Once the walk is guarded, worth rewording to say it stops at the deepest known block (and that at a fresh anchor the head itself is the finalized root).

The last line ("The ancestor walk recovers the real block.") just narrates the loop right below it, so it can go.

src/lean_spec/spec/forks/lstar/fork_choice.py
```diff
@@ -806,7 +804,20 @@ def update_head(self, store: LstarStore) -> LstarStore:
...
+        finalized_slot = store.states[new_head].latest_finalized.slot
+        finalized_root = new_head
+        while store.blocks[finalized_root].slot > finalized_slot:
+            finalized_root = store.blocks[finalized_root].parent_root
```
tcoratger: **Blocker: this walk crashes on a checkpoint-sync boot.**

`finalized_slot` comes from the *state's* finalized checkpoint. On a fresh checkpoint-sync anchor that slot sits below the anchor block's slot — the served finalized state's own `latest_finalized` points at an earlier source, while `from_checkpoint` sets the anchor block slot from `latest_block_header.slot`. At that point `store.blocks` holds only the anchor, so the loop steps to `anchor_block.parent_root`, which isn't in the store, and `store.blocks[finalized_root]` raises `KeyError`.

This is reachable: `update_head` runs on the first attestation duty before any post-anchor block is imported. So a checkpoint-synced node crashes on boot. (Genesis boot is fine — there finalized slot == block slot, so the loop never runs.)

Note the two other ancestor walks in this file already guard the lookup with `while current_root in store.blocks` — this is the only unguarded one.

Suggested guard (also handles the skipped-slot case):

```python
finalized_root = new_head
while (
    finalized_root in store.blocks
    and store.blocks[finalized_root].slot > finalized_slot
):
    parent_root = store.blocks[finalized_root].parent_root
    if parent_root not in store.blocks:
        break
    finalized_root = parent_root
```

If the walk can't reach a block at exactly `finalized_slot`, keep the existing `store.latest_finalized` (the anchor) rather than emitting an unresolved root.

**Related (skipped slot):** the condition is `> finalized_slot`, so if no block sits exactly at `finalized_slot` this lands on an ancestor *below* it and emits `Checkpoint(root=<lower block>, slot=finalized_slot)` — a root whose real slot doesn't match its `.slot` field, which attestation validation later relies on. Worth an `assert store.blocks[finalized_root].slot == finalized_slot`, or preferring the state's stored finalized root when it's already in the store.

Repro: boot from a finalized state with `latest_block_header.slot = 8` and `latest_finalized.slot = 1`, then call `update_head` on the anchor-only store.

src/lean_spec/spec/forks/lstar/validator_duties.py
```diff
@@ -199,29 +199,17 @@ def produce_block_with_signatures(
...
-        # Persist block and state.
-        previous_finalized_slot = store.latest_finalized.slot
         store = store.model_copy(
             update={
                 "blocks": store.blocks | {block_hash: final_block},
                 "states": store.states | {block_hash: final_post_state},
                 "latest_justified": latest_justified,
-                "latest_finalized": latest_finalized,
             }
         )
```
tcoratger, line 213: Dropping the finalized-max and the prune here is correct, but nothing tests this path. The consensus filler builds blocks through the import path, never through block production, and the node pytests that do call this don't assert the finalized checkpoint or pruning. So the symmetric regression — a self-produced block on a losing fork stranding a higher finalized checkpoint, or stale vote data surviving — would go uncaught.

Worth a node test in tests/node/validator/test_service.py that builds a losing-fork store, produces a block, and asserts finalized matches the head-derived value (not a latched max) and the attestation pools are untouched when finalization doesn't advance.

tcoratger (commented):
## Review — finalized derived from canonical head

Thanks for tracking this down. The root-cause writeup is spot on: the independent max latched a losing fork's finalized checkpoint, and because both target derivation and target validation compare by slot only, the diverged slots froze finalization. Deriving finalized from the head's chain is the right fix and restores 3sf-mini parity, and the new vector isolates the freeze well (it fails pre-fix with finalized latched at slot 4, passes on the fix).

One blocker before this can merge, plus two follow-ups — details inline.

**Blocker — checkpoint-sync boot crashes in the new walk.** The ancestor walk reads the *state's* finalized slot but assumes every parent is already in the store. On a checkpoint-synced node the store holds only the anchor block, whose state's finalized slot sits *below* the anchor block slot, so the first `update_head` (called on the first attestation duty, before any block import) steps to a parent root that isn't there and raises `KeyError`. The two existing ancestor walks in this file guard with `while current_root in store.blocks` — this new one doesn't. Vectors miss it because the test genesis rebases the anchor's finalized up to the anchor slot, so the loop never iterates.

**Follow-ups:** the same walk can emit a checkpoint whose root isn't actually at the finalized slot when that slot was skipped; and the block-production change has no test exercising it.

Unrelated: the failing interop CI job is a QUIC peer-connection issue, not caused by this change — but worth a green run before merge.

tcoratger (approved):

PR #1017, opened 2026-06-14
test: vector for proposer scheduling on empty validator registry

tcoratger:
Adds a state_transition negative vector driving round-robin proposer
selection against a zero-validator registry, pinning the empty-registry
rejection that previously had no fixture coverage. The filler's proposer
fallback now delegates to the spec schedule, so the typed rejection
surfaces instead of a bare division-by-zero in the harness.

### DIFF
--- a/packages/testing/src/consensus_testing/test_types/block_spec.py
+++ b/packages/testing/src/consensus_testing/test_types/block_spec.py
@@ -130,8 +130,6 @@ class BlockSpec(CamelModel):
 
     def resolve_proposer_index(self, num_validators: int) -> ValidatorIndex:
         """Return the proposer index, falling back to the spec's round-robin schedule."""
-        # Delegate to the spec so an empty registry surfaces the spec's typed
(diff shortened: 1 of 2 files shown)


packages/testing/src/consensus_testing/test_types/block_spec.py
```diff
@@ -129,8 +129,12 @@ class BlockSpec(CamelModel):
     """
 
     def resolve_proposer_index(self, num_validators: int) -> ValidatorIndex:
-        """Return the proposer index, falling back to round-robin by slot."""
-        return self.proposer_index or ValidatorIndex(int(self.slot) % num_validators)
+        """Return the proposer index, falling back to the spec's round-robin schedule."""
+        # Delegate to the spec so an empty registry surfaces the spec's typed
+        # rejection instead of a bare division-by-zero in the test framework.
```
tcoratger:
```suggestion
```

PR #1114, opened 2026-06-17
test: add test vectors for split and reaggregate attestations

unnawut:
## 🗒️ Description

Adds a new `reaggregation_test` consensus fixture format and 3 vectors covering the split + merge of
aggregate proofs.

A client that imports a block holding a partial attestation must be able to split that attestation's proof
back out of the block's merged multi-message proof, then merge it with its own local partial for the
same attestation.

The test vectors here only touch the split/merge crypto primitives via container methods. It does
not drive syncing i.e. when to split, pool maintenance, re-gossip triggers that SyncService does.

### Three cases

- `test_split_only_when_data_unseen_locally` — no local partial; recovered proof used as-is
- `test_split_then_merge_with_overlapping_local_partial` — overlapping local partial
- `test_split_then_merge_with_disjoint_local_partial` — disjoint local partial

### Example generated vector

`test_split_then_merge_with_overlapping_local_partial` — block proof carries `[V0,V1,V2]`, the importing
node holds a local partial `[V1,V2,V3]`, overlapping on `[V1,V2]`. The combined proof covers their union
`[V0,V1,V2,V3]`:

```json
{
    "network": "Lstar",
    "leanEnv": "test",
    "proofSetting": 1,
    "blockProof": "0x67750500ff16024a…(358378 bytes)",
    "publicKeysPerMessage": [
        ["0x80a6f13b…(52 bytes)", "0x7612d36f…(52 bytes)", "0xd703e67c…(52 bytes)"],
        ["0xdedd8e5a…(52 bytes)"]
    ],
    "attestationMessage": "0x4a07ff61053777fe…(32 bytes)",
    "attestationSlot": 1,
    "blockAttesters": [0, 1, 2],
    "localAttesters": [1, 2, 3],
    "combinedAttesters": [0, 1, 2, 3],
    "reaggregatedProof": "0x60ce0500ffffffff…(381280 bytes)",
    "valid": true
}
```

`publicKeysPerMessage` is the per-component key layout the split needs: the attestation's three keys
first, the proposal key last.

### How a client consumes it

1. Read `blockProof` and `publicKeysPerMessage`.
2. Split `blockProof` by `attestationMessage` and supplying `blockAttesters` (the split needs the signer list as
input — it doesn't return it), to recover the attestation's single-message proof.
3. Verify the recovered proof against the block attesters' keys at `attestationSlot`.
4. Merge the recovered proof with the node's local partial → the re-aggregated proof.
5. Verify the re-aggregated proof against `combinedAttesters`' keys, expecting valid.

 ## Testing

- `uv run fill --fork=lstar --crypto=real -k test_post_block_reaggregation` → 3 passed; determinism gate
passes (these vectors are correctly excluded from it).
- `just check` + `just deadcode` clean.

### DIFF
--- a/packages/testing/src/consensus_testing/__init__.py
+++ b/packages/testing/src/consensus_testing/__init__.py
@@ -43,6 +43,8 @@ from consensus_testing.test_fixtures import (
     NetworkingCodecTest,
     PoseidonPermutationFixture,
     PoseidonPermutationTest,
+    ReaggregationFixture,
+    ReaggregationTest,
     RebindToAlternateHeadRoot,
     SetProposerIndex,
     SlotClockFixture,
@@ -144,6 +146,7 @@ SlotClockTestFiller = Callable[..., SlotClockFixture]
 JustifiabilityTestFiller = Callable[..., JustifiabilityFixture]
 PoseidonPermutationTestFiller = Callable[..., PoseidonPermutationFixture]
 SyncTestFiller = Callable[..., SyncFixture]
+ReaggregationTestFiller = Callable[..., ReaggregationFixture]
 
 __all__ = [
     "CachedMessage",
@@ -183,6 +186,8 @@ __all__ = [
     "FromSlot",
     "FromUnixTime",
     "TotalIntervals",
+    "ReaggregationFixture",
+    "ReaggregationTest",
     "VerifyCheckpoint",
     # Public API
     "AggregatedAttestationSpec",
@@ -278,4 +283,5 @@ __all__ = [
     "JustifiabilityTestFiller",
     "PoseidonPermutationTestFiller",
     "SyncTestFiller",
+    "ReaggregationTestFiller",
 ]
--- a/packages/testing/src/consensus_testing/test_fixtures/__init__.py
+++ b/packages/testing/src/consensus_testing/test_fixtures/__init__.py
@@ -58,6 +58,7 @@ from consensus_testing.test_fixtures.poseidon_permutation import (
     PoseidonPermutationFixture,
     PoseidonPermutationTest,
 )
+from consensus_testing.test_fixtures.reaggregation import ReaggregationFixture, ReaggregationTest
 from consensus_testing.test_fixtures.slot_clock import (
     CurrentInterval,
     CurrentSlot,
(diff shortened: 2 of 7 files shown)


tcoratger (commented):
I'm not sure we should merge this. In fact we would like to try to limit the number of tests as much as possible with real crypto because those tests, via aggregation pipeline, considerably increase the CI timing while all the other mock tests are pretty much free.

Also I think this is important to mention that the aggregation pipeline and all the underlying logic or merging, etc should already be tested properly in the Rust crate and from the spec point of view, we should just trust this layer blindly.

And I think that in the test fixture of this PR, we exercise pretty much only the Rust methods with the Python wrappers over `lean_multisig_py`.

For me the only genuinely leanSpec-Python parts are the participant bitfield union, the key-layout assembly, and the when-to-split gates — and those are either trivial or already covered by `test_service.py` under mocked crypto.

So with `pytest.mark.real_crypto`, these 5 vectors are predominantly proving the Rust prover/verifier is correct, which may inflate the CI timing not much in my opinion.

unnawut:
Totally agree with your concerns on the CI timing and having redundant tests that are leanVM's responsibility, and these tests are mostly only touching leanVM API over the Python wrapper so they're not really adding much coverage to to the spec!

My thinking is it’s useful to have these sanity checks for interop debugging (i.e. seeing this exact test failure and not as a fork choice or sync failure). For example, right now `test_service.py` are all unit tests I think, so they don’t carry over to clients. And because clients see only the initial inputs and final outputs on the test vectors, any issues caught by `test_service.py` would just show as generic test vector mismatch.

But again I agree we don’t need this many. Just one would be enough e.g. `test_split_then_merge_with_overlapping_local_partial` but wdyt?

tcoratger:
> Totally agree with your concerns on the CI timing and having redundant tests that are leanVM's responsibility, and these tests are mostly only touching leanVM API over the Python wrapper so they're not really adding much coverage to to the spec!
> 
> My thinking is it’s useful to have these sanity checks for interop debugging (i.e. seeing this exact test failure and not as a fork choice or sync failure). For example, right now `test_service.py` are all unit tests I think, so they don’t carry over to clients. And because clients see only the initial inputs and final outputs on the test vectors, any issues caught by `test_service.py` would just show as generic test vector mismatch.
> 
> But again I agree we don’t need this many. Just one would be enough e.g. `test_split_then_merge_with_overlapping_local_partial` but wdyt?

Ok let's go for one test only I agree, can you remove the other ones and maybe the useless stuffs if any after deleting all the other tests? :)

tcoratger (approved):

unnawut:
Sorry I saw your comment too late. Thanks for update and merge!

PR #1149, opened 2026-06-17
feat(lstar): add tiered block-production strategy as a selectable alternative

MegaRedHand:
## What

Adds a second proposer-side block-building algorithm for the lstar fork, kept alongside the existing one as an interchangeable mixin.

| Mixin | File | Selection policy |
|---|---|---|
| `BlockProductionMixin` (default) | `block_production.py` | round-based fixed-point: re-scan until a pass adds nothing |
| `TieredBlockProductionMixin` (new) | `block_production_tiered.py` | tiered greedy scorer: finalize > justify > build, projected incrementally |

Both share the `build_block` contract declared on `LstarSpecBase`, so a fork composes exactly one.

## How to switch

The default stays the round-based selection. To select the tiered scorer, swap a single import in `spec.py`:

```python
# from lean_spec.spec.forks.lstar.block_production import BlockProductionMixin
from lean_spec.spec.forks.lstar.block_production_tiered import (
    TieredBlockProductionMixin as BlockProductionMixin,
)
```

then regenerate:

```
uv run fill --fork=lstar --clean -n auto
```

## Test changes: fixtures generate under either mixin

Two fork-choice fillers used to assert the simple builder's exact block body. The two builders include a different number of votes in those scenarios, so generation aborted when the tiered builder was wired in. Both fillers now generate cleanly under either mixin **without any test depending on which mixin is active** (verified: 540 vectors pass and the determinism check passes, under both).

Each filler is made mixin-independent in the way that fits its scenario:

- **`test_produce_block_enforces_max_attestations_data_limit`** — switched to a single-voter, justifiable-target scenario. No vote reaches the two-thirds supermajority, so no entry justifies its target and the entry cap alone bounds the body. Both builders therefore fill to exactly the limit, and the scenario asserts that count for both. (This is the more robust setup originally written for #791; it changes this one committed vector.)

- **`test_justified_divergence_self_heals_in_next_block`** — the divergence here is real and same-scenario: the fixed-point builder keeps a vote already recorded on-chain, the tiered builder drops it as adding no new voters. That body difference is a proposer-strategy choice, not a consensus outcome, so it is no longer asserted. The test instead asserts the justified slot, which is the strategy-independent witness of the self-heal: the chain can only reach justified slot 1 once the block incorporates the slot-1 votes.

The committed vectors are generated by the default (simple) strategy.

## Notes

- `just check` passes (lint, format, ty, codespell, mdformat).
- `just deadcode` passes.
- Supersedes the replace-approach in #791.

### DIFF
--- a/src/lean_spec/spec/forks/lstar/block_production_tiered.py
+++ b/src/lean_spec/spec/forks/lstar/block_production_tiered.py
@@ -1,0 +1,425 @@
+"""
+Lstar fork — proposer-side block building, tiered greedy strategy.
+
+This is an alternative to the round-based fixed-point strategy.
+Exactly one block-production mixin is composed into the fork at a time.
+The two are interchangeable: same contract, different selection policy.
+"""
+
+from collections import defaultdict
+from collections.abc import Set as AbstractSet
+from dataclasses import dataclass
+from enum import IntEnum
+
+from lean_spec.spec.crypto.merkleization import hash_tree_root
+from lean_spec.spec.crypto.xmss.containers import PublicKey
+from lean_spec.spec.forks.lstar._base import LstarSpecBase
+from lean_spec.spec.forks.lstar.aggregation import select_proofs_for_coverage
+from lean_spec.spec.forks.lstar.config import (
+    MAX_ATTESTATIONS_DATA,
+)
+from lean_spec.spec.forks.lstar.containers import (
+    AggregatedAttestation,
+    AttestationData,
+    Block,
+    JustifiedSlots,
+    SingleMessageAggregate,
+    Slot,
+    State,
+    ValidatorIndex,
+)
+from lean_spec.spec.ssz import ZERO_HASH, Boolean, Bytes32
+
+
+class _Tier(IntEnum):
+    """
+    Selection tier for a candidate attestation data entry.
+
+    Declared in priority order: a lower value wins.
+    """
+
+    FINALIZE = 1
+    """Applying the entry crosses two-thirds on target and finalizes the source."""
+    JUSTIFY = 2
+    """Applying the entry crosses two-thirds on target but does not finalize."""
+    BUILD = 3
+    """Adds marginal new voters toward target's two-thirds supermajority."""
+
+
+@dataclass(frozen=True)
+class _EntryScore:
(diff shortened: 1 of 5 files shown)


MegaRedHand:
Note: the dead-code CI check will be red on this PR until #1150 merges.

#1150 fixes a pre-existing stale `vulture_whitelist.py` on `main` (the `hash_tree_root` handlers were renamed `_htr_*` -> `_hash_tree_root_*` in #1124 without updating the whitelist). That fix has been kept out of this PR to keep it focused. Once #1150 lands on `main`, merging `main` here clears the dead-code failure; the only whitelist change this PR then carries is the `TieredBlockProductionMixin` entry.

src/lean_spec/spec/forks/lstar/block_production_tiered.py
```diff
@@ -0,0 +1,425 @@
...
+    new_voter_count: int
+    target_slot: Slot
+    attestation_slot: Slot
+
+    def ordering_key(self, data_root: Bytes32) -> tuple[int, int, int, int, bytes]:
+        """Sort key where the smallest tuple is the best candidate."""
+        return (
+            int(self.tier),
+            -self.new_voter_count,
+            int(self.target_slot),
```
anshalshukla: -int(self.target_slot)

We should give higher target_slot more priority as it will help us move forward justified slot farther in case of finalization/justification stalls.

This will be helpful as it can quickly move latest_justified closer to the tip, this was implemented in Zeam in devent4 as one of the fix

src/lean_spec/spec/forks/lstar/block_production_tiered.py
```diff
@@ -0,0 +1,425 @@
...
+            int(self.attestation_slot),
```
anshalshukla: -int(self.attestation_slot)

latest votes should be given more priority

src/lean_spec/spec/forks/lstar/block_production_tiered.py
MegaRedHand: Made the change. Also, for attestations that reach supermajority, the target slot is taken into account before the number of voters

PR #1166, opened 2026-06-18
feat(validator): source votes from the head chain's justified checkpoint

MegaRedHand:
## What

Change the attestation vote source in `produce_attestation_data` from the store's latest justified (`store.latest_justified`) to the latest justified on the **current head's state** (`store.states[store.head].latest_justified`).

## Why

A validator's vote source should name the justified checkpoint on the chain its vote extends, not the store's global view.

The store can advance its justified checkpoint from a minority fork that the head never extended (see `test_attestation_source_divergence.py`). When that happens the store's justified and the head chain's justified diverge, and sourcing from the store would place the vote's source off what the head chain actually justified.

```
            genesis
              |
            common(1)
            /        \
      block_2(2)     fork_4(4)   <- justifies common (slot 1): 3/4 votes
          |
      block_3(3)                 <- head; its state justifies nothing (slot 0)

store.latest_justified        = common (slot 1)   <- advanced by the fork_4 minority branch
head_state.latest_justified   = genesis (slot 0)  <- what the head chain actually justified
```

After this change a vote produced on `block_3` sources from slot 0 (the head chain), not slot 1 (the store).

## The liveness failure this prevents

`store.latest_justified` is always an ancestor of the head, but a minority fork can push it to a **higher slot** than the head chain's own state has justified. If the current canonical chain advances finalization while not justifying the `latest_justified` or beyond, the STF rules make it skip the repackaged attestations from the fork. While that happens, validators keep voting with the fork's `latest_justified` as source, making useless votes, and the justification divergence is never closed. Step by step:

```
genesis(0) - c1(1) - c2(2) - ... - c8(8) - c9(9) - c10(10) - c11(12)  canonical head chain (head = c11)
                                            |
                                           k10(11)          minority fork off c9
```

1. **The chain advances.** Blocks c1..c9 are built without justifying or finalizing anything.
2. **The chain forks.** Block k10 is proposed, with votes that justify slot 9 (c9), but keep the majority of head weight on the canonical chain.
3. **The canonical chain finalizes, but its justification stays behind the fork.** Earlier, correctly-sourced votes let the canonical chain finalize slot 1 (c1) and justify slot 2 (c2). Its justified slot (2) never climbs ahead of the slot the fork justified (9).
4. **Attestations from the fork can't be repackaged.** Since finalization advanced in the canonical chain, attestations from the fork are invalid and skipped in the STF. Slot 9 is unjustifiable when viewed from the canonical chain: [a delta of 8 is unjustifiable](https://github.com/lambdaclass/ethlambda/blob/main/docs/3sf_mini.md#the-justifiability-schedule).
5. **The chain is stuck.** Under the old rule every new vote sources from `store.latest_justified` = slot 9. To justify any checkpoint, the canonical chain needs votes whose source is slot 2, so those slot-9-sourced votes are skipped in the STF. Because of this, the chain stops including votes, which causes block production to stop due to the justification divergence being unclosable

Sourcing from the head state's justified breaks the stall: every vote anchors on a slot the canonical chain itself has justified (slot 2 here), so new votes to close the divergence are eventually included.

## Genesis handling

The raw genesis state holds a **zero-root placeholder** for its justified checkpoint; the state transition rebases that to the real genesis root on the first block. A vote cast directly on the genesis head normalizes the placeholder the same way, so the source always names a real block (otherwise `validate_attestation` would reject it with `UNKNOWN_SOURCE_BLOCK`).

## Testing

- All 84 affected node unit tests pass.
- All 542 consensus test vectors pass and are byte-identical across hash seeds.
- `just check` passes (lint, format, typecheck, spell, mdformat).

### DIFF
--- a/src/lean_spec/spec/forks/lstar/validator_duties.py
+++ b/src/lean_spec/spec/forks/lstar/validator_duties.py
@@ -15,7 +15,7 @@ from lean_spec.spec.forks.lstar.containers import (
     ValidatorIndex,
 )
 from lean_spec.spec.forks.lstar.errors import RejectionReason, SpecRejectionError
-from lean_spec.spec.ssz import Uint64
+from lean_spec.spec.ssz import Bytes32, Uint64
 
 
 class ValidatorDutiesMixin(LstarSpecBase):
@@ -83,11 +83,20 @@ class ValidatorDutiesMixin(LstarSpecBase):
 
         target_checkpoint = self.get_attestation_target(store)
 
+        # Source the vote from the head chain's own justified checkpoint.
+        # The store can advance its justified from a minority fork the head never extended.
+        # Voting with the head's justified keeps the source on the chain the vote extends.
+        justified_source = store.states[store.head].latest_justified
+
+        # Replace the placeholder genesis root with the real one.
+        if justified_source.root == Bytes32.zero():
+            justified_source = Checkpoint(root=store.head, slot=justified_source.slot)
+
         return self.attestation_data_class(
             slot=slot,
             head=head_checkpoint,
             target=target_checkpoint,
-            source=store.latest_justified,
+            source=justified_source,
         )
 
     def produce_block_with_signatures(
--- a/src/lean_spec/spec/forks/lstar/validator_duties.py
+++ b/src/lean_spec/spec/forks/lstar/validator_duties.py
@@ -92,7 +92,7 @@ class ValidatorDutiesMixin(LstarSpecBase):
         if justified_source.root == Bytes32.zero():
             justified_source = Checkpoint(root=store.head, slot=justified_source.slot)
 
-        # Sanity check: the source must be after target.
+        # Sanity check: the source must be before the target.
         assert justified_source.slot <= target_checkpoint.slot
 
         return self.attestation_data_class(
--- a/src/lean_spec/spec/forks/lstar/validator_duties.py
+++ b/src/lean_spec/spec/forks/lstar/validator_duties.py
@@ -92,6 +92,9 @@ class ValidatorDutiesMixin(LstarSpecBase):
         if justified_source.root == Bytes32.zero():
             justified_source = Checkpoint(root=store.head, slot=justified_source.slot)
 
+        # Sanity check: the source must be after target.
+        assert justified_source.slot <= target_checkpoint.slot
+
         return self.attestation_data_class(
             slot=slot,
             head=head_checkpoint,
--- a/src/lean_spec/spec/forks/lstar/validator_duties.py
+++ b/src/lean_spec/spec/forks/lstar/validator_duties.py
@@ -92,7 +92,7 @@ class ValidatorDutiesMixin(LstarSpecBase):
         if justified_source.root == Bytes32.zero():
             justified_source = Checkpoint(root=store.head, slot=justified_source.slot)
 
-        # Sanity check: the source must be before the target.
+        # Sanity check: the source must be older or equal to the target.
         assert justified_source.slot <= target_checkpoint.slot
 
         return self.attestation_data_class(
--- a/src/lean_spec/spec/forks/lstar/validator_duties.py
+++ b/src/lean_spec/spec/forks/lstar/validator_duties.py
@@ -75,7 +75,7 @@ class ValidatorDutiesMixin(LstarSpecBase):
         return Checkpoint(root=target_block_root, slot=target_block.slot)
 
     def produce_attestation_data(self, store: LstarStore, slot: Slot) -> AttestationData:
-        """Attestation data for the given slot: head, target, and the latest justified source."""
+        """Attestation data for the given slot: head, target, and the head's justified source."""
         head_checkpoint = Checkpoint(
             root=store.head,
             slot=store.blocks[store.head].slot,
--- a/tests/node/validator/test_service.py
+++ b/tests/node/validator/test_service.py
@@ -937,7 +937,7 @@ class TestValidatorServiceIntegration:
         The attestation must reference:
         - head: the current chain head
         - target: the attestation target based on forkchoice
-        - source: the latest justified checkpoint
+        - source: the head chain's justified checkpoint
         """
         attestations_produced: list[SignedAttestation] = []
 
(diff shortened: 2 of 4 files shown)


src/lean_spec/spec/forks/lstar/validator_duties.py
```diff
@@ -83,11 +83,20 @@ def produce_attestation_data(self, store: LstarStore, slot: Slot) -> Attestation
 
         target_checkpoint = self.get_attestation_target(store)
 
+        # Source the vote from the head chain's own justified checkpoint.
+        # The store can advance its justified from a minority fork the head never extended.
+        # Voting with the head's justified keeps the source on the chain the vote extends.
+        justified_source = store.states[store.head].latest_justified
```
tcoratger, line 89: This is the heart of the change and it reads really well 👍

Two small thoughts while we're here:

- The summary docstring a few lines up still says "the latest justified source" — might be worth nudging it to something like "the head chain's justified source" so it matches the new behavior.

- Tiny edge-case I couldn't fully convince myself about: could the head state's justified ever sit at a *higher* slot than `target_checkpoint`? If that's ever possible, the vote would get dropped as source-after-target, which is the exact thing we're trying to avoid. I couldn't actually build a case where it happens, so it may well be fine — but a one-liner like `assert justified_source.slot <= target_checkpoint.slot` would make the invariant explicit and fail loudly if it ever did. Totally optional, just flagging it.

tests/node/validator/test_service.py
```diff
@@ -955,7 +955,9 @@ async def capture_attestation(attestation: SignedAttestation) -> None:
 
         store = real_sync_service.store
         expected_head_root = store.head
-        expected_source = store.latest_justified
+        # The head is the genesis anchor, whose state carries the zero-root justified placeholder.
+        # The produced source normalizes that placeholder to the real genesis root.
+        expected_source = Checkpoint(root=store.head, slot=Slot(0))
```
tcoratger, line 960: Small heads-up on this one: since the head here is the genesis anchor, `store.latest_justified` and `store.states[store.head].latest_justified` both come out as (genesis, slot 0). So I think this assertion passes with *both* the old and the new code — meaning it doesn't really catch the divergence the PR is about. Still a fine smoke test, just not the regression guard.

For actually locking the behavior in, a consensus vector feels like the right spot (forks are tested by vectors). The neat part: `tests/consensus/lstar/fork_choice/test_attestation_source_divergence.py` already builds the perfect setup — store justifies `common` at slot 1 while the head chain (`block_3`) justifies nothing. But it injects explicit attestations, so it skips `produce_attestation_data`.

An honest `GossipAttestationSpec` attestation on `block_3` there would route through `produce_attestation_data`, and we could assert its source is genesis/slot 0 (new behavior) and not `common`/slot 1 (old behavior) — that single check distinguishes the two.

One catch: `AttestationCheck` can't read back a single attestation's source today (it only looks at the aggregated pools, and only exposes `source_slot`), so it'd need a small extension — read `store.attestation_signatures` and add a `source_root` check. Happy to help wire that up if it's useful!

tcoratger (commented):
Really nice fix, and the writeup in the description is super clear — the step-by-step liveness explanation made this easy to follow 🙏

One thing that I noticed: since this changes behavior inside a fork module, the convention here is to pin it with a consensus vector under `tests/consensus/`, not just a node unit test. Right now I don't think anything actually exercises the divergence this PR fixes — so it'd be easy for it to quietly regress later. Left a note inline with a concrete idea for a vector that would lock it in.

Nothing blocking — just a couple of small inline thoughts (a docstring line, a tiny edge-case, and the coverage idea). Thanks for this! 🙂

src/lean_spec/spec/forks/lstar/validator_duties.py
MegaRedHand: Added the docstring change.

MegaRedHand: About the source-after-target case: I believe we've seen some cases of this during devnet runs. This happens when a block advances `latest_justified` between `safe_target` updates. ethlambda, zeam, and maybe other clients, already have a sanity check like that:

https://github.com/lambdaclass/ethlambda/blob/2f619ce5164b4d35b633a14791de48976381517a/crates/blockchain/src/store.rs#L683-L690

I added the sanity check you suggested.

tests/node/validator/test_service.py
MegaRedHand: Added in [3af7632](https://github.com/leanEthereum/leanSpec/pull/1166/commits/3af76327d03b969ae067611d2427e935c3c28545). I had to add an additional source slot check for attestations. Let me know if that's not what you had in mind.

src/lean_spec/spec/forks/lstar/validator_duties.py
```diff
@@ -75,19 +75,31 @@ def get_attestation_target(self, store: LstarStore) -> Checkpoint:
...
+
+        # Replace the placeholder genesis root with the real one.
+        if justified_source.root == Bytes32.zero():
+            justified_source = Checkpoint(root=store.head, slot=justified_source.slot)
+
+        # Sanity check: the source must be after target.
```
tcoratger: Source before the target no?

tcoratger (approved):

src/lean_spec/spec/forks/lstar/validator_duties.py
MegaRedHand: Yes. Good catch

src/lean_spec/spec/forks/lstar/validator_duties.py
```diff
@@ -75,19 +75,31 @@ def get_attestation_target(self, store: LstarStore) -> Checkpoint:
...
+        # The store can advance its justified from a minority fork the head never extended.
+        # Voting with the head's justified keeps the source on the chain the vote extends.
+        justified_source = store.states[store.head].latest_justified
+
+        # Replace the placeholder genesis root with the real one.
+        if justified_source.root == Bytes32.zero():
+            justified_source = Checkpoint(root=store.head, slot=justified_source.slot)
+
+        # Sanity check: the source must be before the target.
+        assert justified_source.slot <= target_checkpoint.slot
```
ensi321, line 96: Do we allow source and target have the same slot? If we do, then comment should say "before or equal". If we don't then `<=` should be replaced with `<`.

MegaRedHand: The STF skips those votes, I think. However, they still have a head vote. This is useful right after genesis, where the only vote possible will have `source = target = 0`.

I'll update the comment

anshalshukla (approved):

g11tech (approved):
Upon deep discussions into the scenario, we zeam agree that even though head is to be selected in the subtree of store's latest justified (as it is now),

the voting as well as the block packing (where we assert that produced block latest justified has to be >= store justified ) needs to be simpler and easy

So this PR makes sense and also @anshalshukla can we do another PR where we remove the asset that we have in the block production for its latest justified>= store's latest justified ?

I think store's latest justified fulfills it's use at the moment of head selection where we find head and safe target in the subtree of store's latest justified

PR #1168, opened 2026-06-25
simplify block building

anshalshukla:
## 🗒️ Description
- Removes assertion in block building as #1166  with voting only happening on the state justified it should not lead to a situation where store justified has moved forward - https://github.com/leanEthereum/leanSpec/pull/1166#pullrequestreview-4562249808
- ~Prefers attestations that have higher target slot, we should prioritize latest votes over stale votes~ nvm it is a block optimization and will need vote counting as well which will complicate the spec
- Block building can be simplified to include attestations only with source as current justified, reverts #716   

## ✅ Checklist

- [ ] Ran local quality checks to avoid unnecessary CI fails:
    ```console
    just check
    ```

### DIFF
--- a/src/lean_spec/spec/forks/lstar/block_production.py
+++ b/src/lean_spec/spec/forks/lstar/block_production.py
@@ -148,6 +148,10 @@ class BlockProductionMixin(LstarSpecBase):
                     if attestation_data.head.root not in known_block_roots:
                         continue
 
+                    # Skip votes whose source slot is not the current justified checkpoint.
+                    if attestation_data.source.slot != current_justified_checkpoint.slot:
+                        continue
+
                     # Reject votes that do not match this chain.
                     #
                     # This also rejects any checkpoint past the chain view.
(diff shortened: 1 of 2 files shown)


tcoratger (approved):

src/lean_spec/spec/forks/lstar/block_production.py
```diff
@@ -148,6 +148,10 @@ def build_block(
                     if attestation_data.head.root not in known_block_roots:
                         continue
 
+                    # Skip votes whose source slot is not the current justified checkpoint.
+                    if attestation_data.source.slot != current_justified_checkpoint.slot:
```
MegaRedHand, line 152: I thought adding this would allow us to simplify the checks here. Otherwise, this just adds more code, right?

anshalshukla: correct, it's just easy to reason through it this way. I intentionally did not add any test to enforce it as it is optional and up to client implementation

PR #1186, opened 2026-07-06
BlockSpec check_state_transition

### LINKED ISSUE #1183: test_process_slots_target_equal_to_state_slot_rejected
Fixture json should include `BlockSpec(slot=Slot(1))`,
but there is no blocks in json `"blocks": []`.

There is no blocks,
because `_build_block_from_spec` calls `process_slots`,
which triggers assert test is expected to test.
But when this assert triggers,
it doesn't return new block.
```python
class StateTransitionFixture(BaseConsensusFixture):
    def generate(...):
        try:
            for block_index, block_spec in enumerate(self.blocks):
                block, cached_state = self._build_block_from_spec(...)
                filled_blocks.append(block)
    def _build_block_from_spec(...):
        if not block_spec.skip_slot_processing:
            slot_advanced_state = fork.process_slots(state, block_spec.slot)
```

Using `skip_slot_processing=True` causes `Expected rejection BLOCK_SLOT_NOT_IN_FUTURE but processing succeeded`.
```diff
 def test_process_slots_target_equal_to_state_slot_rejected(...)
-  BlockSpec(slot=Slot(1)),
+  BlockSpec(slot=Slot(1), skip_slot_processing=True),
```

turuslan:
## 🗒️ Description
Add `check_state_transition` to `BlockSpec`.
Fixes `test_process_slots_target_equal_to_state_slot_rejected` test.

Can be backported to devnet4.

## 🔗 Related Issues or PRs
Closes #1183

## ✅ Checklist
- [ ] Ran local quality checks to avoid unnecessary CI fails:
    ```console
    just check
    ```


leolara:
This fixes it :-)

Not specific about this PR, but in general, my opinion is that the interface should be different, but I can send that in a different Pr after this is merged.

The test writer writes state_transition(pre_state, block), so the interface is hiding that things are being done to the pre_state before calling state_transition by default. I think the default would be to use the pre_state as is, and if we would like to process_slots before calling state_transition, that would be the extra option.

I think that would follow better the principle of Least Surprise.

tcoratger (approved):

PR #1188, opened 2026-07-07
fix(fork-choice): correct equivocation vectors' expected heads to the larger-root fork

### LINKED ISSUE #1187: fill: #1181 equivocation fixtures fail to generate at HEAD (fork choice disagrees with expected head)
## `fill` fails to generate the #1181 equivocation fixtures at HEAD — fork choice disagrees with the vectors' expected head

At current `main` (`4ca7d27e46b9b978c4f0a9b7e8eb6fd06be8286d`), generating the `lstar` fixture set fails: the three equivocation tests added in #1181 assert a head that the fork-choice implementation does not produce. `fill` therefore exits non-zero, which blocks generating (and consuming) the full `lstar` fixture set at HEAD. We hit this integrating gean against HEAD.

### Reproduction

```
uv run fill --clean --fork=lstar --scheme=prod --output=fixtures
```

```
E   AssertionError: Step 6: head_slot = 2, expected 3
E   AssertionError: Step 6: head_slot = 3, expected 2
E   AssertionError: Step 6: head_slot = 3, expected 2
=========================== short test summary info ============================
FAILED tests/consensus/lstar/fork_choice/test_equivocation.py::test_same_slot_equivocating_attesters_count_once[fork_Lstar]
FAILED tests/consensus/lstar/fork_choice/test_equivocation.py::test_equivocation_head_independent_of_arrival_order_a_then_b[fork_Lstar]
FAILED tests/consensus/lstar/fork_choice/test_equivocation.py::test_equivocation_head_independent_of_arrival_order_b_then_a[fork_Lstar]
========== 3 failed, 551 passed, 35 deselected ==========
```

All other 551 tests generate cleanly; only these three fail.

### What the failures suggest

The two arrival-order tests are the interesting signal:

- `..._a_then_b` → `head_slot = 3` (expected `2`)
- `..._b_then_a` → `head_slot = 3` (expected `2`)

Both deliveries produce the **same** head (`3`), so the **order-independence** property #1181 introduced is holding — the head is a pure function of store contents. But the deterministic result disagrees with the vectors' expected value (`3` vs `2`), and `test_same_slot_equivocating_attesters_count_once` disagrees in the other direction (`2` vs `3`).

That reads like an **implementation-vs-vector mismatch introduced with #1181**: the equal-slot tie-break resolves to a different branch than the test vectors encode. Either the tie-break direction in the fork-choice code and the expected head in these vectors were computed against different conventions, or one of the two is off.

For reference, `test_same_slot_equivocating_attesters_count_once`'s own docstring states the intent as *"the two votes tie on slot, so the larger attestation-data root wins … head is fork_b"* — so the *intent* (larger canonical root wins) is clear; the disagreement is about which concrete head that yields in these fixtures.

### Impact

- `fill --fork=lstar` cannot produce the full fixture set at HEAD, so downstream clients cannot regenerate/consume the `lstar` vectors until this is resolved (we worked around it by excluding `test_equivocation.py` to validate the rest).
- These three vectors were added in #1181 (closes #1172), so this appears to be a regression contained to that change rather than pre-existing.

### Question

Is this a known issue? Is the intended fix to adjust the tie-break resolution in `fork_choice.py`, or to regenerate/correct the expected heads in these three vectors?

Happy to provide the full `fill` log or a minimal repro if useful.

mananuf:
## Fix the #1181 equivocation vectors that fail to generate at HEAD

Closes #1187.

At current `main`, `fill --fork=lstar` fails to generate three fixtures — the equivocation tests added in #1181 assert a head the fork-choice implementation does not produce:

```
FAILED test_equivocation.py::test_same_slot_equivocating_attesters_count_once      (head_slot 2, expected 3)
FAILED test_equivocation.py::test_equivocation_head_independent_of_arrival_order_a_then_b  (head_slot 3, expected 2)
FAILED test_equivocation.py::test_equivocation_head_independent_of_arrival_order_b_then_a  (head_slot 3, expected 2)
```

### Root cause: the vectors, not the code

The three vectors hard-code **which** fork carries the larger attestation-data root and derive the expected head from that guess. But the concrete `hash_tree_root` ordering is scenario-dependent: `test_same_slot` uses 8 validators while the arrival-order tests use 6, which changes the genesis state root, hence the fork block roots, hence the vote-data roots. The docstrings guessed the ordering and guessed wrong.

The fork-choice code is correct: `_extract_attestations_from_aggregated_payloads` resolves an equal-slot tie toward the larger attestation-data root deterministically, and the property #1181 targets holds — `a_then_b` and `b_then_a` both produce the same head (`fork_b`, slot 3), independent of arrival order. The disagreements even go in **opposite directions** across tests (`same_slot`: actual `fork_a` vs asserted `fork_b`; arrival-order: actual `fork_b` vs asserted `fork_a`), which a direction bug in the code could not produce — it's the code computing the true per-scenario larger root while each vector guessed the wrong fork.

### The fix

Correct the three vectors' expected heads (and docstrings) to the actual deterministic output — no spec-code change:

- `test_same_slot_equivocating_attesters_count_once`: head is **fork_a (slot 2)**; V0/V1 count once toward fork_a (weights 3 vs 2).
- `test_..._a_then_b` / `test_..._b_then_a`: head is **fork_b (slot 3)**, identical across both arrival orders — which is exactly the order-independence these tests exist to prove.

### Verification

`uv run fill tests/consensus/lstar/fork_choice/test_equivocation.py --fork=lstar --scheme=prod` now generates all fixtures cleanly; the generated heads match (`2`, `3`, `3`). `ruff format`/`ruff check` pass.

### Note

These tests remain sensitive to block-root ordering (inherent to hard-coding a hash-determined winner) — a future change to block building could flip the concrete winner again. Happy to follow up with a more robust formulation (e.g. asserting the equivocator-counted-once and order-independence invariants without pinning the absolute fork) if preferred.

### DIFF
--- a/tests/consensus/lstar/fork_choice/test_equivocation.py
+++ b/tests/consensus/lstar/fork_choice/test_equivocation.py
@@ -259,7 +259,7 @@ def test_same_slot_equivocating_attesters_count_once(
     - one vote at slot 3 targets fork_b from V0, V1, V3, V4.
     - V0 and V1 equivocate by voting on both forks at the same slot.
     - the two votes tie on slot, so the larger attestation-data root wins.
-    - the fork_b vote carries the larger root.
+    - the fork_a vote carries the larger root.
 
     When
     ----
@@ -267,10 +267,10 @@ def test_same_slot_equivocating_attesters_count_once(
 
     Then
     ----
-    - V0 and V1 count once, toward fork_b.
-    - fork_b has effective weight 4 from V0, V1, V3, V4.
-    - fork_a has effective weight 1 from V2.
-    - head is fork_b.
+    - V0 and V1 count once, toward fork_a.
+    - fork_a has effective weight 3 from V0, V1, V2.
+    - fork_b has effective weight 2 from V3, V4.
+    - head is fork_a.
     - no slot is justified by the below-threshold votes.
     """
     fork_choice_test(
@@ -317,8 +317,8 @@ def test_same_slot_equivocating_attesters_count_once(
             TickStep(
                 time=16,
                 checks=StoreChecks(
-                    head_slot=Slot(3),
-                    head_root_label="fork_b",
+                    head_slot=Slot(2),
+                    head_root_label="fork_a",
                     latest_justified_slot=Slot(0),
                     latest_finalized_slot=Slot(0),
                     latest_known_aggregated_target_slots=[Slot(2), Slot(3)],
@@ -327,13 +327,13 @@ def test_same_slot_equivocating_attesters_count_once(
                             validator=ValidatorIndex(0),
                             location="known",
                             attestation_slot=Slot(3),
-                            target_slot=Slot(3),
+                            target_slot=Slot(2),
                         ),
                         AttestationCheck(
                             validator=ValidatorIndex(1),
                             location="known",
                             attestation_slot=Slot(3),
-                            target_slot=Slot(3),
+                            target_slot=Slot(2),
                         ),
                         AttestationCheck(
                             validator=ValidatorIndex(2),
@@ -379,7 +379,7 @@ def test_equivocation_head_independent_of_arrival_order_a_then_b(
     - V1 backs fork_a alone; V2 backs fork_b alone.
     - without V0 the two forks tie one-to-one, so V0's vote decides the head.
     - the two votes tie on slot, so the larger attestation-data root wins.
-    - the fork_a vote carries the larger root.
+    - the fork_b vote carries the larger root.
 
     When
     ----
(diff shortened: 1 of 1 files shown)


mananuf:
Closing this — the fix is incorrect, and the root cause turned out to be **scheme dependence**, not a code-vs-vector mismatch.

CI (`fill --fork=Lstar`, default `test` scheme) produces heads that are the **exact opposite** of what I saw when reproducing/validating under `--scheme=prod` (the scheme our downstream client uses for real-XMSS fixtures):

| test | `test` scheme (CI) | `prod` scheme |
|---|---|---|
| `same_slot` | fork_b (3) | fork_a (2) |
| `a_then_b` / `b_then_a` | fork_a (2) | fork_b (3) |

Different scheme → different validator keys → different genesis `state_root` → different block roots → different attestation-data roots → the equal-slot tie-break (larger attestation-data root) resolves to the **opposite fork**. The fork-choice code is correct and deterministic under each scheme; the vectors' hard-coded heads simply can't be valid for both.

So the existing vectors are correct for the canonical `test` scheme, and this PR's hard-coded `prod` values break the `test`-scheme CI. Reverting would leave an empty PR, so closing. I've corrected #1187 with the accurate analysis — apologies for the noise.

PR #1189, opened 2026-07-09
fix(testing): make equivocation head assertions scheme-independent

leolara:
## Problem

The `Production Test Vectors` job failed on `main` (run 28858797181): 3 fork-choice equivocation vectors failed, 551 passed:

```
test_same_slot_equivocating_attesters_count_once            head_slot = 2, expected 3
test_equivocation_head_independent_of_arrival_order_a_then_b head_slot = 3, expected 2
test_equivocation_head_independent_of_arrival_order_b_then_a head_slot = 3, expected 2
```

## Root cause

PR #1181 breaks an equal-slot equivocation tie toward the fork with the **largest attestation-data root**. That root embeds the target block root, which embeds the genesis state root, which embeds the validator XMSS public keys — so the winner **flips between the `test` and `prod` schemes**. The 3 vectors hardcoded the winning fork, so they passed under `test` (what PR CI runs) and failed under `prod` (what `prod-vectors` runs, on `main` only). Same class of bug as #916.

## Fix

The vectors now assert the **invariant**, not the lottery winner:

- New scheme-independent check `canonical_equivocation_head_among`: reads each fork's attestation root from the store and asserts the head is the largest-root fork. Mirrors the existing `lexicographic_head_among`; never pins which fork wins.
- Rewrote the 3 vectors to use it, dropping the scheme-fragile `head_slot`/`head_root_label` and the equivocators' `target_slot` pins. Given/When/Then docstrings describe the invariant.
- Docstring on the tiebreak (`fork_choice.py`) flags that the winner's identity is key-material-dependent.
- `CLAUDE.md`: new rule — fork-choice head assertions must be scheme-independent.

No consensus-rule change; the tiebreak behavior is unchanged.

## Testing

- The 3 vectors pass under **both** `--scheme=test` and `--scheme=prod`.
- Full fork-choice tree (123 vectors) passes under both schemes.
- `just check` clean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,41 +159,6 @@ jobs:
       - name: Fill test fixtures
         run: just fill-ci
 
-  # Scheme-fragility guard. A root-based fork-choice tiebreak embeds the validator XMSS public
-  # keys, so a vector that hardcodes the winning fork can pass under the default `test` scheme
-  # and fail under `prod`. This reruns the fork-choice vectors under the prod scheme to catch
-  # that pre-merge. See CLAUDE.md, "FORK-CHOICE HEAD ASSERTIONS MUST BE SCHEME-INDEPENDENT".
-  # Determinism is already gated by fill-tests; this job disables it to stay a fast smoke check.
-  fill-tests-prod-scheme:
-    name: Fill fork-choice fixtures under the prod scheme - Python 3.14
-    runs-on: macos-latest
-    timeout-minutes: 20
-    steps:
-      - name: Checkout leanSpec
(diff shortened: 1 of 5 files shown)


tcoratger:
@leolara can we reduce the number of line for the new documentation as per doc writer agent rules to keep things lean and facilitate the reading experience ? :)

leolara:
> @leolara can we reduce the number of line for the new documentation as per doc writer agent rules to keep things lean and facilitate the reading experience ? :)

You are right, we are all learning about how to use these new tools.

tcoratger (approved):

